PR #2, opened 2022-09-15
git: define common refs

slack-coder:
Add helper methods to improve source code clarity and prevent bugs due to typos.

### DIFF
--- a/node/src/git.rs
+++ b/node/src/git.rs
@@ -2,6 +2,7 @@ use std::path::Path;
 use std::str::FromStr;
 
 use git_ref_format as format;
+use once_cell::sync::Lazy;
 
 use crate::collections::HashMap;
 use crate::crypto::PublicKey;
@@ -33,6 +34,44 @@ pub enum ListRefsError {
     Git(#[from] git2::Error),
     #[error("invalid ref: {0}")]
     InvalidRef(#[from] RefError),
+}
+
+pub mod refs {
+    use super::*;
+
+    /// Where project information is kept.
+    pub static IDENTITY_BRANCH: Lazy<RefString> = Lazy::new(|| refname!("radicle/id"));
+
+    pub mod storage {
+        use super::*;
+
+        pub fn branch(remote: &RemoteId, branch: &str) -> String {
+            format!("refs/remotes/{remote}/heads/{branch}")
+        }
+
+        /// Get the branch used to track project information.
+        pub fn id(remote: &RemoteId) -> String {
+            branch(remote, &IDENTITY_BRANCH)
+        }
+    }
+
+    pub mod workdir {
+        pub fn branch(branch: &str) -> String {
+            format!("refs/heads/{branch}")
+        }
+
+        pub fn note(name: &str) -> String {
+            format!("refs/notes/{name}")
+        }
+
+        pub fn remote_branch(remote: &str, branch: &str) -> String {
+            format!("refs/remotes/{remote}/{branch}")
+        }
+
+        pub fn tag(name: &str) -> String {
+            format!("refs/tags/{name}")
+        }
+    }
 }
 
 /// List remote refs of a project, given the remote URL.
--- a/node/src/identity/doc.rs
+++ b/node/src/identity/doc.rs
@@ -27,7 +27,6 @@ pub struct Untrusted;
 #[derive(Clone, Copy, Debug)]
 pub struct Trusted;
 
-pub static REFERENCE_NAME: Lazy<git::RefString> = Lazy::new(|| git::refname!("heads/radicle/id"));
 pub static PATH: Lazy<&Path> = Lazy::new(|| Path::new("radicle.json"));
 
 pub const MAX_STRING_LENGTH: usize = 255;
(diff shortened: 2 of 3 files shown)


node/src/git.rs
```diff
@@ -34,6 +34,35 @@ pub enum ListRefsError {
...
+    pub mod storage {
+        use super::*;
+
+        pub fn peer_branch(remote: &RemoteId, branch: &str) -> String {
+            format!("refs/remotes/{remote}/heads/{branch}")
+        }
+
+        pub fn peer_id(remote: &RemoteId) -> String {
+            peer_branch(remote, IDENTITY_BRANCH)
+        }
```
cloudhead, line 52: I wonder if we should just call these `refs::storage::branch` and `refs::storage::id`, because there's no "local" scope, and `peer_id` for eg. makes it sound like it's the peer's identity, not the project's.

node/src/git.rs
```diff
@@ -34,6 +34,35 @@ pub enum ListRefsError {
     InvalidRef(#[from] RefError),
 }
 
+pub mod refs {
+    use super::*;
+
+    // Where project information is kept.
+    pub const IDENTITY_BRANCH: &str = "radicle/id";
```
FintanH, line 41: You can use [`lazy_static`][0] and the `refname!` macro from `git-ref-format` to define this. We did something similar in `radicle-link`:
```rust
lazy_static! {
    pub static ref DEFAULT_PATH: ext::Qualified = ext::Qualified::from(reflike!("refs/rad/id"));
}
```

[0]: https://docs.rs/lazy_static/1.4.0/lazy_static/

node/src/git.rs
```diff
@@ -34,6 +34,35 @@ pub enum ListRefsError {
...
+        pub fn peer_branch(remote: &RemoteId, branch: &str) -> String {
+            format!("refs/remotes/{remote}/heads/{branch}")
+        }
+
+        pub fn peer_id(remote: &RemoteId) -> String {
+            peer_branch(remote, IDENTITY_BRANCH)
+        }
+    }
+
+    pub mod workdir {
```
FintanH, line 55: For completion would we also add `tags` and `notes` too?

cloudhead: Yeah.

node/src/git.rs
cloudhead: Yes, we're using `once_cell` in this crate so it would look like:
```rust
pub static REMOTES_GLOB: Lazy<refspec::PatternString> =
    Lazy::new(|| refspec::pattern!("refs/remotes/*"));
```

slack-coder: Why is it so complicated to define a constant in rust?

FintanH: > Why is it so complicated to define a constant in rust?

I _think_ because they need to be primitives that can live on the stack, i.e. known at compile time values. Hence, why you can't use `String` but you can use `&str`. Another way I think about it is that if you need to do a computation to get the constant then it can't be constant.


> Why is this better?

This is an improvement over `&str` because it's a typed representation of `refs/rad/id`, the `Qualified` type guarantees that we a have `refs/` prefix -- as opposed to `&str` being any old value. Here's a great article on the benefits of being more specific about types[[0]].

The `lazy_static`, I believe, works around this computation idea I mentioned above. It will run the computation once and only once, so it's constant after that first run. `once_cell` does something similar.

[0]: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

slack-coder: @FintanH thanks for the explanation :)

I wasn't aware about how much of it was due to type being used and personally unknown rust features due to other concerns.

node/src/git.rs
```diff
@@ -34,6 +34,35 @@ pub enum ListRefsError {
...
+        pub fn local_branch(branch: &str) -> String {
+            format!("refs/heads/{branch}")
+        }
+
+        pub fn remote_branch(remote: &str, branch: &str) -> String {
+            format!("refs/remotes/{remote}/{branch}")
+        }
+    }
+}
```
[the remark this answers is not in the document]
slack-coder: Using RefLike will require more elaborate changes, one being error handling.  Since these changes were about defining the set of common refs, it may be best to leave introducing these types in another PR.

slack-coder:
Rebased and updated identity/doc.rs to use these newly defined refs.

node/src/git.rs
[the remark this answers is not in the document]
FintanH: > Using RefLike will require more elaborate changes, one being error handling. Since these changes were about defining the set of common refs, it may be best to leave introducing these types in another PR.

I don't think you'd need error handling. Also, note that @cloudhead said `RefString` and not `RefLike` -- two different iterations of the same concept :) 

You could model the above as:
```rust
/// Create a remote branch under the `remote` with the name `branch`.
pub fn remote_branch<'a>(remote: Component<'a>, branch: RefString) -> RefString {
    // into_inner turns the component into a Cow<'a, RefStr>, which we can `join` with `branch`
    let tail = remote.into_inner().join(branch);
    refs_remotes(tail)
}
```

However, there's no good way of creating a `Component` yet -- other than known ones. I think we could add a `component!` macro that makes it easier.

node/src/git.rs
cloudhead: Yes, worth noting that lazy_static/once_cell do *not* run the computation at compile-time, they simply cache the result after the first run, so it's *kind of* like a constant.

cloudhead (approved):

cloudhead:
Merged!

node/src/git.rs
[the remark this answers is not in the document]
FintanH: Just wanted to reply to this because I actually missed that there is a `component!` macro :)

https://github.com/radicle-dev/radicle-link/blob/master/git-ref-format/macro/src/lib.rs#L49-L55

PR #3, opened 2022-09-23
git: Use RefString types for git::refs

slack-coder:
Lean on rust's type system to improve safety.

### DIFF
--- a/radicle-node/src/test/tests.rs
+++ b/radicle-node/src/test/tests.rs
@@ -22,7 +22,7 @@ use crate::test::simulator;
 use crate::test::simulator::{Peer as _, Simulation};
 use crate::test::storage::MockStorage;
 use crate::LocalTime;
-use crate::{client, identity, rad, service, storage, test};
+use crate::{client, git, identity, rad, service, test};
 
 // NOTE
 //
@@ -397,7 +397,7 @@ fn test_push_and_pull() {
         &repo,
         "alice",
         "alice's repo",
-        storage::BranchName::from("master"),
+        git::refname!("master"),
         alice.signer(),
         alice.storage(),
     )
--- a/radicle-tools/Cargo.toml
+++ b/radicle-tools/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = { version = "1" }
+git-ref-format = { version = "0", features = ["serde", "macro"] }
 
 [dependencies.radicle]
 version = "0"
--- a/radicle-tools/src/rad-init.rs
+++ b/radicle-tools/src/rad-init.rs
@@ -1,5 +1,7 @@
 use std::path::Path;
 
+use radicle::git;
+
 fn main() -> anyhow::Result<()> {
     let cwd = Path::new(".").canonicalize()?;
     let name = cwd.file_name().unwrap().to_string_lossy().to_string();
@@ -9,7 +11,7 @@ fn main() -> anyhow::Result<()> {
         &repo,
         &name,
         "",
-        radicle::git::BranchName::from("master"),
+        git::refname!("master"),
         &profile.signer,
         &profile.storage,
     )?;
--- a/radicle/src/git.rs
+++ b/radicle/src/git.rs
@@ -16,7 +16,7 @@ pub use ext::Error;
 pub use ext::Oid;
 pub use git2 as raw;
 pub use git_ref_format as fmt;
-pub use git_ref_format::{refname, RefStr, RefString};
+pub use git_ref_format::{refname, Component, Qualified, RefStr, RefString};
 pub use git_url as url;
 pub use git_url::Url;
 pub use radicle_git_ext as ext;
@@ -41,6 +41,13 @@ pub enum ListRefsError {
     InvalidRef(#[from] RefError),
 }
 
+impl From<&RemoteId> for RefString {
+    fn from(id: &RemoteId) -> Self {
+        // PublicKey strings contain only legal characters.
+        RefString::try_from(id.to_string()).unwrap()
+    }
+}
+
 pub mod refs {
     use super::*;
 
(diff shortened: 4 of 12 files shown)


radicle-node/src/test/tests.rs
```diff
@@ -397,7 +397,7 @@ fn test_push_and_pull() {
         &repo,
         "alice",
         "alice's repo",
-        storage::BranchName::from("master"),
+        storage::BranchName::try_from("master").unwrap(),
```
cloudhead, line 400: Can we avoid this by having `BranchName::from(refname!("master"))` for eg.? Or perhaps we should just use `RefString` instead of `BranchName`?

radicle/src/git.rs
```diff
@@ -39,6 +39,13 @@ pub enum ListRefsError {
     InvalidRef(#[from] RefError),
 }
 
+impl From<&RemoteId> for RefString {
```
FintanH, line 42: I think it might make more sense to use `Component` here, similar to this https://github.com/radicle-dev/radicle-link/blob/master/link-crypto/src/peer.rs#L205-L213

radicle/src/git.rs
```diff
@@ -48,31 +55,36 @@ pub mod refs {
     pub mod storage {
         use super::*;
 
-        pub fn branch(remote: &RemoteId, branch: &str) -> String {
-            format!("refs/remotes/{remote}/heads/{branch}")
+        pub fn branch(remote: &RemoteId, branch: &RefStr) -> RefString {
+            refname!("refs/remotes")
+                .and::<RefString>(remote.into())
+                .and(refname!("heads"))
+                .and(branch)
```
FintanH, line 62:
```suggestion
        pub fn branch<R>(remote: &RemoteId, branch: R) -> Qualified
        where 
            R: AsRef<RefStr> 
        {
            lit::refs_remotes(
                remote,
                component!("heads").join(branch)
            ).into()
```

The `lit` module has some common patterns to save you typing out `refs/remotes`. The `component!` macro is used for single parts of the ref path. The `AsRef<RefStr>` makes the function more malleable allowing it to accept lots of different ref paths. The `Qualified` return type indicates that this is a ref path that starts with `refs` and has a category (in this case `remotes`).

radicle/src/git.rs
```diff
@@ -48,31 +55,36 @@ pub mod refs {
...
             branch(remote, &IDENTITY_BRANCH)
         }
     }
 
     pub mod workdir {
-        pub fn branch(branch: &str) -> String {
-            format!("refs/heads/{branch}")
+        use super::*;
+
+        pub fn branch(branch: &RefStr) -> RefString {
```
FintanH, line 74: You could apply similar suggestions to these functions.

radicle/src/git.rs
```diff
@@ -48,31 +55,36 @@ pub mod refs {
...
 
-        pub fn note(name: &str) -> String {
-            format!("refs/notes/{name}")
+        pub fn note(name: &RefStr) -> RefString {
+            refname!("refs/notes").join(name)
         }
 
-        pub fn remote_branch(remote: &str, branch: &str) -> String {
-            format!("refs/remotes/{remote}/{branch}")
+        pub fn remote_branch(remote: &RefStr, branch: &RefStr) -> RefString {
```
FintanH, line 82:
```suggestion
        pub fn remote_branch(remote: &Component, branch: &RefStr) -> RefString {
```

radicle/src/rad.rs
```diff
@@ -12,7 +13,7 @@ use crate::storage::refs::SignedRefs;
 use crate::storage::{BranchName, ReadRepository as _, RemoteId, WriteRepository as _};
 use crate::{identity, storage};
 
-pub const REMOTE_NAME: &str = "rad";
+pub static REMOTE_NAME: Lazy<git::RefString> = Lazy::new(|| git::refname!("rad"));
```
FintanH, line 16: This could be `Component` instead

radicle-node/src/test/tests.rs
slack-coder: It looks like BranchName was already type alias'ed to git::refstring in the PR.

I prefer using 'try_from(..).unwrap(..)' for the tests.  Given we can panic (unwrap) happily in tests, so I believe it is ok.  Type support for try_from(..) is common, and we can reduce complexity by consistently using try_from(string) instead of different (string -> some_type) methods for different types.

slack-coder: Updated to use git::refname!(..)

slack-coder:
The learning curve to understanding and working with the git_ref types is high, so I opted to keep most things as refstrings for simplicity.

For instance the example given here, is complicated for joining strings together.  Maybe these types can be improved to make them simpler?

https://github.com/radicle-dev/radicle-node/pull/2#discussion_r975603051

```
/// Create a remote branch under the `remote` with the name `branch`.
pub fn remote_branch(remote: Component, branch: RefString) -> Qualified {
    // into_inner turns the component into a Cow<'a, RefStr>, which we can `join` with `branch`
    let tail = remote.into_inner().join(branch);
    refs_remotes(tail).into()
}
```

radicle/src/git.rs
cloudhead: hmm the current code reads a lot better though.

I think the `refs_remote` api could be improved to maybe:
```rust
lit::refs_remotes(remote).and(component!("heads")).and(branch)
```

radicle-tools/src/rad-init.rs
```diff
@@ -9,7 +9,7 @@ fn main() -> anyhow::Result<()> {
         &repo,
         &name,
         "",
-        radicle::git::BranchName::from("master"),
+        radicle::git::BranchName::try_from("master").unwrap(),
```
cloudhead, line 12: refname!

radicle/src/git.rs
```diff
@@ -39,6 +39,13 @@ pub enum ListRefsError {
     InvalidRef(#[from] RefError),
 }
 
+impl From<&RemoteId> for RefString {
+    fn from(id: &RemoteId) -> Self {
+        // PublicKey strings contain only legal characters.
+        RefString::try_from(id.to_string()).unwrap()
+    }
```
cloudhead, line 46: If you find yourself implementing `Foo::from(some_other_type)`, it's probably better to implement a `From` instance.

radicle/src/git.rs
FintanH: You could do that just need to convert the `refs_remote` into a `Qualified` so:
```rust
lit::refs_remotes(remote).into().and(component!("heads")).and(branch)
```

FintanH:
> The learning curve to understanding and working with the git_ref types is high, so I opted to keep most things as refstrings for simplicity.

I think it's worth learning because you can be precise when talking about different references :) 

> For instance the example given here, is complicated for joining strings together. Maybe these types can be improved to make them simpler?

Ah yes, I mean it could also be represented similar to [this comment][[0]]:

```rust
/// Create a remote branch under the `remote` with the name `branch`.
pub fn remote_branch(remote: Component, branch: impl AsRef<RefStr>) -> Qualified {
    refs_remotes(remote).into().and(branch)
}
```

[0]: https://github.com/radicle-dev/radicle-node/pull/3#discussion_r980993002

slack-coder:
> > The learning curve to understanding and working with the git_ref types is high, so I opted to keep most things as refstrings for simplicity.
> 
> I think it's worth learning because you can be precise when talking about different references :)
> 
> > For instance the example given here, is complicated for joining strings together. Maybe these types can be improved to make them simpler?
> 
> Ah yes, I mean it could also be represented similar to [this comment][[0](https://github.com/radicle-dev/radicle-node/pull/3#discussion_r980993002)]:
> 
> ```rust
> /// Create a remote branch under the `remote` with the name `branch`.
> pub fn remote_branch(remote: Component, branch: impl AsRef<RefStr>) -> Qualified {
>     refs_remotes(remote).into().and(branch)
> }
> ```
The intent is good, and they are well documented.  The problem is implementing them to a high degree adds friction to anyone reading or contributing to the project.  Source code is read more than it is written, so it should be easy to read too.  I've personally had a very hard time trying to do simple tasks with them.  I'd love to improve them to make them more 'ergonomic' and clear.

slack-coder:
Rebased to fix merge conflicts, and used refname!(..) globally in place of BranchName::try_from(..)

radicle-tools/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = { version = "1" }
+git-ref-format = { version = "0", features = ["serde", "macro"] }
```
cloudhead, line 10: I think this is not needed?

cloudhead: Nevermind, it's needed because it's a macro.

cloudhead:
Merged this as it's a nice improvement already. If we want to later make more use of `Component` in a different PR, we can do that :+1:

PR #5, opened 2022-09-28
Rebroadcast logic for stored gossip messages

cloudhead:

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -28,7 +28,7 @@ use crate::git;
 use crate::git::Url;
 use crate::identity::{Doc, Id};
 use crate::service::config::ProjectTracking;
-use crate::service::message::Address;
+use crate::service::message::{Address, Announcement, AnnouncementMessage};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
 use crate::service::peer::{Session, SessionError, SessionState};
 use crate::storage;
@@ -37,6 +37,7 @@ use crate::storage::{Inventory, ReadRepository, RefUpdate, WriteRepository, Writ
 pub use crate::service::config::{Config, Network};
 pub use crate::service::message::{Envelope, Message};
 
+use self::gossip::Gossip;
 use self::message::{InventoryAnnouncement, NodeFeatures};
 use self::reactor::Reactor;
 
@@ -132,6 +133,8 @@ pub struct Service<A, S, G> {
     storage: S,
     /// Tracks the location of projects.
     routing: Routing,
+    /// State relating to gossip.
+    gossip: Gossip,
     /// Peer sessions, currently or recently connected.
     sessions: Sessions,
     /// Keeps track of peer states.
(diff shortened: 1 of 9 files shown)


FintanH (approved):
From a glance, looks good to me :)

Still need to get familiar with the MVP node code

cloudhead:
Sorry should have marked this one as draft, didn't realize someone was looking at it :smile: 

Anyway mostly refactored stuff.

radicle-node/src/service/message.rs
```diff
@@ -173,46 +166,119 @@ impl wire::Decode for NodeAnnouncement {
...
+/// Announcement messages are messages that are relayed between peers.
+#[derive(Clone, PartialEq, Eq)]
+pub enum AnnouncementMessage {
+    /// Inventory announcement.
+    Inventory(InventoryAnnouncement),
+    /// Node announcement.
+    Node(NodeAnnouncement),
+    /// Refs announcement.
+    Refs(RefsAnnouncement),
 }
```
slack-coder, line 199: What about naming them 'gossip' or 'relayed' messages?  Describing how they are treated as a message type instead of how they are currently used.

radicle-node/src/service/message.rs
```diff
@@ -173,46 +166,119 @@ impl wire::Decode for NodeAnnouncement {
...
 
-impl RefsAnnouncement {
-    /// Verify a signature on this message.
-    pub fn verify(&self, signer: &NodeId, signature: &crypto::Signature) -> bool {
-        let msg = wire::serialize(self);
-        signer.verify(&msg, signature).is_ok()
+impl AnnouncementMessage {
+    /// Sign this announcement message.
+    pub fn signed<S: crypto::Signer>(self, signer: S) -> Announcement {
```
slack-coder, line 203: Was there a reason to change the method to past tense?

radicle-node/src/service.rs
```diff
@@ -886,6 +879,34 @@ impl DerefMut for Sessions {
...
+        pub fn received(&mut self, ann: Announcement, time: Timestamp) {
+            self.received.push((time, ann));
+        }
+
+        pub fn filtered<'a>(
+            &'a self,
+            filter: &'a Filter,
+            start: Timestamp,
+            end: Timestamp,
+        ) -> impl Iterator<Item = Message> + '_ {
```
slack-coder, line 906: Same as above re: action vs past tense.

radicle-node/src/service/message.rs
cloudhead: All methods that are "functional" in that they move self and return a "changed" type use the past tense. So eg. in this case, the method doesn't sign `AnnouncementMessage`, it returns a signed announcement.

radicle-node/src/service.rs
cloudhead: Same reasoning, we return a different type, whereas I'd expect a method called "filter" to return `Item = (Timestamp, Announcement)`

PR #7, opened 2022-10-03
Implement commit parsing/formatting

cloudhead:
To be able to support git commits with multiple signature headers, we have to implement our own parsing and formatting.

### DIFF
--- a/radicle/src/git.rs
+++ b/radicle/src/git.rs
@@ -274,3 +274,180 @@ pub fn run<P: AsRef<Path>, S: AsRef<std::ffi::OsStr>>(
         String::from_utf8_lossy(&output.stderr),
     ))
 }
+
+/// Parsing and formatting of commit objects.
+/// This module exists to work with commits that have multiple signature headers.
+pub mod commit {
+    use std::str::FromStr;
+
+    /// A parsed commit object.
+    /// Contains the full commit header and body.
+    ///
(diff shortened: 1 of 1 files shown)


FintanH (approved):
Nice, would this be useful to have in `radicle-git`?

cloudhead:
> Nice, would this be useful to have in `radicle-git`?

Yup, I think eventually we can move some of the git code into `radicle-git` and depend on that, instead of having it here.

FintanH:
> Yup, I think eventually we can move some of the git code into `radicle-git` and depend on that, instead of having it here.

Sweet, would be good to keep track of those things as issues in `radicle-git` and have some understanding of when you'd like to move them over. I imagine you're keeping it here for quicker iteration, right?

cloudhead:
> > Yup, I think eventually we can move some of the git code into `radicle-git` and depend on that, instead of having it here.
> 
> Sweet, would be good to keep track of those things as issues in `radicle-git` and have some understanding of when you'd like to move them over. I imagine you're keeping it here for quicker iteration, right?

Yeah, I think 99% of it should be in `radicle::git` and `radicle::storage::git`. It's indeed easier for me to make changes here until the code and interfaces stabilize. There is also the option of having more of a monorepo and moving `radicle-git` in here in the long run.

slack-coder:
Looks good to me too

PR #9, opened 2022-10-03
Service: drop unresponsive peers

slack-coder:
Use pings to check connection health with network peers and drop them if they prove unresponsive.

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -33,12 +33,12 @@ use crate::node;
 use crate::service::config::ProjectTracking;
 use crate::service::message::{Address, Announcement, AnnouncementMessage};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
-use crate::service::peer::{SessionError, SessionState};
+use crate::service::peer::{PingState, SessionError, SessionState};
 use crate::storage;
 use crate::storage::{Inventory, ReadRepository, RefUpdate, WriteRepository, WriteStorage};
 
 pub use crate::service::config::{Config, Network};
-pub use crate::service::message::{Envelope, Message};
+pub use crate::service::message::{Envelope, Message, ZeroBytes};
 pub use crate::service::peer::Session;
 
 use self::gossip::Gossip;
@@ -53,6 +53,11 @@ pub const ANNOUNCE_INTERVAL: LocalDuration = LocalDuration::from_secs(30);
 pub const SYNC_INTERVAL: LocalDuration = LocalDuration::from_secs(60);
 pub const PRUNE_INTERVAL: LocalDuration = LocalDuration::from_mins(30);
 pub const MAX_CONNECTION_ATTEMPTS: usize = 3;
+
+/// Duration to wait on an unresponsive peer before dropping its connection.
+pub const STALE_CONNECTION_TIMEOUT: LocalDuration = LocalDuration::from_secs(60);
+
+pub const KEEP_ALIVE_DELTA: LocalDuration = LocalDuration::from_secs(30);
 pub const MAX_TIME_DELTA: LocalDuration = LocalDuration::from_mins(60);
 
 /// Network node identifier.
@@ -362,6 +367,8 @@ where
         if now - self.last_idle >= IDLE_INTERVAL {
             debug!("Running 'idle' task...");
 
+            self.keep_alive(&now);
+            self.disconnect_unresponsive_peers(&now);
             self.maintain_connections();
             self.reactor.wakeup(IDLE_INTERVAL);
             self.last_idle = now;
@@ -485,7 +492,7 @@ where
         let peer = self
             .sessions
             .entry(ip)
-            .or_insert_with(|| Session::new(*addr, Link::Outbound, persistent));
+            .or_insert_with(|| Session::new(*addr, Link::Outbound, persistent, &self.rng));
 
         peer.attempted();
     }
@@ -522,7 +529,12 @@ where
         } else {
             self.sessions.insert(
                 ip,
-                Session::new(addr, Link::Inbound, self.config.is_persistent(&address)),
+                Session::new(
+                    addr,
+                    Link::Inbound,
+                    self.config.is_persistent(&address),
+                    &self.rng,
+                ),
             );
         }
     }
@@ -724,13 +736,14 @@ where
         } else {
             return Err(SessionError::NotFound(remote.ip()));
         };
+        peer.last_active = self.clock.local_time();
 
         if envelope.magic != self.config.network.magic() {
             return Err(SessionError::WrongMagic(envelope.magic));
         }
         debug!("Received {:?} from {}", &envelope.msg, peer.ip());
 
-        match (&peer.state, envelope.msg) {
+        match (&mut peer.state, envelope.msg) {
             (
                 SessionState::Initial,
                 Message::Initialize {
@@ -764,6 +777,7 @@ where
                     since: self.clock.local_time(),
                     addrs,
                     git,
+                    ping: Default::default(),
                 };
             }
             (SessionState::Initial, _) => {
@@ -811,6 +825,19 @@ where
                     peer.ip()
                 );
                 return Err(SessionError::Misbehavior);
+            }
+            (SessionState::Negotiated { .. }, Message::Ping { ponglen, .. }) => {
+                let resp = Message::Pong {
+                    zeroes: ZeroBytes::new(ponglen),
+                };
+                self.reactor.write(peer.addr, resp);
+            }
+            (SessionState::Negotiated { ping, .. }, Message::Pong { zeroes }) => {
+                if let PingState::AwaitingResponse(ponglen) = *ping {
+                    if (ponglen as usize) == zeroes.len() {
+                        *ping = PingState::Ok;
+                    }
+                }
             }
             (SessionState::Disconnected { .. }, msg) => {
                 debug!("Ignoring {:?} from disconnected peer {}", msg, peer.ip());
@@ -881,6 +908,29 @@ where
     fn prune_routing_entries(&mut self) -> Result<(), storage::Error> {
         // TODO
         Ok(())
+    }
+
+    fn disconnect_unresponsive_peers(&mut self, now: &LocalTime) {
+        let stale = self
+            .sessions
+            .negotiated()
+            .filter(|(_, _, session)| session.last_active < *now - STALE_CONNECTION_TIMEOUT);
+        for (_, _, session) in stale {
+            self.reactor
+                .disconnect(session.addr, DisconnectReason::Error(SessionError::Timeout));
+        }
+    }
+
+    /// Ensure connection health by pinging connected peers.
+    fn keep_alive(&mut self, now: &LocalTime) {
+        let inactive_sessions = self
+            .sessions
+            .negotiated_mut()
+            .filter(|(_, session)| session.last_active < *now - KEEP_ALIVE_DELTA)
+            .map(|(_, session)| session);
+        for session in inactive_sessions {
+            session.ping(&mut self.reactor, &self.rng).ok();
+        }
     }
 
     fn maintain_connections(&mut self) {
@@ -1079,6 +1129,11 @@ impl Sessions {
                 SessionState::Negotiated { id, .. } => Some((ip, id, sess)),
                 _ => None,
             })
+    }
+
+    /// Iterator over mutable fully negotiated peers.
+    pub fn negotiated_mut(&mut self) -> impl Iterator<Item = (&IpAddr, &mut Session)> {
+        self.0.iter_mut().filter(move |(_, p)| p.is_negotiated())
     }
 }
 
--- a/radicle-node/src/service/message.rs
+++ b/radicle-node/src/service/message.rs
@@ -315,6 +315,23 @@ pub enum Message {
     /// Gossip announcement. These messages are relayed to peers, and filtered
     /// using [`Message::Subscribe`].
     Announcement(Announcement),
+
+    /// Ask a connected peer for a Pong.
+    ///
+    /// Use to check if the remote peer is responsive or a side-effect free way to keep a
+    /// connection alive.
+    Ping {
+        /// The desired response length
+        ponglen: u16,
+        /// The ping payload.
+        zeroes: ZeroBytes,
+    },
+
+    /// Response to `Ping` message.
+    Pong {
+        /// The pong payload.
+        zeroes: ZeroBytes,
+    },
 }
 
 impl Message {
@@ -373,7 +390,26 @@ impl fmt::Debug for Message {
             Self::Announcement(Announcement { node, message, .. }) => {
                 write!(f, "Announcement({}, {:?})", node, message)
             }
+            Self::Ping { ponglen, zeroes } => write!(f, "Ping({ponglen}, {:?})", zeroes),
+            Self::Pong { zeroes } => write!(f, "Pong({:?})", zeroes),
         }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ZeroBytes(u16);
+
+impl ZeroBytes {
+    pub fn new(arg: u16) -> Self {
+        ZeroBytes(arg)
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.0 == 0
+    }
+
+    pub fn len(&self) -> usize {
+        self.0.into()
     }
 }
 
--- a/radicle-node/src/service/peer.rs
+++ b/radicle-node/src/service/peer.rs
@@ -1,5 +1,19 @@
 use crate::service::message::*;
 use crate::service::*;
+use crate::wire;
+
+use std::mem::size_of;
+
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+pub enum PingState {
+    #[default]
+    /// The peer has not been sent a ping.
+    None,
+    /// A ping has been sent and is waiting on the peer's response.
+    AwaitingResponse(u16),
+    /// The peer was successfully pinged.
+    Ok,
+}
 
 #[derive(Debug, Default, Clone)]
 #[allow(clippy::large_enum_variant)]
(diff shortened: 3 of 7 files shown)


radicle-node/src/wire/message.rs
```diff
@@ -324,6 +346,26 @@ impl wire::Decode for Address {
...
+
+impl wire::Decode for ZeroBytes {
+    fn decode<R: std::io::Read + ?Sized>(reader: &mut R) -> Result<Self, wire::Error> {
+        let zeroes = u16::decode(reader)?;
+        for _ in 0..zeroes {
+            _ = u8::decode(reader)?;
+        }
+        Ok(Self(zeroes))
+    }
+}
```
cloudhead, line 367: Property test for these would be nice.

radicle-node/src/service/peer.rs
```diff
@@ -52,6 +62,10 @@ pub struct Session {
     /// Peer subscription.
     pub subscribe: Option<Subscribe>,
 
+    // How can I move PingState into the SessionState?
```
cloudhead, line 65: ?

radicle-node/src/service/peer.rs
```diff
@@ -35,6 +43,8 @@ pub enum SessionError {
     NotFound(net::IpAddr),
     #[error("peer misbehaved")]
     Misbehavior,
+    #[error("peer became unresponsive")]
```
cloudhead, line 46:
```suggestion
    #[error("peer timed out")]
```

radicle-node/src/service/peer.rs
```diff
@@ -1,6 +1,14 @@
 use crate::service::message::*;
 use crate::service::*;
 
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+pub enum PingState {
+    #[default]
+    None,
+    AwaitingResponse(u16),
+    Ok,
+}
```
cloudhead, line 10: Maybe document the variants?

radicle-node/src/service/message.rs
```diff
@@ -365,10 +376,22 @@ impl fmt::Debug for Message {
...
 
+// Should ZeroBytes be in wire/messsages instead?
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ZeroBytes(pub u16);
+
+impl fmt::Display for ZeroBytes {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "ZeroBytes({})", self.0)
+    }
+}
```
cloudhead, line 393: Do we need a `Display` instance for this?

radicle-node/src/service/message.rs
```diff
@@ -365,10 +376,22 @@ impl fmt::Debug for Message {
...
             }
+            Self::Ping { ponglen, zeroes } => write!(f, "Ping({ponglen}, {zeroes})"),
+            Self::Pong { zeroes } => write!(f, "Pong({zeroes})"),
         }
     }
 }
 
+// Should ZeroBytes be in wire/messsages instead?
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ZeroBytes(pub u16);
```
cloudhead, line 387: I wouldn't expose the byte length publically, kind of defeats part of the purpose of this struct. I'd instead have a `len() -> usize` or `len() -> Size` method on it.

radicle-node/src/service/message.rs
```diff
@@ -307,6 +307,17 @@ pub enum Message {
     /// Gossip announcement. These messages are relayed to peers, and filtered
     /// using [`Message::Subscribe`].
     Announcement(Announcement),
+    Ping {
+        /// The desired response length
+        ponglen: u16,
+        /// The ping payload size.
```
cloudhead, line 313:
```suggestion
        /// The ping payload.
```

radicle-node/src/service/message.rs
```diff
@@ -307,6 +307,17 @@ pub enum Message {
...
+        zeroes: ZeroBytes,
+    },
+
+    Pong {
+        /// The pong payload size.
```
cloudhead, line 318:
```suggestion
        /// The pong payload.
```

radicle-node/src/service.rs
```diff
@@ -554,6 +567,38 @@ where
         }
     }
 
+    pub fn ping(&mut self, peer_ip: &net::IpAddr) -> Result<(), peer::SessionError> {
+        let peer = if let Some(peer) = self.sessions.get_mut(peer_ip) {
+            peer
+        } else {
+            return Err(SessionError::NotFound(*peer_ip));
+        };
```
cloudhead, line 575: I'd suggest `let peer = self.sessions.get_mut(..).ok_or_else(|| SessionError::NotFound(*peer_ip))?`

radicle-node/src/service.rs
```diff
@@ -554,6 +567,38 @@ where
...
+
+        let ping_upper = 16;
+        let ponglen = self.rng.u16(1..ping_upper);
```
cloudhead, line 578: For the upper, we should move it to a constant `MAX_PING_BYTES`, and we should set it to `u16::MAX - 4`, which is the maximum message size minus the message type length, minus the ping size.

radicle-node/src/service.rs
```diff
@@ -554,6 +567,38 @@ where
...
+        let msg = Message::Ping {
+            ponglen,
+            zeroes: message::ZeroBytes(self.rng.u16(1..ping_upper)),
+        };
+        self.reactor.write(peer.addr, msg);
+
+        // How can I access the messages' ponglen field?
+        peer.ping = PingState::AwaitingResponse(ponglen);
```
cloudhead, line 586:
```suggestion
        peer.ping = PingState::AwaitingResponse(ponglen);
        
```
Why do you need to access the field?

radicle-node/src/service.rs
```diff
@@ -554,6 +567,38 @@ where
...
+        let peer_ip = remote.ip();
+        let peer = if let Some(peer) = self.sessions.get(&peer_ip) {
+            peer
+        } else {
+            return Err(SessionError::NotFound(remote.ip()));
+        };
+
+        // When should peer.ping be set to None again?
+        Ok(peer.ping)
+    }
```
cloudhead, line 600: Looks like this function just exists to get the ping state? It should be removed in favor of accessing the session directly, if it's only for testing.

radicle-node/src/service.rs
```diff
@@ -777,6 +836,25 @@ where
         // TODO
     }
 
+    fn disconnect_unresponsive_peers(&mut self, now: LocalTime) {
+        let stale = self
+            .sessions
+            .negotiated()
+            .filter(|(_, session)| STALE_CONNECTION_TIMEOUT <= now - session.last_active);
```
cloudhead, line 843: Usually we put the variable on the left side of the comparison, since it is the "subject"

radicle-node/src/service.rs
```diff
@@ -777,6 +836,25 @@ where
...
+            self.reactor
+                .disconnect(session.addr, DisconnectReason::Error(SessionError::Timeout));
+        }
+    }
+
+    // Ensure connection health by pinging connected peers.
+    fn keep_alive(&mut self) {
+        let peer_ips: Vec<_> = self.sessions.negotiated().map(|(ip, _)| *ip).collect();
+        for ip in peer_ips {
+            let _ok = self.ping(&ip);
```
cloudhead, line 854:
```suggestion
            self.ping(&ip).ok();
```

cloudhead: Maybe we should only ping peers who have a time since `last_active` that is greater than `PING_INTERVAL`?

radicle-node/src/service.rs
```diff
@@ -50,6 +50,8 @@ pub const SYNC_INTERVAL: LocalDuration = LocalDuration::from_secs(60);
 pub const PRUNE_INTERVAL: LocalDuration = LocalDuration::from_mins(30);
 pub const MAX_CONNECTION_ATTEMPTS: usize = 3;
 pub const MAX_TIME_DELTA: LocalDuration = LocalDuration::from_mins(60);
+pub const STALE_CONNECTION_TIMEOUT: LocalDuration = LocalDuration::from_secs(30);
+pub const PING_INTERVAL: LocalDuration = LocalDuration::from_secs(10);
```
cloudhead, line 54: I'd set these to 60 and 30 to start

radicle-node/src/test/tests.rs
```diff
@@ -34,6 +34,54 @@ use crate::{client, git, identity, rad, service, test};
...
+    .initialize([&mut alice, &mut bob]);
+
+    alice.command(service::Command::Connect(bob.addr()));
+    sim.run_while([&mut alice, &mut bob], |s| !s.is_settled());
+
+    alice.ping(&bob.addr().ip()).unwrap();
+    sim.run_while([&mut alice, &mut bob], |s| !s.is_settled());
+
+    assert_eq!(PingState::Ok, alice.ping_resp(&bob.addr()).unwrap());
+}
```
cloudhead, line 56: Not sure what this test is testing, not clear to me :thinking:

radicle-node/src/test/tests.rs
```diff
@@ -34,6 +34,54 @@ use crate::{client, git, identity, rad, service, test};
...
+    assert_eq!(
+        1, /* number of disconnects */
+        alice
+            .outbox()
+            .filter(|p| { matches!(p, Io::Disconnect(_, _)) })
+            .count(),
+        "disconnect unresponsive peer bob"
+    )
+}
+
```
cloudhead, line 84: We should also test that our service responds to `ping` with a valid `pong`, and that invalid pongs are ignored.

cloudhead (commented):
This is looking good! Left some feedback.

radicle-node/src/service/message.rs
slack-coder: It's used by Message's fmt::Debug, is there an alternative option?

cloudhead: I would think implementing `fmt::Debug` would be enough?

radicle-node/src/service/message.rs
slack-coder: Would this affect match statements on messages?

radicle-node/src/service.rs
```diff
@@ -153,6 +158,8 @@ pub struct Service<A, S, G> {
     last_idle: LocalTime,
     /// Last time the service synced.
     last_sync: LocalTime,
+    /// Last time the service tried pinging its connected peers.
+    last_ping: LocalTime,
```
cloudhead, line 162: Can get rid of this

slack-coder: fixed

radicle-node/src/service.rs
cloudhead: This got lost.

radicle-node/src/service/message.rs
cloudhead: Don't think so..

radicle-node/src/test/tests.rs
```diff
@@ -34,6 +34,73 @@ use crate::{client, git, identity, rad, service, test};
...
+    alice.connect_to(&bob);
+    assert_eq!(1, alice.sessions().negotiated().count(), "bob connects");
+    alice.elapse(STALE_CONNECTION_TIMEOUT + LocalDuration::from_secs(1));
+    assert_eq!(
+        1, /* number of disconnects */
+        alice
+            .outbox()
+            .filter(|p| { matches!(p, Io::Disconnect(addr, _) if addr == &bob.addr()) })
+            .count(),
+        "disconnect unresponsive peer bob"
```
cloudhead, line 71: Maybe this is better off with an `assert_matches!`?

radicle-node/src/test/tests.rs
```diff
@@ -34,6 +34,73 @@ use crate::{client, git, identity, rad, service, test};
...
+#[test]
+fn test_ping_response() {
+    let mut alice = Peer::new("alice", [8, 8, 8, 8], MockStorage::empty());
+    let bob = Peer::new("bob", [9, 9, 9, 9], MockStorage::empty());
+
+    alice.connect_to(&bob);
+    alice.receive(
+        &bob.addr(),
+        Message::Ping {
+            ponglen: 1,
```
cloudhead, line 46: I'd feel more confident if you used a more arbitrary value here :smile:

radicle-node/src/service/peer.rs
```diff
@@ -1,6 +1,17 @@
 use crate::service::message::*;
 use crate::service::*;
 
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+pub enum PingState {
+    #[default]
+    // The peer has not been sent a ping.
```
cloudhead, line 7: Docstrings

radicle-node/src/service/message.rs
slack-coder: There's a complication.  The match statement must be declarative (?) but making the field private requires using a public constructor.  Match statements can't contain functions (constructors).

radicle-node/src/test/tests.rs
```diff
@@ -34,6 +34,69 @@ use crate::{client, git, identity, rad, service, test};
...
+    let step: LocalDuration = STALE_CONNECTION_TIMEOUT / 10;
+    while elapsed < STALE_CONNECTION_TIMEOUT + step {
+        alice.elapse(step);
+        bob.elapse(step);
+        sim.run_while([&mut alice, &mut bob], |s| !s.is_settled());
+
+        elapsed = elapsed + step;
+    }
+
+    assert_eq!(1, alice.sessions().len(), "bob remains connected")
```
cloudhead, line 97: Should we check bob's sessions also for completeness?

radicle-node/src/service/message.rs
```diff
@@ -307,6 +307,17 @@ pub enum Message {
     /// Gossip announcement. These messages are relayed to peers, and filtered
     /// using [`Message::Subscribe`].
     Announcement(Announcement),
+    Ping {
```
cloudhead, line 310:
```suggestion

    /// Used to keep a connection alive.
    Ping {
```

radicle-node/src/service/message.rs
```diff
@@ -307,6 +307,17 @@ pub enum Message {
...
+        /// The desired response length
+        ponglen: u16,
+        /// The ping payload.
+        zeroes: ZeroBytes,
+    },
+
+    Pong {
```
cloudhead, line 317:
```suggestion
    /// Response to `Ping` message.
    Pong {
```

radicle-node/src/service/message.rs
```diff
@@ -365,10 +376,26 @@ impl fmt::Debug for Message {
             Self::Announcement(Announcement { node, message, .. }) => {
                 write!(f, "Announcement({}, {:?})", node, message)
             }
+            Self::Ping { ponglen, zeroes } => write!(f, "Ping({ponglen}, {:?})", zeroes),
+            Self::Pong { zeroes } => write!(f, "Pong({:?})", zeroes),
         }
     }
 }
 
+/// Should ZeroBytes be in wire/messsages instead?
```
cloudhead, line 385:
```suggestion
```

radicle-node/src/service/peer.rs
```diff
@@ -89,4 +111,20 @@ impl Session {
...
+
+    pub fn ping(&mut self, reactor: &mut Reactor, rng: &Rng) -> Result<(), SessionError> {
+        if let SessionState::Negotiated { ping, .. } = &mut self.state {
+            let ponglen = rng.u16(0..MAX_MESSAGE_SIZE_BYTES);
+            let msg = Message::Ping {
+                ponglen,
+                zeroes: message::ZeroBytes(
+                    rng.u16(0..(MAX_MESSAGE_SIZE_BYTES - (size_of::<u16>() as u16))),
+                ),
+            };
```
slack-coder, line 123: @cloudhead how do you feel about this approach?

Instead of having a PING_SIZE constant (and maybe PONG_SIZE), define the max message size as a single constant.  The intent is communicated clearly within the logic now.

radicle-node/src/service/message.rs
cloudhead: Why do you need to use a match statement though?

radicle-node/src/service/peer.rs
cloudhead: Yes, I like the approach, however you forgot one small detail: we need to account for both the message type (2 bytes) and the ponglen (2 bytes). Would be worth adding a test with the max size, to make sure the entire message once encoded is exactly `2^16 - 1`

I would also not expose the internals of `ZeroBytes`, but rather have a constructor, eg. `ZeroBytes::new(size: Size)`.

cloudhead: Ah and it might make sense for readability to use `size_of::<wire::Size>` to be clearer on intent.

radicle-node/src/service/message.rs
slack-coder: What's being gained by making it private?  We'd be adding restrictions for no gains.

There may be confusion in the intent of the type.  It represents a length of zero bytes on the wire, not in memory.

slack-coder: It is meant to be a type'd number, not similar to a struct.  Is using a single element tuple not the way to do this in rust?

radicle-node/src/service/peer.rs
```diff
@@ -1,5 +1,19 @@
...
+
+use std::mem::size_of;
+
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+pub enum PingState {
+    #[default]
+    /// The peer has not been sent a ping.
+    None,
+    /// A ping has been sent and is waiting on the peer's response.
+    AwaitingResponse(usize),
```
cloudhead, line 13: Should be `u16`

cloudhead: Or `wire::Size`

radicle-node/src/service.rs
```diff
@@ -777,6 +799,29 @@ where
         // TODO
     }
 
+    fn disconnect_unresponsive_peers(&mut self, now: LocalTime) {
```
FintanH, line 802: nit: I think this could be borrowed, same as `keep_alive` below
```suggestion
    fn disconnect_unresponsive_peers(&mut self, now: &LocalTime) {
```

PR #10, opened 2022-10-04
Implement proper `NodeFeatures` type

cloudhead:

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -28,6 +28,7 @@ use crate::crypto::{Signer, Verified};
 use crate::git;
 use crate::git::Url;
 use crate::identity::{Doc, Id};
+use crate::node;
 use crate::service::config::ProjectTracking;
 use crate::service::message::{Address, Announcement, AnnouncementMessage};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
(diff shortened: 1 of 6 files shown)


radicle-node/src/service/features.rs
```diff
@@ -0,0 +1,134 @@
+use std::{fmt, ops};
+
```
slack-coder, line 2: Is this included in its own module because of file size?  The module name spacing may be better missing one level.

For instance service::Features::None vs. service::features::NodeFeatures::None

slack-coder:
Otherwise :+1:

radicle-node/src/service/features.rs
FintanH: Or `service::features::Node::None`, if there is going to be another service feature type -- leaving room for that.

cloudhead: Yeah, agree this isn't optimal. Here's some other alternatives:

1. since these are constants, have a `constants` module and then `radicle_node::constants::NodeFeatures::SEED`
2. have them at the top level: `radicle_node::Features::SEED`
3. have them in the `node` module of `radicle`: `radicle::node::Features`
4. bitcoin calls these `ServiceFlags`, perhaps it makes sense in this context: `radicle_node::service::Flags`

FintanH: My personal preference would be 3. I love a good qualified hierarchy of modules :)

PR #11, opened 2022-10-04
Routing table persistence

cloudhead:


radicle-node/src/client.rs
```diff
@@ -73,24 +89,27 @@ impl<R: Reactor> Client<R> {
...
 
-    pub fn run(mut self, config: Config) -> Result<(), nakamoto_net::error::Error> {
+    pub fn run(mut self, config: Config) -> Result<(), Error> {
         let network = config.service.network;
         let rng = fastrand::Rng::new();
         let time = LocalTime::now();
         let storage = self.profile.storage;
         let signer = self.profile.signer;
         let addresses = HashMap::with_hasher(rng.clone().into());
+        let routing = routing::Table::open("routing.db")?;
```
slack-coder, line 99: Would making it a configuration based path make it better?  Like path::join(config.data, "routing.db")

radicle-node/src/service.rs
```diff
@@ -124,15 +122,15 @@ pub enum Command {
...
+pub struct Service<R, A, S, G> {
     /// Service configuration.
     config: Config,
     /// Our cryptographic signer and key.
     signer: G,
     /// Project storage.
     storage: S,
-    /// Tracks the location of projects.
-    routing: Routing,
+    /// Network routing table. Keeps track of where projects are located.
```
slack-coder, line 132: nitpick, the first period may not be required

radicle-node/src/service.rs
```diff
@@ -612,7 +609,8 @@ where
                 } else {
                     return Ok(false);
                 }
-                self.process_inventory(&message.inventory, *node, git);
+                self.process_inventory(&message.inventory, *node, git)
+                    .unwrap();
```
slack-coder, line 613: Is this known to never panic?  Maybe a comment why if so would help

slack-coder:
otherwise looks good to me

radicle-node/src/client.rs
cloudhead: Ah yes, forgot to do this, good catch!

radicle-node/src/service.rs
cloudhead: Nah, we're not there yet, ie. there are *lots* of unwraps in the codebase that need to be handled correctly. We'll have to do a pass at some point and fix them.

radicle-node/src/service/routing/schema.sql
```diff
@@ -0,0 +1,15 @@
+--
+-- Routing table SQL schema.
+--
+create table if not exists "routing" (
+  "id"           integer   primary key,
```
adaszko, line 5: It’s a matter of style of course but the quotes shouldn’t be necessary

radicle-node/src/service.rs
```diff
@@ -124,15 +122,15 @@ pub enum Command {
...
     /// Service configuration.
     config: Config,
     /// Our cryptographic signer and key.
     signer: G,
     /// Project storage.
     storage: S,
-    /// Tracks the location of projects.
-    routing: Routing,
+    /// Network routing table. Keeps track of where projects are located.
+    routing: R,
```
adaszko, line 133: I’d use a concrete type here for simplicity but you may have more context on what’s going to happen with this portion of code in the near term

radicle/Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2021"
 [features]
 default = []
 test = ["quickcheck"]
+sql = ["rusqlite"]
```
adaszko, line 11: Nitpicking: “sql” sounds a bit imprecise to me.  How about “persistent_routing_table”?

adaszko (approved):

radicle/Cargo.toml
cloudhead: The crate feature is just enabling the `From/ToSql` instances

PR #12, opened 2022-10-05
Add variable bloom filter sizes

cloudhead:
This should help small nodes waste less bandwidth, while standardizing on set sizes.

### DIFF
--- a/radicle-node/src/service/filter.rs
+++ b/radicle-node/src/service/filter.rs
@@ -1,37 +1,67 @@
+#![allow(clippy::identity_op)]
 use std::ops::{Deref, DerefMut};
 
 pub use bloomy::BloomFilter;
 
 use crate::identity::Id;
 
-/// Size in bytes of subscription bloom filter.
-pub const FILTER_SIZE: usize = 1024 * 16;
+/// Size in bytes of *large* bloom filter.
+/// It can store about 13'675 items with a false positive rate of 1%.
+pub const FILTER_SIZE_L: usize = 16 * 1024;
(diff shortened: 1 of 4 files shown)


radicle-node/src/service/filter.rs
```diff
@@ -1,38 +1,68 @@
...
+pub const FILTER_SIZE_L: usize = 16 * 1024;
+/// Size in bytes of *medium* bloom filter.
+/// It can store about 3'419 items with a false positive rate of 1%.
+pub const FILTER_SIZE_M: usize = 4 * 1024;
+/// Size in bytes of *small* bloom filter.
+/// It can store about 855 items with a false positive rate of 1%.
+pub const FILTER_SIZE_S: usize = 1 * 1024;
+
+/// Valid filter sizes.
+pub const FILTER_SIZES: [usize; 3] = [FILTER_SIZE_S, FILTER_SIZE_M, FILTER_SIZE_L];
```
slack-coder, line 19: I'm surprised rust can't deduce the array size at compile time :(

slack-coder: Its just the size.  Would be nice to be able to do something like:

```
pub const FILTER_SIZES: [usize; _] = [..]
```

slack-coder:
Otherwise :+1:

radicle-node/src/service/filter.rs
cloudhead: Hmm I bet you could create an RFC for that and get it accepted :grin:

PR #36, opened 2022-10-10
meta: add nix tooling

FintanH:
Adds nix tooling to allow people to develop using nix dependencies.

Uses niv to pin nixpkgs, rust-overlay, and niv itself.

This will give Rust tooling that is up to date with rustc-1.64.0.


default.nix
```diff
@@ -0,0 +1,22 @@
...
+, rust-overlay ? pkgs.rust-bin.stable.latest.default
+}:
+  with pkgs;
+  mkShell {
+    name = "build";
+    buildInputs = [
+        # cargo tooling
+        cargo-deny
+        cargo-watch
+        pkgs.rust-bin.nightly."2022-07-01".rustfmt
```
cloudhead, line 14: What's up with this?

FintanH: Oh, I can get rid of that. We need nightly in `radicle-git` for formatting -- which I copied this from. I'll remove it here.

FintanH:
@cloudhead you happy to merge this?

cloudhead:
> @cloudhead you happy to merge this?

Happy is a big word :joy: I'm just getting used to the idea of having all these obscure nix files in my repo. Is there a way to like, tuck them away somehow?

cloudhead:
You know like, `.nix/`, and have the shell/default files in there as well? That way it doesn't distract non-nix users?

FintanH:
> You know like, `.nix/`, and have the shell/default files in there as well? That way it doesn't distract non-nix users?

Poor nix users trying to not distract non-nix users :(

I've pushed a change to do that and I can always start the shell with `nix-shell .nix/shell.nix` -- annoying but fair.

PR #37, opened 2022-10-10
radicle: namespace peers

FintanH:
The previous iteration of the storage layout for projects organised peer's subtrees by the `refs/remotes` category. This layout leads to problems when linking a working copy to the storage project. The fetch refspec would look something like:

  `fetch = +refs/remotes/abc/heads/*:refs/remotes/rad/*`

This works perfectly fine when running:

  git fetch rad

but will fail when one tries to specify:

  `git fetch rad main`

Instead, the command would need to be:

  `git fetch remotes/abc/heads/main`

This patch proposes to move the use of `refs/namespaces` for each peer. The layout looks similar in that each peer has a subtree, but instead it is under `refs/namespaces/<peer>/refs/...`.

The previous refspec would look more familiar in its new form:

  `fetch = +refs/heads/*:refs/remotes/rad/*`

With this in place running:

  `git --namespace=abc rad main`

Will work as expected and place the reference under `refs/remotes/rad/main`. Similary, `git --namespace=abc push` will work as expected.

Another benefit of this approach is that the use of namespacing means that a tag is pushed it will be placed under the respective namespace rather than in the global `refs/tags` category.

The actual implementation needed to work around the fact that `git2`/`libgit2` does not have any options for fetching or pushing using namespaces. The working copy code uses processes to shell out to `git` for namespacing needs.

### DIFF
--- a/radicle/src/git.rs
+++ b/radicle/src/git.rs
@@ -18,7 +18,7 @@ pub use ext::NotFound;
 pub use ext::Oid;
 pub use git2 as raw;
 pub use git_ref_format as fmt;
-pub use git_ref_format::{refname, Component, Qualified, RefStr, RefString};
+pub use git_ref_format::{lit, name, refname, Component, Namespaced, Qualified, RefStr, RefString};
 pub use git_url as url;
 pub use git_url::Url;
 pub use radicle_git_ext as ext;
@@ -33,6 +33,14 @@ pub enum RefError {
     InvalidName(format::RefString),
     #[error("invalid ref format: {0}")]
     Format(#[from] format::Error),
+    #[error("expected ref to begin with 'refs/namespaces' but found '{0}'")]
+    MissingNamespace(format::RefString),
+    #[error("ref name contains invalid namespace identifier '{name}'")]
+    Id {
+        name: format::RefString,
+        #[source]
+        err: Box<dyn std::error::Error + Send + Sync + 'static>,
+    },
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -43,9 +51,11 @@ pub enum ListRefsError {
     InvalidRef(#[from] RefError),
 }
 
-impl From<&RemoteId> for RefString {
+impl<'a> From<&RemoteId> for Component<'a> {
     fn from(id: &RemoteId) -> Self {
-        RefString::try_from(id.to_string()).expect("encoded public keys are valid ref strings")
+        let refstr =
+            RefString::try_from(id.to_string()).expect("encoded public keys are valid ref strings");
+        Component::from_refstring(refstr).expect("encoded public keys are valid refname components")
     }
 }
 
@@ -56,17 +66,19 @@ pub mod refs {
     pub static IDENTITY_BRANCH: Lazy<RefString> = Lazy::new(|| refname!("radicle/id"));
 
     pub mod storage {
+
         use super::*;
 
-        pub fn branch(remote: &RemoteId, branch: &RefStr) -> RefString {
-            refname!("refs/remotes")
-                .and::<RefString>(remote.into())
-                .and(refname!("heads"))
-                .and(branch)
+        /// Create the [`Namespaced`] `branch` under the `remote` namespace, i.e.
+        /// `refs/namespaces/<remote>/refs/heads/<branch>`
+        pub fn branch<'a>(remote: &RemoteId, branch: &RefStr) -> Namespaced<'a> {
+            Qualified::from(git_ref_format::lit::refs_heads(branch)).add_namespace(remote.into())
         }
 
         /// Get the branch used to track project information.
-        pub fn id(remote: &RemoteId) -> RefString {
+        ///
+        /// `refs/namespaces/<remote>/refs/heads/radicle/id`
+        pub fn id(remote: &RemoteId) -> Namespaced {
             branch(remote, &IDENTITY_BRANCH)
         }
     }
@@ -74,18 +86,22 @@ pub mod refs {
     pub mod workdir {
         use super::*;
 
+        /// Create a [`RefString`] that corresponds to `refs/heads/<branch>`.
         pub fn branch(branch: &RefStr) -> RefString {
             refname!("refs/heads").join(branch)
         }
 
+        /// Create a [`RefString`] that corresponds to `refs/notes/<name>`.
         pub fn note(name: &RefStr) -> RefString {
             refname!("refs/notes").join(name)
         }
 
+        /// Create a [`RefString`] that corresponds to `refs/remotes/<remote>/<branch>`.
         pub fn remote_branch(remote: &RefStr, branch: &RefStr) -> RefString {
             refname!("refs/remotes").and(remote).and(branch)
         }
 
+        /// Create a [`RefString`] that corresponds to `refs/tags/<branch>`.
         pub fn tag(name: &RefStr) -> RefString {
             refname!("refs/tags").join(name)
         }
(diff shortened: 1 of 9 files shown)


radicle/src/storage.rs
```diff
@@ -174,13 +176,17 @@ pub struct Remote<V> {
     pub id: PublicKey,
     /// Git references published under this remote, and their hashes.
     pub refs: SignedRefs<V>,
+    // TODO(finto): Why does this live here? They could be a delegate of one project and not of another?
```
FintanH, line 179: I'm curious about this :eyes:

radicle/src/storage.rs
```diff
@@ -174,13 +176,17 @@ pub struct Remote<V> {
...
     /// Whether this remote is of a project delegate.
     pub delegate: bool,
     /// Whether the remote is verified or not, ie. whether its signed refs were checked.
     verified: PhantomData<V>,
 }
 
 impl<V> Remote<V> {
+    // TODO(finto): This function seems out of place in the API for a couple of reasons:
+    // * The SignedRefs aren't guaranteed to be by the `id`
+    // * I could write `Remote::<Verified>::new(id, refs) and because of the above, it's a LIE
```
FintanH, line 189: Also, this looks like a leaky function, but a second pair of eyes might help confirm that :)

radicle/src/git.rs
```diff
@@ -16,7 +16,7 @@ pub use ext::Error;
 pub use ext::Oid;
 pub use git2 as raw;
 pub use git_ref_format as fmt;
-pub use git_ref_format::{refname, Component, Qualified, RefStr, RefString};
+pub use git_ref_format::{lit, name, refname, Component, Namespaced, Qualified, RefStr, RefString};
```
cloudhead, line 19: This diff line reminds me of when Kim would patch my Haskell code but first he would import half of Control.Applicative

radicle/src/git.rs
```diff
@@ -104,27 +119,34 @@ pub fn remote_refs(url: &Url) -> Result<HashMap<RemoteId, Refs>, ListRefsError>
...
-
-    let mut components = suffix.components();
-    let id = components
-        .next()
-        .ok_or_else(|| RefError::InvalidName(input.to_owned()))?;
-    let id = T::from_str(&id.to_string()).map_err(|_| RefError::InvalidName(input.to_owned()))?;
-    let refstr = components.collect::<format::RefString>();
-
-    Ok((id, refstr))
+    match input.namespaced() {
```
cloudhead, line 135: Meta: I find this API or naming strange: it sounds like it is namespacing `input`, ie. "give me a namespaced version of input", even though input is already namespaced. I think perhaps this should be called `as_namespaced()` or `Namespaced::try_from(input)`

radicle/src/git.rs
```diff
@@ -197,22 +204,76 @@ pub fn write_tree<'r>(
...
+    Str: AsRef<RefStr>,
+{
+    let name = name
+        .into()
+        .map_or_else(|| "".to_owned(), |name| name.as_ref().as_str().to_owned());
+    let out = Command::new("git")
+        .current_dir(repo.path())
+        .env("GIT_NAMESPACE", Component::from(namespace).as_str())
+        .args(["push", remote, &name])
+        .output()?;
```
cloudhead, line 268: Maybe use the `run` function below? Might need to be changed to accept a namespace param?

radicle/src/git.rs
```diff
@@ -197,22 +204,76 @@ pub fn write_tree<'r>(
...
+    log::trace!(
+        "stdout: {}, stderr: {}",
+        String::from_utf8_lossy(&out.stdout),
+        String::from_utf8_lossy(&out.stderr)
+    );
+    Ok(())
+}
+
+// TODO: Would be nice if libgit2 respected namespaces https://github.com/libgit2/libgit2/issues/6414
+// TODO: could pass refspecs for pushing
```
cloudhead, line 245: I think this is a good idea, I'm not a fan of optional parameters in rust using `Option`; we could take a slice of pushspecs instead :shrug:

radicle/src/storage.rs
cloudhead: This basically says that this remote belongs to a delegate of the project, ie. the public key is part of the project delegation. Does that make sense?

radicle/src/storage.rs
```diff
@@ -235,20 +241,32 @@ pub trait WriteStorage: ReadStorage {
...
     fn reference(
         &self,
         remote: &RemoteId,
-        reference: &RefStr,
+        reference: &Qualified,
     ) -> Result<Option<git2::Reference>, git2::Error>;
+
+    /// Get the [`git2::Commit`] found using its `oid`.
+    ///
+    /// Returns `None` if the commit did not exist.
```
cloudhead, line 263: Thanks for adding docs!

radicle/src/storage/refs.rs
```diff
@@ -18,7 +18,13 @@ use crate::git::Oid;
...
 use crate::storage::{ReadRepository, RemoteId, WriteRepository};
 
-pub static SIGNATURE_REF: Lazy<git::RefString> = Lazy::new(|| git::refname!("radicle/signature"));
+pub static SIGNATURE_REF: Lazy<git::Qualified> = Lazy::new(|| {
+    git::Qualified::from_components(
+        git::name::component!("radicle"),
+        git::name::component!("signature"),
+        None,
+    )
+});
```
cloudhead, line 27: WTB `git::qualified!`

radicle/src/storage/refs.rs
```diff
@@ -282,7 +288,7 @@ impl SignedRefs<Verified> {
             }
         }
 
-        let sigref = format!("refs/remotes/{remote}/{sigref}");
+        let sigref = sigref.add_namespace(remote.into());
```
cloudhead, line 291: Meta: another function name I think could be improved: `add` sounds like it's adding it to the end, I'd propose `with_namespace(remote)` or `namespaced(remote)`

radicle/src/git.rs
```diff
@@ -197,22 +204,76 @@ pub fn write_tree<'r>(
...
-    let push = format!("refs/heads/*:refs/remotes/{remote_id}/heads/*");
-    let remote = repo.remote_with_fetch(remote_name, remote_url.to_string().as_str(), &fetch)?;
-    repo.remote_add_push(remote_name, &push)?;
+    let fetch = format!("+refs/heads/*:refs/remotes/{name}/*");
+    let remote = repo.remote_with_fetch(name, url.to_string().as_str(), &fetch)?;
 
     Ok(remote)
 }
 
+// TODO: Would be nice if libgit2 respected namespaces https://github.com/libgit2/libgit2/issues/6414
```
cloudhead, line 224: I tend to not like having external URLs in repos because I hate broken/mutable links :grimacing:

cloudhead:
This is great, thanks. Mainly minor points. I am realizing though that you may have some conflicts with master due to the unwrap-ifying, but I can help with that.

radicle/src/git.rs
FintanH: Fair enough, I can create an issue instead to keep track of it.

radicle/src/storage.rs
FintanH: Ah, I think the comment threw me off with the use of "a project", but this `Remote` struct is just for a single project?

radicle/src/git.rs
FintanH: lol, that's how you know it's a good abstraction -- everything is immensely useful ;P

I'm not the biggest fan of this re-exporting style. The rust-analyzer shortcut for importing will always import from the crate itself, and I never knew what was re-exported from here or not.

FintanH:
> This is great, thanks. Mainly minor points. I am realizing though that you may have some conflicts with master due to the unwrap-ifying, but I can help with that.

No worries, I fixed everything up and I'll make issues for the namespace stuff in radicle-git!

radicle/src/storage.rs
cloudhead: Yeah, it's just for a single project. The refs it contains are scoped to a project/repo.

radicle/src/git.rs
cloudhead: Yeah I mean I have it imported as `fmt` above, so you could also do `fmt::lit`, `fmt::name` etc.

radicle/src/storage.rs
FintanH: Cool! I updated the doc string :)

cloudhead (approved):

radicle/src/git.rs
FintanH: Right, is that the standard to go by then or could I convince you to just import the crate where needed :eyes:? Would you want me to fix that up before merging?

PR #42, opened 2022-10-12
Node announcement handling

### LINKED ISSUE #24: Node announcement handling
To build an address book, peers communicate via gossip, sending `NodeAnnouncement` messages.

These need to be handled by storing the addresses in an address book (`address_book` module), and the addresses need to be persisted to disk, via an `sqlite` backend (same as routing table).

Node announcement messages should also be relayed to peers like other gossip messages (related: #17).

* [x] Storage
* [x] Relay
* [x] Message validation
* [ ] Test refs announcement relay logic

cloudhead:
Implements storage for #24 

What's left is still testing the relay functionality, as well as dropping old messages.


radicle-node/src/address/types.rs
```diff
@@ -0,0 +1,135 @@
+use std::ops::{Deref, DerefMut};
```
cloudhead, line 1: This file was mostly moved from address_book.rs.

PR #45, opened 2022-10-13
Correct handling of announcement timestamps

### LINKED ISSUE #44: Correct handling of announcement timestamps
While working on #42 I realized that our timestamp handling is inadequate due to tracking a single timestamp per peer. Since we now have three announcement message types, we need at least three timestamps per peer. But even that is not enough, since we don't want to conflate ref announcements for different projects with each other.

The solution is to also keep a map of timestamps per peer, for each tracked project. That way we are able to know when a ref announcement is interesting to us or not.

cloudhead:
Since we now have three announcement message types, we need at least three timestamps per peer. But even that is not enough, since we don't want to conflate ref announcements for different projects with each other.

The solution is to also keep a map of timestamps per peer, for each tracked project. That way we are able to know when a ref announcement is interesting to us or not.

Closes #44

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -568,19 +568,8 @@ where
 
     pub fn received_message(&mut self, addr: &net::SocketAddr, envelope: Envelope) {
         match self.handle_message(addr, envelope) {
-            Ok(relay) => {
-                if let Some(ann) = relay {
-                    let negotiated = self
-                        .sessions
-                        .negotiated()
-                        .filter(|(ip, _)| **ip != addr.ip())
-                        .map(|(_, p)| p);
-
-                    self.reactor.relay(ann, negotiated.clone());
-                }
-            }
             Err(SessionError::NotFound(ip)) => {
-                error!("Session not found for {}", ip);
+                error!("Session not found for {ip}");
             }
             Err(err) => {
                 // If there's an error, stop processing messages from this peer.
@@ -590,6 +579,7 @@ where
                 // FIXME: The peer should be set in a state such that we don'that
                 // process further messages.
             }
+            Ok(()) => {}
         }
     }
 
@@ -726,7 +716,7 @@ where
         &mut self,
         remote: &net::SocketAddr,
         envelope: Envelope,
-    ) -> Result<Option<Announcement>, peer::SessionError> {
+    ) -> Result<(), peer::SessionError> {
         let peer_ip = remote.ip();
         let peer = if let Some(peer) = self.sessions.get_mut(&peer_ip) {
             peer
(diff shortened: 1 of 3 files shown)


FintanH:
Had a play with your branch and came up with the below. Basically, moves the logic into methods on `Node` and uses `Update` (maybe `NodeUpdate` instead?) to see if we need to return early at the use sites. I opted to try it out because I thought it would be nicer to keep the announcement logic cleaner.

```
diff --git a/radicle-node/src/service.rs b/radicle-node/src/service.rs
index f9fe172..b044f08 100644
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -5,7 +5,6 @@ pub mod peer;
 pub mod reactor;
 pub mod routing;
 
-use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, HashMap};
 use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
@@ -621,9 +620,7 @@ where
             AnnouncementMessage::Inventory(message) => {
                 // Discard inventory messages we've already seen, otherwise update
                 // out last seen time.
-                if timestamp > peer.last_inventory {
-                    peer.last_inventory = timestamp;
-                } else {
+                if let Update::Stale = peer.last_inventory(timestamp) {
                     debug!("Ignoring stale inventory announcement from {node}");
                     return Ok(false);
                 }
@@ -651,21 +648,11 @@ where
                 if self.config.is_tracking(&message.id) {
                     // Discard inventory messages we've already seen, otherwise update
                     // out last seen time.
-                    match peer.last_refs.entry(message.id) {
-                        Entry::Vacant(e) => {
-                            e.insert(timestamp);
-                        }
-                        Entry::Occupied(mut e) => {
-                            let last = e.get_mut();
-
-                            if timestamp > *last {
-                                *last = timestamp;
-                            } else {
-                                debug!("Ignoring stale refs announcement from {node}");
-                                return Ok(false);
-                            }
-                        }
+                    if let Update::Stale = peer.last_ref(message.id, timestamp) {
+                        debug!("Ignoring stale refs announcement from {node}");
+                        return Ok(false);
                     }
+
                     // TODO: Check refs to see if we should try to fetch or not.
                     // FIXME: This code is wrong: we shouldn't be fetching from the connected peer,
                     // we should fetch from the origin.
@@ -1023,6 +1010,48 @@ pub struct Node {
     pub last_node: Timestamp,
 }
 
+pub enum Update {
+    Fresh,
+    Stale,
+}
+
+impl Node {
+    pub fn last_ref(&mut self, id: Id, t: Timestamp) -> Update {
+        let new = self
+            .last_refs
+            .entry(id)
+            .and_modify(|last| {
+                if t > *last {
+                    *last = t;
+                }
+            })
+            .or_insert(t);
+        if *new == t {
+            Update::Fresh
+        } else {
+            Update::Stale
+        }
+    }
+
+    pub fn last_inventory(&mut self, t: Timestamp) -> Update {
+        if t > self.last_inventory {
+            self.last_inventory = t;
+            Update::Fresh
+        } else {
+            Update::Stale
+        }
+    }
+
+    pub fn last_node(&mut self, t: Timestamp) -> Update {
+        if t > self.last_node {
+            self.last_node = t;
+            Update::Fresh
+        } else {
+            Update::Stale
+        }
+    }
+}
+
 #[derive(Debug)]
 /// Holds currently (or recently) connected peers.
 pub struct Sessions(AddressBook<IpAddr, Session>);
 ```

cloudhead:
Ah yeah, that's quite nice, will play around with it.

cloudhead:
Had to use the `Entry` match statement to handle the case where `timestamp == t` correctly. Also opted for `bool` for simplicity.

FintanH (approved):

PR #46, opened 2022-10-13
ssh: remove unused traits, simplify code

cloudhead:
We consolidate the Private, Public, and Encodable traits into one.


FintanH (approved):
LGTM

PR #50, opened 2022-10-16
Implement `git-remote-rad` helper

cloudhead:
This handles `rad://` remotes, by setting the namespace correctly.

I'm not sure how to test this yet... since I'd like to have `git-push` and `git-fetch` invoke it :thinking:, but I've tested it manually and it works.

Note that instead of verifying that the namespace matches a key in SSH-agent, we simply set the namespace to whatever the profile's public key is, unlike what the issue states.

Closes #32

### DIFF
--- a/radicle-remote-helper/src/lib.rs
+++ b/radicle-remote-helper/src/lib.rs
@@ -1,0 +1,203 @@
+#![allow(clippy::collapsible_if)]
+use std::path::PathBuf;
+use std::str::FromStr;
+use std::{env, io, process};
+
+use thiserror::Error;
+
+use radicle::crypto::{PublicKey, Signer};
+use radicle::node::Handle;
+use radicle::ssh;
+use radicle::storage::{ReadRepository, WriteStorage};
+
+/// The service invoked by git on the remote repository, during a push.
+const GIT_RECEIVE_PACK: &str = "git-receive-pack";
+
+#[derive(Debug, Error)]
+pub enum Error {
+    /// Remote repository not found (or empty).
+    #[error("remote repository `{0}` not found")]
+    RepositoryNotFound(PathBuf),
+    /// Secret key is not registered, eg. with ssh-agent.
+    #[error("public key `{0}` is not registered with ssh-agent")]
+    KeyNotRegistered(PublicKey),
+    /// Public key doesn't match the remote namespace we're pushing to.
+    #[error("public key `{0}` does not match remote namespace")]
+    KeyMismatch(PublicKey),
+    /// Invalid command received.
+    #[error("invalid command `{0}`")]
+    InvalidCommand(String),
+    /// Invalid arguments received.
+    #[error("invalid arguments: {0:?}")]
+    InvalidArguments(Vec<String>),
+    /// Error with the remote url.
+    #[error("invalid remote url: {0}")]
+    RemoteUrl(#[from] UrlError),
+}
+
+#[derive(Debug, Error)]
+pub enum UrlError {
+    /// Failed to parse.
+    #[error(transparent)]
+    Parse(#[from] radicle::git::url::parse::Error),
+    /// Unsupported URL scheme.
+    #[error("{0}: unsupported scheme: expected `rad://`")]
+    UnsupportedScheme(radicle::git::Url),
+    /// Missing host.
+    #[error("{0}: missing id")]
+    MissingId(radicle::git::Url),
+    /// Invalid remote repository identifier.
+    #[error("{0}: id: {1}")]
+    InvalidId(radicle::git::Url, radicle::identity::IdError),
+    /// Invalid public key.
+    #[error("{0}: key: {1}")]
+    InvalidKey(radicle::git::Url, radicle::crypto::PublicKeyError),
+}
+
+/// A git remote URL.
+///
+/// `rad://<id>/[<pubkey>]`
+///
+/// Eg. `rad://zUBDc1UdoEzbpaGcNXqauQkERJ8r` without the public key,
+/// and `rad://zUBDc1UdoEzbpaGcNXqauQkERJ8r/zCQTxdZGCzQXWBV3XbY3fgkHM3gfkLGyYMd2nL5R2MxQv` with.
+///
+#[derive(Debug)]
+pub struct Url {
+    pub id: radicle::identity::Id,
+    pub public_key: Option<PublicKey>,
+}
+
+impl FromStr for Url {
+    type Err = UrlError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let url: radicle::git::Url = s.as_bytes().try_into()?;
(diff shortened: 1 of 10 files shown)


radicle/src/node.rs
```diff
@@ -40,8 +42,9 @@ pub struct Connection {
 }
 
 impl Connection {
-    pub fn connect<P: AsRef<Path>>(path: P) -> Result<Self, io::Error> {
-        let stream = UnixStream::connect(path)?;
+    pub fn connect<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
+        let stream = UnixStream::connect(path)
+            .map_err(|e| Error::WithContext("failed to connect to node", e))?;
```
FintanH, line 47: Is this instead of the `io::Error`? I'd recommend doing:
```rust
pub enum Error {
  #[error("failed to connect to node: {0}")]
  Connect(#[source] io::Error),
}
```

I like this because we have specific errors for the scenarios that can "go wrong". I also wonder about adding the error in the display if it's the source. Aren't there ways to unravel those to get full displays in a pretty way?

radicle/src/node.rs
```diff
@@ -17,6 +17,8 @@ pub const DEFAULT_SOCKET_NAME: &str = "radicle.sock";
 pub enum Error {
     #[error("i/o error: {0}")]
     Io(#[from] io::Error),
```
FintanH, line 19: Is the `Io` variant still needed?

radicle-remote-helper/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    InvalidRemoteUrl(radicle::git::Url),
+    /// Invalid remote repository identifier.
+    InvalidRemoteId(radicle::git::Url, radicle::identity::IdError),
+}
+
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
+}
```
FintanH, line 33: Pretty sure you can just do the below because of default implementations:
```suggestion
impl std::error::Error for Error {}
```

radicle-remote-helper/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+
+impl std::fmt::Display for Error {
```
FintanH, line 35: Any reason not to use `thiserror`?

FintanH (commented):
> Note that instead of verifying that the namespace matches a key in SSH-agent, we simply set the namespace to whatever the profile's public key is, unlike what the issue states.

I might be wrong, but this doesn't allow a `git fetch` from another namespace? And also, if someone does a `git push other-remote` wouldn't that just push to _my_ remote even though that's not what I meant to do? Hence, the reason for checking the namespace and saying, "Hey, you're not authorised to do that".

radicle/src/node.rs
cloudhead: Yeah, makes sense.

cloudhead:
> I might be wrong, but this doesn't allow a `git fetch` from another namespace? And also, if someone does a `git push other-remote` wouldn't that just push to _my_ remote even though that's not what I meant to do? Hence, the reason for checking the namespace and saying, "Hey, you're not authorised to do that".

Hmm you're right about fetching, but you wouldn't be able to push to another remote anyway, since you don't have the key.

Need to think about this!

radicle-remote-helper/src/lib.rs
cloudhead: I was trying to avoid using anything unnecessary given this is a standalone binary, but since we depend on radicle anyway, I can just add it :+1:

FintanH: > I was trying to avoid using anything unnecessary given this is a standalone binary, but since we depend on radicle anyway, I can just add it +1

Heh, had a feeling it was that and came to the same conclusion :)

FintanH:
> Hmm you're right about fetching, but you wouldn't be able to push to another remote anyway, since you don't have the key.

Ya, exactly why I suggested we should check that the remote you're pushing to, i.e. the namespace, is owned by the you.

radicle-remote-helper/src/lib.rs
```diff
@@ -61,6 +65,16 @@ pub fn run(profile: radicle::Profile) -> Result<(), Box<dyn std::error::Error +
         .ok_or_else(|| Error::InvalidRemoteUrl(url.clone()))?
         .parse()
         .map_err(|e| Error::InvalidRemoteId(url.clone(), e))?;
+    let public_key: PublicKey = if url.path.is_empty() {
```
FintanH, line 68: Can I suggest we have a `Url` type in here to show what's going on a bit better and to have conversion traits? :)

cloudhead: You want to re-implement `Url` just for the remote helper? Not sure I understand.

FintanH: I'm thinking something like:
```rust
pub struct Url {
  id: radicle::identity::Id,
  key: Option<PublicKey>,
}

impl TryFrom<&str> for Url {
  type Error = Error;
  fn try_from(s: &str) -> Result<Self, Self::Error> {
    radicle::git::Url::try_from(s).and_then(Self::try_from)
  }
}

impl TryFrom<radicle::git::Url> for Url {
  type Error = Error;
  fn try_from(s: &radicle::git::Url) -> Result<Self, Self::Error> {
    /* parsing logic here */
  }
}
```

cloudhead: Ohh, gotcha, yes.

cloudhead: Done.

cloudhead:
Alright, this should be ready to go now @FintanH

radicle-remote-helper/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+#[derive(Debug, Error)]
+pub enum UrlError {
+    /// Failed to parse.
+    #[error(transparent)]
+    Parse(#[from] radicle::git::url::parse::Error),
+    /// Unsupported URL scheme.
+    #[error("{0}: unsupported scheme")]
+    UnsupportedScheme(radicle::git::Url),
```
FintanH, line 45: Might be nice to report the scheme that it's expecting

radicle-remote-helper/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// Missing host.
+    #[error("{0}: missing id")]
+    MissingId(radicle::git::Url),
+    /// Invalid remote repository identifier.
+    #[error("{0}: id: {1}")]
+    InvalidId(radicle::git::Url, radicle::identity::IdError),
+    /// Invalid remote repository identifier.
```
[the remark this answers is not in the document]
cloudhead: woopsidoodlez

PR #52, opened 2022-10-17
Fix DID key format

cloudhead:
See commit.


FintanH (approved):
Any reason you didn't opt to use https://docs.rs/did-key/0.1.1/did_key?

cloudhead:
First time I see that, but looking at the dependencies, it's tied to a different key implementation anyway.

FintanH:
> First time I see that, but looking at the dependencies, it's tied to a different key implementation anyway.

Right, good point :)

PR #54, opened 2022-10-18
Enforce max size for `Message` type on wire

### LINKED ISSUE #21: Make sure all `wire` messages fit with `64KB`
Due to the planned noise protocol framing, we need to ensure that all messages sent between peers fit within 64KB.

We should audit the code (mainly the `wire` module), find a strategy for large messages (eg. splitting them or limiting the amount of data that can go in), and if possible, add some assertions to catch any messages that are bigger than 64KB.

cloudhead:
Closes #21

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -31,7 +31,7 @@ use crate::git::Url;
 use crate::identity::{Doc, Id};
 use crate::node;
 use crate::service::config::ProjectTracking;
-use crate::service::message::{Address, Announcement, AnnouncementMessage};
+use crate::service::message::{Address, Announcement, AnnouncementMessage, Ping};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
 use crate::service::peer::{PingState, SessionError, SessionState};
 use crate::storage;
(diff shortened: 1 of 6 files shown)


radicle-node/src/service.rs
```diff
@@ -834,11 +834,17 @@ where
...
                 return Err(SessionError::Misbehavior);
             }
-            (SessionState::Negotiated { .. }, Message::Ping { ponglen, .. }) => {
-                let resp = Message::Pong {
-                    zeroes: ZeroBytes::new(ponglen),
-                };
-                self.reactor.write(peer.addr, resp);
+            (SessionState::Negotiated { .. }, Message::Ping(Ping { ponglen, .. })) => {
+                // Ignore pings which ask for too much data.
+                if ponglen > Ping::MAX_PONG_ZEROES {
```
FintanH, line 839: A debug or trace would likely be helpful if we ever end up needing to debug why a ping was ignored

FintanH (approved):

radicle-node/src/service.rs
cloudhead: Lol I had one there and removed it because I thought it was too niche; will add it back.

PR #57, opened 2022-10-19
Set repository canonical refs

cloudhead:
We set the `HEAD` and eg. `refs/heads/master` refs when updating repositories, which allows users to easily clone with git.

For now, `HEAD` is chosen based on a unanimity quorum of delegates. This can be made more sophisticated in the future.

We've also had to ignore some refs during verification, since these have cropped in for eg. due to the staging copy clone and the changes included here.

### DIFF
--- a/radicle-remote-helper/src/lib.rs
+++ b/radicle-remote-helper/src/lib.rs
@@ -8,7 +8,7 @@ use thiserror::Error;
 use radicle::crypto::{PublicKey, Signer};
 use radicle::node::Handle;
 use radicle::ssh;
-use radicle::storage::{ReadRepository, WriteStorage};
+use radicle::storage::{ReadRepository, WriteRepository, WriteStorage};
 
 /// The service invoked by git on the remote repository, during a push.
 const GIT_RECEIVE_PACK: &str = "git-receive-pack";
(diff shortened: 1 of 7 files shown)


FintanH (approved):
This makes sense. Something I've been thinking about is how does another peer verify that the quorum is legit without calculating `head` all over again.

I don't have a solid answer but would like to put it on our radar.

cloudhead:
Ah, so right now the canonical refs of a peer are ignored/discarded, and each peer re-computes their own. You're saying there could be an efficiency gain in verifying existing refs without having to recompute them from scratch?

FintanH:
> Ah, so right now the canonical refs of a peer are ignored/discarded, and each peer re-computes their own. You're saying there could be an efficiency gain in verifying existing refs without having to recompute them from scratch?

Ya, exactly. Something along the lines of the signed refs approach but with a combination of delegate signatures.

Another thought just occurred too. Should we detect forks as well?

cloudhead:
> Another thought just occurred too. Should we detect forks as well?

How would you do that? Divergence at the tips is expected..

FintanH:
> How would you do that? Divergence at the tips is expected..

Mmmm right. So there's always a merge base I suppose?

cloudhead:
> > How would you do that? Divergence at the tips is expected..
> 
> Mmmm right. So there's always a merge base I suppose?

Yeah, unless the first commit is not even shared, but then it's not even a fork, it's just the wrong project :D

FintanH:
> > > How would you do that? Divergence at the tips is expected..
> > 
> > 
> > Mmmm right. So there's always a merge base I suppose?
> 
> Yeah, unless the first commit is not even shared, but then it's not even a fork, it's just the wrong project :D

Right :sweat_smile: I guess it might be interesting to be able to tell how far ahead or behind you are from the latest. But that can be in another area of the UX anyway :)

PR #58, opened 2022-10-19
httpd: implement `git` fetch-only server

cloudhead:
Introduces a new `radicle-httpd` crate which will serve as the basis for heartwood's HTTP backend.

The first part of this is a git server which allows projects to be fetched from storage using plain `git`.

Most of this code was copied from `radicle-client-services/git-server`, and heavily stripped down.

Depends on #57 

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,23 @@
+use std::process::Command;
+
+fn main() {
+    // Set a build-time `GIT_HEAD` env var which includes the commit id;
+    // such that we can tell which code is running.
+    let hash = Command::new("git")
+        .arg("rev-parse")
+        .arg("--short")
+        .arg("HEAD")
+        .output()
+        .ok()
+        .and_then(|output| {
+            if output.status.success() {
+                String::from_utf8(output.stdout).ok()
+            } else {
+                None
+            }
+        })
+        .unwrap_or_else(|| String::from("unknown"));
+
+    println!("cargo:rustc-env=GIT_HEAD={}", hash);
+    println!("cargo:rustc-rerun-if-changed=.git/HEAD");
+}
--- a/radicle-httpd/Cargo.toml
+++ b/radicle-httpd/Cargo.toml
@@ -1,0 +1,33 @@
+[package]
+name = "radicle-httpd"
+license = "MIT OR Apache-2.0"
+version = "0.1.0"
+authors = ["Alexis Sellier <alexis@radicle.xyz>"]
+edition = "2021"
+default-run = "radicle-httpd"
+build = "../build.rs"
+
+[features]
+default = []
+logfmt = [
+  "tracing-logfmt",
+  "tracing-subscriber/env-filter"
+]
+
+[dependencies]
+anyhow = { version = "1" }
+flate2 = { version = "1" }
+lexopt = { version = "0.2.1" }
+thiserror = { version = "1" }
+tokio = { version = "1.21", default-features = false, features = ["macros", "rt-multi-thread"] }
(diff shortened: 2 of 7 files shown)


build.rs
```diff
@@ -0,0 +1,21 @@
+use std::process::Command;
+
+fn main() {
```
FintanH, line 3: What's this for? :thinking:

cloudhead: Setting a build-time `GIT_HEAD` env var, so that we can include it in the logs when starting the server.

FintanH: Cool, thanks for explaining :)

radicle-httpd/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+            .with(tracing_logfmt::layer())
+    }
+}
+
+#[cfg(not(feature = "logfmt"))]
+mod logger {
+    pub fn subscriber() -> impl tracing::Subscriber {
+        tracing_subscriber::FmtSubscriber::new()
+    }
+}
```
slack-coder, line 23: Curious, what's the purpose for the logfmt feature?

I'm also a little concerned about the long term maintenance burden to support different features.

radicle-httpd/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+tracing-subscriber = { version = "0.3", default-features = false, features = ["std", "ansi", "fmt"] }
+tracing-logfmt = { version = "0.2", optional = true }
+axum = { version = "0.5.16", default-features = false }
+axum-server = { version = "0.4.2", default-features = false }
+hyper = { version = "0.14.17", default-features = false }
+tower-http = { version = "0.3.4", default-features = false, features = ["trace"] }
+
+[dependencies.radicle]
+path = "../radicle"
+version = "0.2.0"
```
slack-coder, line 33: Side note, and not directly related to these changes.

Why specify the patch version in the Cargo file?  Specially for versions with patch level '0'.  [Cargo's spec](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) define version's with a patch level as lower bounds.  It will resolve '0.2.0' to the highest patch level, maybe '0.2.9' for instance.

Specifying just the major and minor version number would remove redundant information, and may be best to include it only when necessary.  For example using 'axum-server' version '0.4.*' (or 0.4) and leaving Cargo to choose the best patch version.

radicle-httpd/src/main.rs
cloudhead: It's what we use when we want the logs to be parsed for grafana dashboards for example.

radicle-httpd/Cargo.toml
cloudhead: Hmm, in the above case it's not necessary, but I thought that eg. `0.2.1` would select `0.2.1` and above, vs. `0.2` which might select `0.2.0`? Maybe I'm wrong, I find semver really confusing.

slack-coder: > ... but I thought that eg. `0.2.1` would select `0.2.1` and above, vs. `0.2` which might select `0.2.0`? ...

That's correct. Though since cargo selects the best possible patch version, why include them unless it affects builds?  We'll avoid maintaining minor patch versions in the Cargo.toml.

cloudhead: Ah gotcha, yeah it can be omitted; but it might be useful to know if it's not selecting the latest..

cloudhead:
Merging this, but still happy to receive comments on it.

PR #59, opened 2022-10-19
Fix various issues

cloudhead:
Fixing various TODOs and long standing issues.

Might be easier to review commit-by-commit.

### DIFF
--- a/radicle-node/src/control.rs
+++ b/radicle-node/src/control.rs
@@ -4,7 +4,7 @@ use std::io::BufReader;
 use std::io::LineWriter;
 use std::os::unix::net::UnixListener;
 use std::os::unix::net::UnixStream;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::{fs, io, net};
 
 use crate::client;
@@ -17,12 +17,22 @@ use crate::service::FetchResult;
 pub enum Error {
     #[error("failed to bind control socket listener: {0}")]
     Bind(io::Error),
+    #[error("invalid socket path specified: {0}")]
+    InvalidPath(PathBuf),
 }
 
 /// Listen for commands on the control socket, and process them.
 pub fn listen<P: AsRef<Path>, H: Handle>(path: P, handle: H) -> Result<(), Error> {
     // Remove the socket file on startup before rebinding.
     fs::remove_file(&path).ok();
+    fs::create_dir_all(
+        path.as_ref()
+            .parent()
+            .ok_or_else(|| Error::InvalidPath(path.as_ref().to_path_buf()))?,
+    )
+    .ok();
+
+    log::info!("Binding control socket {}..", path.as_ref().display());
 
     let listener = UnixListener::bind(path).map_err(Error::Bind)?;
     for incoming in listener.incoming() {
--- a/radicle/src/storage.rs
+++ b/radicle/src/storage.rs
@@ -41,8 +41,6 @@ pub enum Error {
     Id(#[from] IdError),
     #[error("i/o: {0}")]
     Io(#[from] io::Error),
-    #[error("invalid repository head")]
-    InvalidHead,
 }
 
 /// Fetch error.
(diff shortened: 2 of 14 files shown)


radicle-node/src/control.rs
```diff
@@ -17,12 +17,22 @@ use crate::service::FetchResult;
 pub enum Error {
     #[error("failed to bind control socket listener: {0}")]
     Bind(io::Error),
+    #[error("invalid socket path specified: {0}")]
```
FintanH, line 20: This is invalid because the directory doesn't exist, right? Maybe it would be more helpful to state that to help the end user remedy the situation.

`<tangent>` I wish there was some mechanism for writing copy for errors (even templates) that you could link to from the `error` macro. That way you could have really informative errors without cluttering the code itself. Dhall does iirc and I quite liked that approach `</tangent>`

radicle/src/storage/git.rs
```diff
@@ -400,14 +368,33 @@ impl Repository {
         Ok(remotes)
     }
 
-    pub fn sign_refs<G: Signer>(&self, signer: G) -> Result<SignedRefs<Verified>, Error> {
-        let remote = signer.public_key();
-        let refs = self.references(remote)?;
-        let signed = refs.signed(&signer)?;
+    /// Return all references that are public, ie. that are replicated, verified and signed.
```
FintanH, line 371: nit: The public doesn't fit right with me. It's more that you're returning a filtered set of `namespaced_references`, right? The canonical references are also public per se.

radicle/src/storage/git/transport.rs
```diff
@@ -102,7 +99,8 @@ impl git2::transport::SmartSubtransport for Smart {
             Ok(stream)
         } else {
             Err(git2::Error::from_str(&format!(
-                "repository {id} does not have an associated stream"
+                "repository {} does not have an associated stream",
```
FintanH, line 102: `<rant>` I can't believe they didn't allow for `{url.id}` as part of that improvement `</rant>`

radicle/src/storage.rs
```diff
@@ -250,10 +242,20 @@ pub trait ReadRepository {
 
     fn blob_at<'a>(&'a self, oid: Oid, path: &'a Path) -> Result<git2::Blob<'a>, git_ext::Error>;
 
-    /// Get the canonical head of this repository.
+    /// Get the head of this repository.
+    ///
+    /// Returns the value of `HEAD` if it is set, otherwise computes the canonical head
```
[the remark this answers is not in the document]
cloudhead: Hehe yes, ok; though it's `HEAD` not `refs/HEAD` ;)

radicle-node/src/control.rs
FintanH: > Actually no, it's if the path doesn't have a parent, ie. I think it'll only happen if `/` is passed in :D

Ok, then maybe we should say that :smile: `"the socket path '{0}' must have a parent, i.e. it should not be '/'"`

> Yeah, tbh I'm still generally unsatisfied with rust errors; I still find myself getting the dreaded `File not found (os error 2)` sometimes with no other context and I'm like shrug

The wooooooorst

radicle/src/storage/git/transport.rs
cloudhead: I know, when I realized that I was like "oh, so actually this won't change my life".

radicle/src/storage/git.rs
cloudhead: 5f4fc8236ee08ae279a1e09f033bfb5acddcf607

radicle/src/storage.rs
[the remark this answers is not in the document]
cloudhead: 5f4fc8236ee08ae279a1e09f033bfb5acddcf607

FintanH (approved):

radicle-node/src/control.rs
cloudhead: > Ok, then maybe we should say that smile `"the socket path '{0}' must have a parent, i.e. it should not be '/'"`

I almost feel like returning an easter egg if this happens :thinking:

PR #60, opened 2022-10-19
Port Collaborative Objects

FintanH:
An _**extremely**_ rough draft of porting over the cob code. It's compiling but needs some cleanup and testing, testing, testing

TODO:
- [x] Move signatures from trailers to git gpgsig headers
- [x]  Update documentation
  - [x] Would be good to have docs for `create`, `get`, `update`, and `list` 
  - [x] Change `lib.rs` docs
  - [x] Add docs for other traits
- [x] Clean up traits
  - [x] Storage traits unified
  - [x] Explore different identity-based trait setups
- [x] Add testing
  - [x] `create`
  - [x] `get`
  - [x] `list`
  - [x] `update`

### DIFF
--- a/radicle-cob/Cargo.toml
+++ b/radicle-cob/Cargo.toml
@@ -1,0 +1,47 @@
+[package]
+name = "radicle-cob"
+version = "0.1.0"
+authors = [
+  "Alex Good <alex@memoryandthought.me>",
+  "Fintan Halpenny <fintan.halpenny@gmail.com>",
+]
+edition = "2021"
+license = "GPL-3.0-or-later"
+description = "Library for implementing Radicle Collaborative Objects"
+keywords = ["radicle", "collaborative objects", "cob", "cobs"]
+
+[lib]
+
+[dependencies]
+git-commit = { version = "0.2" }
+git-ref-format = { version = "0.1" }
+git-trailers = { version = "0.1" }
+log = { version = "0.4.17" }
+petgraph = { version = "0.5" }
+radicle-git-ext = { version = "0.2" }
+serde_json = { version = "1.0" }
+thiserror = { version = "1.0" }
+
+[dependencies.git2]
+version = "0.15.0"
+default-features = false
+features = ["vendored-libgit2"]
+
+[dependencies.radicle-crypto]
+path = "../radicle-crypto"
+version = "0.1"
+features = ["ssh"]
+
+[dependencies.serde]
+version = "1.0"
+features = ["derive"]
+
+[dev-dependencies]
+ed25519-compact = { version = "1.0.12", features = ["pem"] }
+tempfile = { version = "3" }
+quickcheck = { version = "1", default-features = false }
+
+[dev-dependencies.radicle-crypto]
+path = "../radicle-crypto"
+version = "0.1"
+features = ["test"]
--- a/radicle-cob/src/backend/git/change.rs
+++ b/radicle-cob/src/backend/git/change.rs
@@ -1,0 +1,284 @@
+// Copyright © 2022 The Radicle Link Contributors
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+use std::convert::TryFrom;
+
+use git_commit::{self as commit, Commit};
+use git_ext::Oid;
+use git_trailers::OwnedTrailer;
+
+use crate::{
+    change::{self, store, Change},
+    history::entry,
+    signatures::{Signature, Signatures},
+    trailers, HistoryType,
+};
+
+const MANIFEST_BLOB_NAME: &str = "manifest";
+const CHANGE_BLOB_NAME: &str = "change";
+
+pub mod error {
+    use std::str::Utf8Error;
+    use std::string::FromUtf8Error;
+
+    use git_ext::Oid;
+    use git_trailers::Error as TrailerError;
+    use thiserror::Error;
+
+    use crate::signatures::error::Signatures;
+    use crate::trailers;
+
+    #[derive(Debug, Error)]
+    pub enum Create {
+        #[error(transparent)]
+        FromUtf8(#[from] FromUtf8Error),
+        #[error(transparent)]
+        Git(#[from] git2::Error),
+        #[error(transparent)]
+        Signer(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
+        #[error(transparent)]
+        Utf8(#[from] Utf8Error),
+    }
+
+    #[derive(Debug, Error)]
+    pub enum Load {
+        #[error(transparent)]
+        Read(#[from] git_commit::error::Read),
+        #[error(transparent)]
+        Signatures(#[from] Signatures),
+        #[error(transparent)]
+        Git(#[from] git2::Error),
+        #[error("a 'manifest' file was expected be found in '{0}'")]
+        NoManifest(Oid),
+        #[error("the 'manifest' found at '{0}' was not a blob")]
+        ManifestIsNotBlob(Oid),
+        #[error("the 'manifest' found at '{id}' was invalid: {err}")]
+        InvalidManifest {
+            id: Oid,
+            #[source]
+            err: serde_json::Error,
+        },
+        #[error("a 'change' file was expected be found in '{0}'")]
+        NoChange(Oid),
+        #[error("the 'change' found at '{0}' was not a blob")]
+        ChangeNotBlob(Oid),
+        #[error(transparent)]
+        AuthorTrailer(#[from] trailers::error::InvalidAuthorTrailer),
+        #[error(transparent)]
+        ResourceTrailer(#[from] super::trailers::error::InvalidResourceTrailer),
+        #[error("non utf-8 characters in commit message")]
+        Utf8(#[from] FromUtf8Error),
+        #[error(transparent)]
+        Trailer(#[from] TrailerError),
+    }
+}
+
+impl change::Storage for git2::Repository {
+    type CreateError = error::Create;
+    type LoadError = error::Load;
+
+    type ObjectId = Oid;
+    type Author = Oid;
+    type Resource = Oid;
+    type Signatures = Signatures;
+
+    fn create<Signer>(
+        &self,
+        author: Option<Self::Author>,
+        resource: Self::Resource,
+        signer: &Signer,
+        spec: store::Create<Self::ObjectId>,
+    ) -> Result<Change, Self::CreateError>
+    where
+        Signer: crypto::Signer,
+    {
+        let change::Create {
+            typename,
+            tips,
+            message,
+            contents,
+        } = spec;
+        let manifest = store::Manifest {
+            typename,
+            history_type: (&contents).into(),
+        };
+
+        let revision = write_manifest(self, &manifest, &contents)?;
+        let tree = self.find_tree(revision)?;
+
+        let signature = {
+            let sig = signer.sign(revision.as_bytes());
+            let key = signer.public_key();
+            Signature::from((*key, sig))
+        };
+
+        let id = write_commit(
+            self,
+            author,
+            resource,
+            tips,
+            message,
+            signature.clone(),
+            tree,
+        )?;
+        Ok(Change {
+            id,
+            revision: revision.into(),
+            signatures: signature.into(),
+            author,
+            resource,
+            manifest,
+            contents,
+        })
+    }
+
+    fn load(&self, id: Self::ObjectId) -> Result<Change, Self::LoadError> {
+        let commit = Commit::read(self, id.into())?;
+        let (author, resource) = parse_trailers(commit.trailers())?;
+        let signatures = Signatures::try_from(&commit)?;
+
+        let tree = self.find_tree(commit.tree())?;
+        let manifest = load_manifest(self, &tree)?;
+        let contents = load_contents(self, &tree, &manifest)?;
+
+        Ok(Change {
+            id,
+            revision: tree.id().into(),
+            signatures,
+            author,
+            resource,
+            manifest,
+            contents,
+        })
+    }
+}
+
+fn parse_trailers<'a>(
+    mut trailers: impl Iterator<Item = &'a OwnedTrailer>,
+) -> Result<(Option<Oid>, Oid), error::Load> {
+    let (author, resource) = trailers.try_fold((None, None), |(author, resource), trailer| {
+        match trailers::AuthorCommitTrailer::try_from(trailer) {
+            Ok(trailer) => Ok((Some(trailer.oid().into()), resource)),
+            Err(err) => match err {
+                trailers::error::InvalidAuthorTrailer::NoTrailer
+                | trailers::error::InvalidAuthorTrailer::NoValue => Ok((author, resource)),
+                trailers::error::InvalidAuthorTrailer::WrongToken => {
+                    let resource = trailers::ResourceCommitTrailer::try_from(trailer)?;
+                    Ok((author, Some(resource.oid().into())))
+                }
+                err => Err(error::Load::from(err)),
+            },
+        }
+    })?;
+    let resource = resource
+        .ok_or_else(|| error::Load::from(trailers::error::InvalidResourceTrailer::NoTrailer))?;
+    Ok((author, resource))
+}
+
+fn load_manifest(
+    repo: &git2::Repository,
+    tree: &git2::Tree,
+) -> Result<store::Manifest, error::Load> {
+    let manifest_tree_entry = tree
+        .get_name(MANIFEST_BLOB_NAME)
+        .ok_or_else(|| error::Load::NoManifest(tree.id().into()))?;
+    let manifest_object = manifest_tree_entry.to_object(repo)?;
+    let manifest_blob = manifest_object
+        .as_blob()
+        .ok_or_else(|| error::Load::ManifestIsNotBlob(tree.id().into()))?;
+    serde_json::from_slice(manifest_blob.content()).map_err(|err| error::Load::InvalidManifest {
+        id: tree.id().into(),
+        err,
+    })
+}
+
+fn load_contents(
+    repo: &git2::Repository,
+    tree: &git2::Tree,
+    manifest: &store::Manifest,
+) -> Result<entry::Contents, error::Load> {
+    Ok(match manifest.history_type {
+        HistoryType::Automerge => {
+            let contents_tree_entry = tree
+                .get_name(CHANGE_BLOB_NAME)
+                .ok_or_else(|| error::Load::NoChange(tree.id().into()))?;
+            let contents_object = contents_tree_entry.to_object(repo)?;
+            let contents_blob = contents_object
+                .as_blob()
+                .ok_or_else(|| error::Load::ChangeNotBlob(tree.id().into()))?;
+            entry::Contents::automerge(contents_blob.content())
+        }
+    })
+}
+
+fn write_commit<O>(
+    repo: &git2::Repository,
+    author: Option<O>,
+    resource: O,
+    tips: Vec<O>,
+    message: String,
+    signature: Signature,
+    tree: git2::Tree,
+) -> Result<Oid, error::Create>
+where
+    O: AsRef<git2::Oid>,
+{
+    let author = author.map(|author| *author.as_ref());
+    let resource = *resource.as_ref();
+
+    let mut parents = tips.iter().map(|o| *o.as_ref()).collect::<Vec<_>>();
+    parents.push(resource);
+    parents.extend(author);
+
+    let mut trailers: Vec<OwnedTrailer> =
+        vec![trailers::ResourceCommitTrailer::from(resource).into()];
+    trailers.extend(author.map(|author| trailers::AuthorCommitTrailer::from(author).into()));
+
+    {
+        let author = repo.signature()?;
+        let mut headers = commit::Headers::new();
+        headers.push(
+            "gpgsig",
+            &String::from_utf8(crypto::ssh::ExtendedSignature::from(signature).to_armored())?,
+        );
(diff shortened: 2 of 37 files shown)


radicle-cob/src/change_graph/evaluation.rs
```diff
@@ -0,0 +1,153 @@
...
+
+enum RejectionReason {
+    InvalidSignatures,
+    MissingAuthor {
+        missing_author_oid: Oid,
+    },
+    ErrorFindingAuthor {
+        author_commit_oid: Oid,
+        error: Box<dyn std::error::Error>,
+    },
```
cloudhead, line 121: Maybe these author-related errors aren't needed anymore if we're just verifying a signature with a key?

radicle-cob/src/sign.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+impl<T> Key for T
+where
+    Self: PublicKey + ToString + Ord,
+    Self::Signature: ToString,
+    for<'de> Self: serde::Deserialize<'de>,
+    for<'de> Self::Signature: serde::Deserialize<'de>,
+{
+}
```
cloudhead, line 45: I wonder if it makes sense to have a `radicle-crypto` crate, that `radicle-cob` can depend on, so that we don't need generic key types. It would simplify things a lot.

radicle-cob/src/sign/signatures.rs
```diff
@@ -0,0 +1,291 @@
...
+    fn try_from(Trailer { values, .. }: Trailer) -> Result<Self, Self::Error> {
+        let mut iter = values.iter().flat_map(|val| val.split_whitespace());
+
+        let key = iter
+            .next()
+            .ok_or(error::Signature::Missing("public key"))
+            .and_then(|key| {
+                K::deserialize(
+                    key.deref().into_deserializer() as StrDeserializer<serde::de::value::Error>
+                )
```
cloudhead, line 64: Are we using json inside commit messages? :grimacing:

radicle-cob/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+regex = "1.5.5"
+serde_json = "1.0"
+thiserror = "1.0"
+toml = "0.5"
+tracing = "0.1"
+either = "1.6"
+multihash = "0.11"
+multibase = "0.9"
+lazy_static = "1.4"
+tempfile = "3.3"
```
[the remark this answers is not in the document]
FintanH: > `git-trailers` -> if this is just for signatures, we have a function in `storage::git::trailers` that does the job, we also have commit header (gpgsig) parsing in radicle-ssh

It's also needed to tag the author and "authority" in the commit though. As well as that, `radicle-cob` isn't depending on `radicle`, it'll be the other way around.

Happy to move to the gpgsig version, but wanted to get to a place where I had stuff working first :)

>  regex -> very curious why we need regex for a low level library such as this

It's used for `TypeName` validation.

>  toml -> can we use json, since we're already importing serde_json? our other docs (eg. identity) are all json

Sure, happy to do that. Maybe @alexjg can remind us why he opted for toml in the first place just to clarify first.

>  tracing -> not sure this is relevant without async, I think log should work

Yup, sounds good. Was planning on switching since I saw you used `log` everywhere else.

>  either -> probably not needed anymore, as it was used for direct/indirect delegations if I remember

You're right, it can be removed!

>  lazy_static -> we should replace with once_cell::Lazy

:+1:

radicle-cob/src/test/crypto.rs
```diff
@@ -0,0 +1,229 @@
...
+    type Key = PublicKey;
+
+    fn sign(&self, bytes: &[u8]) -> Result<Self::Signature, Self::Error> {
+        Ok(Signature(ed25519::SecretKey::sign(self, bytes, None)))
+    }
+
+    fn public_key(&self) -> Self::Key {
+        PublicKey(ed25519::SecretKey::public_key(self))
+    }
+}
```
cloudhead, line 229: All this would also go away with a `radicle-crypto` crate

radicle-cob/src/change_graph/evaluation.rs
FintanH: So I think we still need to _find_ a key. Which is something we need to think about. At the moment we're not content addressing the `did:key`, right? But that probably won't be the case in the future if we need a history of the keys.

radicle-cob/src/sign.rs
FintanH: > I wonder if it makes sense to have a `radicle-crypto` crate, that `radicle-cob` can depend on, so that we don't need generic key types. It would simplify things a lot.

My guy... I was asking this here https://discord.com/channels/841318878125490186/880000282388475925/1031513523588706355 :smile:

radicle-cob/src/sign/signatures.rs
FintanH: No, it's just the strings :) serde does not imply json ;)

radicle-cob/src/test/crypto.rs
FintanH: Yes, well I'm glad we got on the same page. I'll definitely move it out :joy:

radicle-cob/src/type_name.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+impl fmt::Display for TypeName {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.0.as_str())
+    }
+}
+
+lazy_static::lazy_static! {
+    static ref TYPENAME_REGEX: regex::Regex = regex::Regex::new(r"^([a-zA-Z0-9])+(\.[a-zA-Z0-9]+)*$").unwrap();
```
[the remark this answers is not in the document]
FintanH: But using what? Just like basic `str` functionality?

radicle-cob/src/sign.rs
cloudhead: Aha I see, yes, the answer is yes :)

radicle-cob/src/sign/signatures.rs
cloudhead: Ah, so deserializing strings into some type T via the Deserialize instance? Perhaps we can use FromStr/ToString here if I understand correctly.

radicle-cob/src/type_name.rs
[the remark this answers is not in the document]
cloudhead: Yeah, it's just `split('.')` and `part.all(|c| c.is_ascii_lphanumeric())` on each part, more or less

radicle-cob/src/change_graph/evaluation.rs
FintanH: So `Change` has fields for `author` and `authorizing` which are the content identifiers for the identities of who made the change and the parent identity. So those need to be looked up as commits in the internal logic.

Perhaps you're suggesting that the `author` should be removed? But that doesn't really leave room for tagging the change with metadata.

radicle-cob/src/sign/signatures.rs
FintanH: It doesn't matter now that we've agreed on `radicle-crypto` ;)

radicle-cob/src/type_name.rs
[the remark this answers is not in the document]
FintanH: Aye, I'll make the change :+1:

radicle-cob/src/change_graph/evaluation.rs
FintanH: Oh well there's a `signatures` field.

> author: author information; keep this, but it's not used for validation

I'll look into this

radicle-cob/src/sign/signatures.rs
cloudhead: Oh true.

radicle-cob/src/change_graph/evaluation.rs
cloudhead: Ok, if the signatures field includes the pubkey, that should work.

FintanH: Hmmmm so I think what you're looking for is that the author and authorizing should be optional, would that be right? These rejections would go away if it was ok to not find an author in the git backend.

FintanH:
This is almost ready for final review. The last things I wanna fix up are:
* Use gpgsig for the signatures
* Resolve whether authors and authorities need to be optional https://github.com/radicle-dev/heartwood/pull/60#discussion_r1008270870
* Update the `lib.rs` doc string to represent the new state of matters.

radicle-cob/src/change_graph/evaluation.rs
cloudhead: Well, the author is optional because it's not needed for verification. It should just be a `did` or something.
The authorizing identity is the project under which the cob lives, so I think that's probably needed?

cloudhead:
> Use gpgsig for the signatures

Note that we aren't using this yet for the identity branch, though the code for parsing and rendering gpgsig signatures should all be there:

https://github.com/radicle-dev/heartwood/blob/master/radicle/src/git.rs#L504
https://github.com/radicle-dev/heartwood/blob/master/radicle-crypto/src/ssh.rs#L269

Only issue is the former is under the `radicle` crate, so perhaps you want to move that to `radicle-git`?

radicle-cob/src/change_graph/evaluation.rs
FintanH: >  The authorizing identity is the project under which the cob lives, so I think that's probably needed?

So before we were carrying it to do the authorization logic within this library. But I've shifted the concerns to the end user -- i.e. the authority's content id is stored in the history and they can choose to fetch that identity and do validation in the `traverse`. So it's not necessarily needed, but we can still make it mandatory.

FintanH:
> Only issue is the former is under the radicle crate, so perhaps you want to move that to radicle-git?

Ya, I was thinking I'd end up doing that :+1:

radicle-cob/src/change_graph/evaluation.rs
cloudhead: Ah gotcha, but this sounds like it's redundant? I would think that only one authorizing identity is needed per change graph?

FintanH: The identity can change over time, so you need to keep the current snapshot of that identity -- for checking what keys are in that snapshot, for example.

FintanH:
Oh, one annoying thing about the gpgsig is that it doesn't have the public key that was used to make the signature :( But I guess that can become a field in the `Change`.

cloudhead:
> Oh, one annoying thing about the gpgsig is that it doesn't have the public key that was used to make the signature :( But I guess that can become a field in the `Change`.

It does though?

FintanH:
> > Oh, one annoying thing about the gpgsig is that it doesn't have the public key that was used to make the signature :( But I guess that can become a field in the `Change`.
> 
> It does though?

```
-----BEGIN SSH SIGNATURE-----
U1NIU0lHAAAAAQAAADMAAAALc3NoLWVkMjU1MTkAAAAgvjrQogRxxLjzzWns8+mKJAGzEX
4fm2ALoN7pyvD2ttQAAAADZ2l0AAAAAAAAAAZzaGE1MTIAAABTAAAAC3NzaC1lZDI1NTE5
AAAAQIQvhIewOgGfnXLgR5Qe1ZEr2vjekYXTdOfNWICi6ZiosgfZnIqV0enCPC4arVqQg+
GPp0HqxaB911OnSAr6bwU=
-----END SSH SIGNATURE-----
```

Which part is the key and which part is the signature?

cloudhead:
> > > Oh, one annoying thing about the gpgsig is that it doesn't have the public key that was used to make the signature :( But I guess that can become a field in the `Change`.
> > 
> > 
> > It does though?
> 
> ```
> -----BEGIN SSH SIGNATURE-----
> U1NIU0lHAAAAAQAAADMAAAALc3NoLWVkMjU1MTkAAAAgvjrQogRxxLjzzWns8+mKJAGzEX
> 4fm2ALoN7pyvD2ttQAAAADZ2l0AAAAAAAAAAZzaGE1MTIAAABTAAAAC3NzaC1lZDI1NTE5
> AAAAQIQvhIewOgGfnXLgR5Qe1ZEr2vjekYXTdOfNWICi6ZiosgfZnIqV0enCPC4arVqQg+
> GPp0HqxaB911OnSAr6bwU=
> -----END SSH SIGNATURE-----
> ```
> 
> Which part is the key and which part is the signature?

well you have to parse it with `ExtendedSignature` :D

radicle-cob/src/change_graph/evaluation.rs
cloudhead: Ah, so not the id, but the commit hash of the 'current' doc?

FintanH: Ya, exactly :) If you look at the tests, I'm resolving the `authorizing` to an `Oid` to look up that doc and filtering by the author name. The key use case would be similar but filtering on `is_delegate` of some sort.

FintanH: I was able to convince myself that, in fact, I'm wrong. `authorizing` being the project does need to be passed because the reference it's stored under needs its identifier. I still think it makes sense to set it in the `Entry` for inspecting the project at that `Oid` though.

However, perhaps what I should have explored is the `author` being optional, or at least the `PublicKey` and then the optional "identity" (DID or otherwise) -- which is what I think you suggested in discord :)

radicle-cob/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+
+[dependencies.git2]
+version = "0.15.0"
+default-features = false
+features = ["vendored-libgit2"]
+
+# TODO: point to crates.io
+[dependencies.git-commit]
+path = "../../radicle-git/git-commit"
+version = "0.1"
```
FintanH, line 33: Last thing to do :pray: 

https://github.com/radicle-dev/radicle-git/pull/43

radicle-cob/Cargo.toml
```diff
@@ -0,0 +1,56 @@
+[package]
+name = "radicle-cob"
+version = "0.1.0"
+authors = [
+  "Alex Good <alex@memoryandthought.me>",
+  "Fintan Halpenny <fintan.halpenny@gmail.com>",
+]
+edition = "2018"
```
cloudhead, line 8: Can we bump this to 2021?

radicle-cob/src/object/collaboration/create.rs
```diff
@@ -0,0 +1,99 @@
...
+/// The `resource` is the parent of this object, for example a
+/// software project.
+///
+/// The `args` are the metadata for this [`CollaborativeObject`]. See
+/// [`Create`] for further information.
+pub fn create<S, Signer, Person, Resource>(
+    storage: &S,
+    signer: Signer,
+    resource: &Resource,
+    args: Create<Person>,
```
cloudhead, line 51: I'm seeing this "author" type sometimes be called "Person" and sometimes "Author"; I'd suggest "Author" is more applicable since it's not really related to human-ness. Perhaps this also relates to or resolves the question in git-commit around the "Person" type.

radicle-cob/src/type_name.rs
```diff
@@ -0,0 +1,78 @@
...
+/// * `xyz.rad.patches.releases`
+#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+pub struct TypeName(String);
+
+impl TypeName {
+    /// A string representation of the typename which will match the typename in
+    /// regular expressions. This primarily escapes periods
+    pub fn regex_safe_string(&self) -> String {
+        self.0.replace('.', "\\.")
+    }
```
cloudhead, line 28: Are we still using this after removing regex?

radicle-cob/src/tests.rs
```diff
@@ -0,0 +1,253 @@
...
+fn copy_to(
+    repo: &git2::Repository,
+    from: &test::RemoteProject,
+    to: &test::RemoteProject,
+    typename: &TypeName,
+    object: ObjectId,
+) -> Result<(), git2::Error> {
+    let original = {
+        let name = format!(
+            "refs/rad/{}/cobs/{}/{}",
```
cloudhead, line 236: I'm probably missing something, but what are these refs used for? Looks unfamiliar to me.

radicle-cob/src/test/crypto.rs
```diff
@@ -0,0 +1,229 @@
+use std::{
```
cloudhead, line 1: We can depend on radicle-crypto now!

radicle-cob/src/sign/signatures.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+impl From<(PublicKey, crypto::Signature)> for Signature {
+    fn from((key, sig): (PublicKey, crypto::Signature)) -> Self {
+        Self { key, sig }
+    }
+}
+
+// FIXME(kim): This should really be a HashMap with a no-op Hasher -- PublicKey
+// collisions are catastrophic
```
cloudhead, line 47: :eyes:

radicle-cob/src/sign/message.rs
```diff
@@ -0,0 +1,41 @@
...
+        for (i, trailer) in self.trailers.iter().enumerate() {
+            write!(f, "{}", trailer.display(": "))?;
+            if i < self.trailers.len() {
+                f.write_str("\n")?;
+            }
+        }
+
+        Ok(())
+    }
+}
```
cloudhead, line 41: Can we use headers for this?

FintanH: Hmmm true, we probably could. I'd feel nervous about using headers since they're not really established in any documentation and seem to be part of internal `git` whereas this is exactly what trailers are for.

I'm happy to shift to headers, but just wanted to convey the above at least :)

radicle-cob/src/sign/signatures.rs
FintanH: :ghost:

radicle-cob/src/test/crypto.rs
FintanH: Oh ya, we do! I just forgot to delete this file :smile:

radicle-cob/src/tests.rs
FintanH: I had to make up a storage layout for the tests so I just have them under:
```
refs/rad/<project>/<person>/cobs/...
```

These can be whatever you want in `radicle` :kissing_closed_eyes:

radicle-cob/src/type_name.rs
FintanH: Probably not! Can try remove and see what happens

radicle-cob/src/object/collaboration/create.rs
FintanH: Ya, I can change it :+1:

radicle-cob/src/sign/message.rs
cloudhead: Yeah good point. No real issue with trailers, the only thing is that it means we have header parsing code as well as trailer parsing code, but it's not such a big deal. Let's just try to keep it in one place somewhere.

radicle-cob/src/tests.rs
cloudhead: Oh I see!

radicle-cob/src/sign/message.rs
FintanH: I was thinking I could add trailers to `git-commit` so they can be parsed there

cloudhead: Yes great idea. There's some code here that may be useful as well: https://github.com/radicle-dev/heartwood/blob/master/radicle/src/storage/git.rs#L656 -- I think I'll probably delete it once we move the sigs to the headers.

FintanH:
@cloudhead I think this just needs a final review and I can rebase the fixups if it all looks good :+1:

radicle-cob/src/backend/git/change.rs
```diff
@@ -0,0 +1,284 @@
...
+        );
+        let author = commit::Author::try_from(&author)?;
+
+        let commit = Commit::new(
+            tree.id(),
+            parents,
+            author.clone(),
+            author,
+            headers,
+            message.unwrap_or_else(|| "new change".to_string()),
```
cloudhead, line 254: Can this be made non-optional?

radicle-cob/src/test/identity.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct Name(String);
+
+impl Name {
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
```
cloudhead, line 29: Still using this?

radicle-cob/src/trailers.rs
```diff
@@ -0,0 +1,116 @@
...
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+mod author_commit {
+    super::oid_trailer! {AuthorCommitTrailer, "X-Rad-Author"}
+}
+mod resource_identity {
+    super::oid_trailer! {ResourceCommitTrailer, "X-Rad-Resource"}
+}
```
cloudhead, line 11: I think we can remove the `X-`, no one will murder us :D

cloudhead:
Ok looks like it's pretty much done, just found a few minor things that you may want to look into.

radicle-cob/src/test/identity.rs
FintanH: Again, it's mostly there to simulate a testing scenario so that I can have projects and remotes of that project :)

FintanH:
Made the message fields non-optional and remove the `X-` (hope no one kills me). Let's merge this thing :pray:

PR #61, opened 2022-10-19
radicle: fix error typo

slack-coder:
Correct radicle::node::Error::Connect's string.

### DIFF
--- a/radicle/src/node.rs
+++ b/radicle/src/node.rs
@@ -15,7 +15,7 @@ pub const DEFAULT_SOCKET_NAME: &str = "radicle.sock";
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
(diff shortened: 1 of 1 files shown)


radicle/src/node.rs
```diff
@@ -15,7 +15,7 @@ pub const DEFAULT_SOCKET_NAME: &str = "radicle.sock";
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-    #[error("failed to connecto to node: {0}")]
+    #[error("failed to connect to to node: {0}")]
```
cloudhead, line 18: Lol, you fixed one type and introduced another :sweat_smile:

FintanH: I kind of like `connecto` :P

slack-coder: hahahahahaahaha

PR #62, opened 2022-10-19
rad-auth: allow running multiple times

slack-coder:
Instead of failing over when a key has been created, run as normal using the previously created profile.

### DIFF
--- a/radicle/src/keystore.rs
+++ b/radicle/src/keystore.rs
@@ -58,15 +58,21 @@ impl UnsafeKeystore {
         Ok(())
     }
 
-    pub fn get(&self) -> Result<(PublicKey, SecretKey), Error> {
-        let public = fs::read(self.path.join("radicle.pub"))?;
+    pub fn get(&self) -> Result<Option<(PublicKey, SecretKey)>, Error> {
+        let public = self.path.join("radicle.pub");
+        let secret = self.path.join("radicle");
+        if !public.exists() && !secret.exists() {
+            return Ok(None);
+        }
+
+        let public = fs::read(public)?;
         let public = String::from_utf8(public)?;
         let public = PublicKey::from_pem(&public)?;
 
-        let secret = fs::read(self.path.join("radicle"))?;
+        let secret = fs::read(secret)?;
         let secret = String::from_utf8(secret)?;
         let secret = SecretKey::from_pem(&secret)?;
 
-        Ok((public, secret))
+        Ok(Some((public, secret)))
     }
 }
--- a/radicle/src/profile.rs
+++ b/radicle/src/profile.rs
@@ -13,11 +13,23 @@
 use std::path::PathBuf;
 use std::{env, io};
 
+use thiserror::Error;
+
 use crate::crypto::{KeyPair, PublicKey, SecretKey, Signature, Signer};
-use crate::keystore::{Error, UnsafeKeystore};
+use crate::keystore::UnsafeKeystore;
 use crate::node;
 use crate::storage::git::transport;
 use crate::storage::git::Storage;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] io::Error),
+    #[error(transparent)]
+    KeyStore(#[from] crate::keystore::Error),
+    #[error("no profile found at the filepath '{0}'")]
+    NotFound(PathBuf),
+}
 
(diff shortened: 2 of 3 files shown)


FintanH (commented):
Is it ok to just generate it if it doesn't exist? What if someone _expected_ the profile to exist but there was an issue loading it?

I'd suggest reporting the error and giving the user the choice to generate one, or suggest how to create one.

slack-coder:
I agree, or expose this via a sub-command or argument.  However the state of things is as an MVP with the objective to setup test nodes.  Adding new UI and arguments is outside the scope now.

cloudhead:
Yeah so the issue is that we don't know why it failed, and it may be for other reasons than it not existing, since we're not matching on the error.

slack-coder:
I took what feels like an odd approach to reporting the error.  I use a generic NotFound error, and elsewhere return an option.  Is it kosher rust?

radicle/src/keystore.rs
```diff
@@ -18,6 +18,8 @@ pub enum Error {
     FromUtf8Error(#[from] std::string::FromUtf8Error),
     #[error("invalid key format: {0}")]
     Crypto(#[from] crypto::Error),
+    #[error("{0} not found")]
```
FintanH, line 21: I think it would make more sense to report what was missing. So more like:
```rust
#[error("profile keys did not exist at 'radicle.pub` or 'radicle'")]
KeysNotFound,
```

slack-coder: Its the error being reported by the Profile::load() though.  Would we be good with wrapping the KeyNotFound error with NotFound(..).  Having a generic to signal when something doesn't exist is pretty useful, in place of returning an Option::None.

FintanH: Why put it in `keystore::Error` then? Maybe there should be a `profile::Error` that contains a variant for `keystore::Error` and `NotFound`. I also still think the error should be informative on how to fix the issue, `"profile not found"` is not very helpful :)

slack-coder: > Why put it in `keystore::Error` then? Maybe there should be a `profile::Error` that contains a variant for `keystore::Error` and `NotFound`.

Done :+1:

> I also still think the error should be informative on how to fix the issue, `"profile not found"` is not very helpful :)

From which perspective?  User facing or developer facing?  I've elaborated the error by including the RAD_HOME path the profile was attempted to load from.

FintanH (approved):
Way better :)

radicle/src/keystore.rs
FintanH: Both :smile: It's good to have information for solving an issue whether you're a dev or a user -- and in Radicle's case a lot of users are devs :P

radicle/src/profile.rs
```diff
@@ -13,11 +13,23 @@
...
 use crate::node;
 use crate::storage::git::Storage;
 
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] io::Error),
+    #[error(transparent)]
+    KeyStore(#[from] crate::keystore::Error),
+    #[error("no profile found at the filepath \"{0}\"")]
```
cloudhead, line 29:
```suggestion
    #[error("no profile found at the file path {0:?}")]
```

cloudhead: (The debug output using `:?` will add quotes to strings and paths)

slack-coder: I believe using the string version is still better.  Its a direct way of showing whats happening with the cost of a few extra characters.  Its also consistent with how other types are handled when reporting errors.  Adding inconsistencies makes the code harder to understand.

cloudhead: If you search the codebase for an escaped double quote, I don't think you'll find one :)
I generally avoid using escape sequences because they add noise and there's almost always a way to do without (single quotes, back-ticks or `:?`).

FintanH: Ya, I usually opt for `'{0}'`. They're also useful for needing to copy-paste values. TIL about `:?` working like that with paths :)

slack-coder: Opted for single ticks.

"TIL about :? working like that with paths"

Indeed it'd help newcomers to the code base if we opt for single ticks over `:?`.  Its a small extra detail they'd need to know, but death by a thousand paper cuts and all.

PR #64, opened 2022-10-25
Switch to custom git transports

cloudhead:
See individual commits for details.

This was a bit of a pain in the arse to get working, but I'm pretty happy with where I ended up.


radicle/src/storage/git/transport/remote.rs
```diff
@@ -0,0 +1,131 @@
+//! Git sub-transport used for fetching radicle data.
+//!
+//! To have control over the communication, and to allow git streams to be multiplexed over
+//! existing TCP connections, we implement the [`git2::transport::SmartSubtransport`] trait.
+//!
+//! We choose `rad` as the URL scheme for this custom transport, and include only the identity
```
FintanH: This should be `heartwood`, right?

radicle/src/storage/git/transport/remote.rs
```diff
@@ -0,0 +1,131 @@
...
+//! of the repository we're looking to fetch, eg. `rad://zP1GztjSdYNHK7jpdrXbaJ6Ki2Ke`, since
```
FintanH: Same here

radicle/src/storage/git/transport/remote.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub fn insert(&self, id: Id, stream: Stream) {
+        #[allow(clippy::unwrap_used)]
+        self.streams.lock().unwrap().insert(id, stream);
+    }
+}
+
+impl git2::transport::SmartSubtransport for Smart {
+    /// Run a git service on this transport.
+    ///
+    /// Based on the URL, which must be of the form `rad://zP1GztjSdYNHK7jpdrXbaJ6Ki2Ke`,
```
FintanH: And here

radicle/src/storage/git/transport/local/url.rs
```diff
@@ -47,6 +48,16 @@ impl Url {
     pub const SCHEME: &str = "rad";
```
FintanH, line 48: Is there a reason the scheme is `"rad"` and not `"rad://"`?

radicle/src/storage/git/transport/remote.rs
```diff
@@ -3,72 +3,46 @@
 //! To have control over the communication, and to allow git streams to be multiplexed over
 //! existing TCP connections, we implement the [`git2::transport::SmartSubtransport`] trait.
 //!
-//! We choose `rad` as the URL scheme for this custom transport, and include only the identity
-//! of the repository we're looking to fetch, eg. `rad://zP1GztjSdYNHK7jpdrXbaJ6Ki2Ke`, since
-//! we expect a connection to a host to already be established.
+//! We choose `heartwood` as the URL scheme for this custom transport, and include the node we'd
```
FintanH, line 6: Oh haha, nvm...

FintanH (approved):
IGNORE ME!

But looks good to me :)

radicle/src/storage/git/transport/local/url.rs
cloudhead: Mainly it's because we use this var to register the scheme with git2, and it expects the scheme without the `://`. I think technically speaking, `://` is also not part of the scheme, if I'm not mistaken.

radicle/src/storage/git/transport/local/url.rs
```diff
@@ -0,0 +1,106 @@
+//! Git local transport URLs.
+use std::str::FromStr;
```
slack-coder, line 3: Was it considered to call the modules after their schemes?  'rad'/'heartwood' versus 'local'/'remote'?  The scheme names are unlikely to change, and its more transparent on the filesystem level about which scheme is held where.  Adding new schemes, even if unlikely, would be straightforward too.

radicle/src/storage/git/transport/local/url.rs
```diff
@@ -0,0 +1,106 @@
...
+    #[error("invalid url format: expected `rad://<repo>[/<namespace>]`")]
+    InvalidFormat,
+    /// Unsupported URL scheme.
+    #[error("unsupported scheme: expected `rad://`")]
+    UnsupportedScheme,
+    /// Invalid repository identifier.
+    #[error("repo: {0}")]
+    InvalidRepository(#[source] IdError),
+    /// Invalid namespace.
+    #[error("namespace: {0}")]
```
slack-coder, line 27: Adding either ' or " around the values would make it really clear if the invalid namespace or repo is empty or has empty spaces.  What do you think?

radicle/src/storage/git/transport/local/url.rs
```diff
@@ -0,0 +1,106 @@
...
+    InvalidNamespace(#[source] crypto::PublicKeyError),
+}
```
slack-coder, line 29: Is it rust idiom to prepend the error with the type, even in cases where it has a dedicated module?

radicle/src/storage/git/transport/local/url.rs
FintanH: I like that idea :)

radicle/src/storage/git/transport/local/url.rs
FintanH: Tbh, I avoid it. I usually use `Error` if it's a single error in the module or I have an `error` module with different classes of errors:
```rust
// storage/error.rs

/// Things go wrong during `find`
pub enum Find {
  /* variants */
}

/// Things go wrong during `update`
pub enum Update {
  /* variants */
}
```

I think you'll find all sorts of ways in Rust, including the dreaded God `Error` -- it has all the error variants and _all_ your functions return it so you don't know how they particularly fail it could be any of the variants really.

radicle/src/storage/git/transport/local.rs
```diff
@@ -1,2 +1,107 @@
...
+            let result = child
+                .wait()
+                .map_err(|e| git2::Error::from_str(e.to_string().as_str()))?;
+
+            if !result.success() {
+                if let Some(code) = result.code() {
+                    log::error!("Child exited with error {}", code);
+                } else {
+                    log::error!("Child exited with unknown error code");
+                }
```
slack-coder, line 85: should a failure be returned here instead?  The caller may be the only one with context to understand how its handled.

radicle/src/storage/git/transport/local.rs
```diff
@@ -1,2 +1,107 @@
...
+use once_cell::sync::OnceCell;
+
+use crate::storage;
+use crate::storage::git::Storage;
+
+use super::ChildStream;
+
+thread_local! {
+    /// Stores a storage instance per thread.
+    static THREAD_STORAGE: OnceCell<Storage> = OnceCell::default();
```
slack-coder, line 19: The comment may be more helpful with a reason for it.   My understanding is its at most one per thread to avoid race conditions.

radicle/src/storage/git/transport/local/url.rs
cloudhead: yeah I considered that, but I chose a more descriptive approach, ie. naming them around their use, as especially for the remote transport, the scheme name `heartwood` is really just an internal detail (I was considering naming it `internal` at some point).

radicle/src/storage/git/transport/local/url.rs
cloudhead: Yeah they are both idiomatic. I tend to use prefixes when the modules are re-exported; eg. in this case I'm accessing the types via `transport::remote::{Url, UrlError}`, everywhere, not `transport::remote::url::Url`.

radicle/src/storage/git/transport/local.rs
cloudhead: Yeah good point, I only returned an error in the case the wait fails.

radicle/src/storage/git/transport/local/url.rs
cloudhead: Yeah, it's just easier to add `://` to `rad` than to remove it!

PR #67, opened 2022-10-29
Move `radicle/id` and `radicle/signature`

### LINKED ISSUE #22: Move the identity branch into the `radicle/` hierarchy
Currently, the identity branch lives under `refs/heads/radicle/id`. Though this makes things easier in some ways, it's probably better to move it to `refs/radicle/id`, and have an easy way to set it from a regular head.

This should also prevent the accidental fetching of the identity ref during a `git fetch rad`.

Once this is done, we can imagine some kind of tool that sets `refs/radicle/id` to the current `HEAD`, after checking that `HEAD` has the signature threshold required for the project. This will allow regular patch-based workflows to be used, and then the update can be committed to the hidden ref. Naturally, we'll also need a tool to checkout the identity branch into your working copy, to be able to edit it.

cloudhead:
We move the identity branch out of `heads/` and to the old location:

    refs/rad/id

We move the signed refs branch to:

    refs/rad/sigrefs

Keeping both "special" branches under the same hierarchy makes things a bit easier to work with, and tells the user that they should be updated with special tooling.

We also revert to `rad/` because we no longer have the issue of having `rad/rad/id` when doing a checkout of the remote branch in a working copy.

Closes #22

### DIFF
--- a/radicle/src/git.rs
+++ b/radicle/src/git.rs
@@ -19,7 +19,7 @@ pub use ext::Oid;
 pub use git2 as raw;
 pub use git_ref_format as fmt;
 pub use git_ref_format::{
-    lit, name, qualified, refname, Component, Namespaced, Qualified, RefStr, RefString,
+    component, lit, name, qualified, refname, Component, Namespaced, Qualified, RefStr, RefString,
 };
 pub use radicle_git_ext as ext;
 pub use storage::git::transport::local::Url;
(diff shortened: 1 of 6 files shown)


slack-coder (approved):

radicle/src/git.rs
```diff
@@ -79,17 +76,36 @@ pub mod refs {
     pub mod storage {
         use super::*;
 
+        /// Where the project's identity document is stored.
+        ///
+        /// `refs/rad/id`
+        ///
+        pub static IDENTITY_BRANCH: Lazy<Qualified> = Lazy::new(|| {
```
FintanH, line 83: nit: a branch is technically something that is under `refs/heads` or `refs/remotes`, I believe.

FintanH (approved):

radicle/src/git.rs
cloudhead: Hmmm, yeah I've seen it called branch in a couple places in link I think. I do think that it's more descriptive than `IDENTITY_REF`, and it is "branch-like", but perhaps let's see if it's confusing people and change it then.

FintanH: Ya, good points! :) Let's leave it as-is then

cloudhead: Eg. in identity-resolution.adoc:
image

PR #70, opened 2022-10-30
node: use method to access session's Node ID

slack-coder:
Make accessing an important session value, the Node ID, as a method to Session.

### DIFF
--- a/radicle-node/src/service/session.rs
+++ b/radicle-node/src/service/session.rs
@@ -122,4 +122,11 @@ impl Session {
         }
         Ok(())
     }
+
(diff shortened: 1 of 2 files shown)


radicle-node/src/service/session.rs
```diff
@@ -1,6 +1,7 @@
 use crate::service::message;
 use crate::service::message::Message;
 use crate::service::net;
+use crate::service::session;
```
cloudhead, line 4: Why?

cloudhead (commented):
Makes sense!

PR #71, opened 2022-10-30
node: implement node announcement sybil resistance

### LINKED ISSUE #28: Node announcement DoS protection
To prevent DoS by flooding the networking with bogus `NodeAnnouncement` messages, we include a small proof-of-work within the announcement message that is verified on receipt, before the message is relayed and processed.

We can use the same principle as bitcoin mining: include a nonce in the message and check that the message hash is below a certain number.

cloudhead:
We implement a small proof-of-work on the node announcement message to prevent the routing table from being filled by fake entries.

We choose `scrypt` for its memory-hard properties.

Closes #28 

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -687,23 +687,24 @@ where
                     }
                 }
             }
-            AnnouncementMessage::Node(NodeAnnouncement {
-                features,
-                alias,
-                addresses,
-                ..
-            }) => {
+            AnnouncementMessage::Node(
+                ann @ NodeAnnouncement {
+                    features,
+                    alias,
+                    addresses,
+                    ..
+                },
+            ) => {
                 // Discard node messages we've already seen, otherwise update
-                // out last seen time.
+                // our last seen time.
                 if !peer.node_announced(timestamp) {
                     debug!("Ignoring stale node announcement from {announcer}");
                     return Ok(false);
                 }
 
-                // If this node isn't a seed, we're not interested in adding it
-                // to our address book, but other nodes may be, so we relay the message anyway.
-                if !features.has(Features::SEED) {
-                    return Ok(relay);
+                if !ann.validate() {
+                    warn!("Dropping node announcement from {announcer}: invalid proof-of-work");
+                    return Ok(false);
                 }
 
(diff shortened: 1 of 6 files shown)


radicle-node/src/service.rs
```diff
@@ -710,6 +711,12 @@ where
                     }
                 };
 
+                // If this node isn't a seed, we're not interested in adding it
+                // to our address book, but other nodes may be, so we relay the message anyway.
+                if !features.has(Features::SEED) {
+                    return Ok(relay);
+                }
```
slack-coder, line 718: :+1:

radicle-node/src/service.rs
```diff
@@ -1217,7 +1224,9 @@ mod gossip {
             timestamp,
             alias,
             addresses,
+            nonce: 0,
         }
+        .solve()
```
slack-coder, line 1229: Would a more elaborate name be better?  Something related to proof of work like solve_proof_of_work(..) or gen_proof_of_work(..).  It better clarifies whats being solved.

radicle-node/src/service/message.rs
```diff
@@ -271,6 +320,14 @@ pub struct Announcement {
 }
 
 impl Announcement {
+    /// Proof-of-work iterations for announcements.
+    #[cfg(test)]
+    pub const POW_PARAMS: (u8, u32, u32) = (1, 1, 1);
+    #[cfg(not(test))]
+    pub const POW_PARAMS: (u8, u32, u32) = (15, 8, 1);
```
slack-coder, line 327: Is there a simple way to explain why the numbers are chosen?  Coming from the view of someone new to scrypt, and understanding when and how we should change them.

slack-coder:
Will scrypt with these parameters have a notable affect the memory requirements?  Or will our nodes still be able to run on rasp pis?

radicle-node/src/service/message.rs
cloudhead: Yeah I can add some docs :+1:

cloudhead:
> Will scrypt with these parameters have a notable affect the memory requirements? Or will our nodes still be able to run on rasp pis?

Good question. It should be fine with these settings, as these are the settings recommended for general use!

Issue #25, opened 2022-10-06
Connection and re-connection logic


cloudhead:
Currently, we only re-connect to "persistent" peers.

We should maintain a target number of outbound peers, eg. `8`, and attempt to connect to new peers using our address book (#24) when one of the connections drops.

The goal is to always be connected to the target number.
We should also accept incoming connections up to a certain limit, eg. `128`.

The target outbound and max inbound numbers are important to set deliberately, as they affect the network topology. Since connections are a "scarce" resource, we don't want the outbound number to be too high.

cloudhead:
Closed by #75

PR #75, opened 2022-11-01
node: keep healthy connection to network

slack-coder:
Check to keep a healthy connection to the Radicle network when a disconnect occurs (up to 30 seconds), or when a node disconnects.

Also:
 - Rename peer -> session
 - Update recorded sessions when disconnect is official.

References #25

### DIFF
--- a/radicle-node/src/main.rs
+++ b/radicle-node/src/main.rs
@@ -14,6 +14,7 @@ type Reactor = nakamoto_net_poll::Reactor<net::TcpStream>;
 struct Options {
     connect: Vec<Address>,
     listen: Vec<net::SocketAddr>,
+    external_addresses: Vec<Address>,
 }
 
 impl Options {
@@ -22,6 +23,7 @@ impl Options {
 
         let mut parser = lexopt::Parser::from_env();
         let mut connect = Vec::new();
+        let mut external_addresses = Vec::new();
         let mut listen = Vec::new();
 
         while let Some(arg) = parser.next()? {
@@ -34,6 +36,10 @@ impl Options {
                     let addr = parser.value()?.parse()?;
                     listen.push(addr);
                 }
+                Long("external-address") => {
+                    let addr = parser.value()?.parse()?;
+                    external_addresses.push(addr);
+                }
                 Long("help") => {
                     println!("usage: radicle-node [--connect <addr>]..");
                     process::exit(0);
@@ -41,7 +47,11 @@ impl Options {
                 _ => return Err(arg.unexpected()),
             }
         }
-        Ok(Self { connect, listen })
+        Ok(Self {
+            connect,
+            listen,
+            external_addresses,
+        })
     }
 }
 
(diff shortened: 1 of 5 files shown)


slack-coder:
Making some updates

slack-coder:
Cleaned up the commits and changes, now ready for review.

slack-coder:
Updated to completion for maintaining connections to the Radicle network.

radicle-node/src/main.rs
```diff
@@ -34,14 +36,22 @@ impl Options {
                     let addr = parser.value()?.parse()?;
                     listen.push(addr);
                 }
+                Long("external-address") => {
```
slack-coder, line 39: bitcoind uses '-externalip'

radicle-node/src/service.rs
```diff
@@ -943,14 +947,49 @@ where
...
-        if self.sessions.len() < TARGET_OUTBOUND_PEERS {
-            let delta = TARGET_OUTBOUND_PEERS - self.sessions.len();
-
-            for _ in 0..delta {
-                // TODO: Connect to random peer.
+    fn choose_addresses(&mut self) -> Vec<Address> {
+        let mut initializing: Vec<Address> = Vec::new();
+        let mut negotiated: HashMap<NodeId, &Session> = HashMap::new();
+        for s in self.sessions.values() {
+            if !s.link.is_outbound() {
```
slack-coder, line 954: why is 's.link.is_outbound()' preferred over 's.link == outbound'?

cloudhead: because it doesn't require the `Link` type to be in scope :shrug:

radicle-node/src/main.rs
cloudhead: ok yes this makes sense, how do they specify the port?

slack-coder: I recommend to keep ports.  There is an important distinction where radicle serves user information in a way bitcoin does not.  Bitcoin's information is held on a unified ledger vs. projects sparsely hosted across the network.  Prioritizing connectivity is more important for us (multiple users on a local area network etc.).

radicle-node/src/test/tests.rs
```diff
@@ -648,6 +661,80 @@ fn test_persistent_peer_reconnect() {
...
+
+    let mut address_book = address::Book::memory().unwrap();
+    for peer in &unconnected {
+        add_peer(&mut address_book, peer);
+    }
+
+    let mut alice = Peer::config(
+        "alice",
+        Config {
+            project_tracking: ProjectTracking::Allowed(HashSet::default()),
```
cloudhead, line 698: Is this necessary?

radicle-node/src/test/tests.rs
```diff
@@ -648,6 +661,80 @@ fn test_persistent_peer_reconnect() {
...
+    assert_eq!(
+        connected.len(),
+        alice.sessions().len(),
+        "alice should be connected to all peers"
+    );
+
+    for peer in connected.iter() {
+        alice.disconnected(
+            &peer.addr(),
+            &nakamoto::DisconnectReason::Protocol(DisconnectReason::Error(session::Error::Timeout)),
```
cloudhead, line 719: We want to use an exogenous reason for disconnection here, so not one provoked by the protocol; as the protocol might behave differently.

PR #80, opened 2022-11-03
Zeroize all the things

### LINKED ISSUE #79: Properly zeroize secret keys and passphrases

cloudhead:
Closes #79 

I opted for Zeroize for the passphrase because (1) it's already a dependency, and (2) it seems to do the same thing

### DIFF
--- a/radicle-crypto/src/ssh/keystore.rs
+++ b/radicle-crypto/src/ssh/keystore.rs
@@ -3,6 +3,7 @@ use std::path::{Path, PathBuf};
 use std::{fs, io};
 
 use thiserror::Error;
+use zeroize::Zeroizing;
 
 use crate::{KeyPair, PublicKey, SecretKey, Signature, Signer, SignerError};
 
@@ -82,7 +83,7 @@ impl Keystore {
 
     /// Load the secret key from the store, decrypting it with the given passphrase.
     /// Returns `None` if it wasn't found.
-    pub fn secret_key(&self, passphrase: &str) -> Result<Option<SecretKey>, Error> {
+    pub fn secret_key(&self, passphrase: &str) -> Result<Option<Zeroizing<SecretKey>>, Error> {
         let path = self.path.join("radicle");
         if !path.exists() {
             return Ok(None);
@@ -93,7 +94,7 @@ impl Keystore {
 
         match secret.key_data() {
             ssh_key::private::KeypairData::Ed25519(pair) => {
-                Ok(Some(SecretKey::new(pair.to_bytes())))
+                Ok(Some(SecretKey::from(pair.to_bytes()).into()))
             }
             _ => Err(Error::InvalidKeyType),
         }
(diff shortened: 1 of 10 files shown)


dr-orlovsky:
Fyi we had a long discussion in rust-secp256k1 back in the days on Zeroize PR introduced to secret key. TL;DR: it doesn't work and gives false sense of security. Rust compiler does a lot of memory copying and to get Zeroize working you need to have `Pin` types first and do quite a complex setup

cloudhead:
> Fyi we had a long discussion in rust-secp256k1 back in the days on Zeroize PR introduced to secret key. TL;DR: it doesn't work and gives false sense of security. Rust compiler does a lot of memory copying and to get Zeroize working you need to have `Pin` types first and do quite a complex setup

Hmm interesting. Yeah it doesn't seem like a fool-proof solution, but I guess it's still worth the tiny bit of extra code :man_shrugging:

radicle-crypto/src/ssh/keystore.rs
```diff
@@ -82,7 +86,10 @@ impl Keystore {
 
     /// Load the secret key from the store, decrypting it with the given passphrase.
     /// Returns `None` if it wasn't found.
-    pub fn secret_key(&self, passphrase: &str) -> Result<Option<SecretKey>, Error> {
+    pub fn secret_key(
+        &self,
+        passphrase: Passphrase,
+    ) -> Result<Option<Zeroizing<SecretKey>>, Error> {
```
FintanH, line 92: Any reason why you didn't opt to implement/derive Zeroizing for `SecretKey`?

radicle-crypto/src/ssh/keystore.rs
```diff
@@ -3,9 +3,13 @@ use std::path::{Path, PathBuf};
 use std::{fs, io};
 
 use thiserror::Error;
+use zeroize::Zeroizing;
 
 use crate::{KeyPair, PublicKey, SecretKey, Signature, Signer, SignerError};
 
+/// A secret key passphrase.
+pub type Passphrase = Zeroizing<String>;
```
FintanH, line 11: Is it possible to `Zeroizing<&str>`? :thinking:

FintanH (approved):
:+1:

radicle-crypto/src/ssh/keystore.rs
cloudhead: Yeah, I tried that first actually but couldn't get it working; you can give it a spin if you want.

radicle-crypto/src/ssh/keystore.rs
cloudhead: Hehe nope, I tried that. I wonder if it's possible in theory though, since it has to have control over the drop, so it might not be possible with borrows.

FintanH: Ah ya, good point :) Slight optimization might be to do:
```
pub struct Passphrase(Zeroizing<String>);

impl From<&str> for Zeroizing {
  /* yada yada */
}

impl From<String> for Zeroizing {
  /* yada yada */
}
```

Would save some `to_owned().into()`s.

cloudhead: Yeah good idea :thinking:

PR #81, opened 2022-11-04
Update to Rust 1.65

cloudhead:


radicle/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(clippy::match_like_matches_macro)]
+#![allow(clippy::explicit_auto_deref)] // TODO: This can be removed when the clippy bugs are fixed
```
FintanH, line 2: Oh, what's this? :eyes:

cloudhead: There's a clippy regression -- some of the warnings from this lint are false-positives, so I had to disable it.

FintanH: Ohhh as in it wouldn't compile when you followed the clippy warning? Because I think I applied some changes for that warning but everything seemed to compile fine, so just wondering if the bug is just clippy warning you or if there's something wrong with applying the change.

radicle-node/src/wire.rs
```diff
@@ -129,6 +129,8 @@ impl Encode for PublicKey {
 
 impl<const T: usize> Encode for &[u8; T] {
     fn encode<W: io::Write + ?Sized>(&self, writer: &mut W) -> Result<usize, io::Error> {
+        // TODO: This can be removed when the clippy bugs are fixed
+        #[allow(clippy::explicit_auto_deref)]
```
slack-coder, line 133: Any idea which one? Linking would make it easy for others to know when to remove the fix

slack-coder:
Otherwise :+1:

radicle/src/lib.rs
cloudhead: Oh, no like if I apply the change, I get a compile error.

radicle-crypto/src/lib.rs
```diff
@@ -270,7 +270,7 @@ impl PublicKey {
         buf[..2].copy_from_slice(&Self::MULTICODEC_TYPE);
         buf[2..].copy_from_slice(self.0.deref());
 
-        multibase::encode(multibase::Base::Base58Btc, &buf)
+        multibase::encode(multibase::Base::Base58Btc, buf)
```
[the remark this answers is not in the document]
cloudhead: Added comment to commit.

radicle-node/src/wire.rs
cloudhead: I think it's this: https://github.com/rust-lang/rust-clippy/issues/9710

already merged so should find its way soon.

PR #91, opened 2022-11-08
Add `rad-track`, `rad-untrack` commands

xphoniex:


radicle-cli/src/commands/track.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::terminal::args::{Args, Error, Help};
+
+pub const HELP: Help = Help {
+    name: "track",
+    description: "Manage radicle project tracking relationships",
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad track <id> [--fetch]
```
xphoniex, line 18: Current implementation needs `<id>` passed to it, previously it'd have worked in working dir, and it'd have found out the `(urn, repo)` tuple from `project::cwd()`.

I can't find anything like that in `radicle/src/identity/project/id.rs` or `radicle/src/identity/project.rs`.

What should I do?

radicle-cli/src/commands/track.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
+    let profile = ctx.profile()?;
+    let node = radicle::node::connect(&profile.node())?;
+
+    term::info!(
+        "Establishing 🌱 tracking relationship for {}",
+        term::format::highlight(&options.id.to_human())
```
xphoniex, line 78: is `id.to_human()` good enough? I'd prefer if we display something about repo as well, e.g. name like before.

radicle-cli/src/commands/track.rs
```diff
@@ -0,0 +1,94 @@
...
+    );
+    term::blank();
+
+    if node.track(&options.id)? {
```
xphoniex, line 82: previous implementation can [distinguish](https://github.com/radicle-dev/radicle-cli/blob/master/track/src/lib.rs#L118-L124) between when relationship `exists` and when it's just `established`. I looked at `node` implementation and looks like only when it's `established` we get a `true`?

radicle-cli/src/commands/track.rs
```diff
@@ -0,0 +1,94 @@
...
+        term::success!(
+            "Tracking relationship for {} established",
+            term::format::tertiary(&options.id.to_human()),
+        );
+    }
+
+    if options.fetch {
```
xphoniex, line 89: does this need any term message?

radicle-cli/src/commands/track.rs
xphoniex: I switched to `payload.name` instead https://github.com/radicle-dev/heartwood/compare/5d50941d85946a65945fe7cb4a83cd69856870fd..460dee1f002c8887f616a7aa74d2431c280e1b42

radicle-cli/src/commands/track.rs
cloudhead: Check `radicle::rad::remote` function

radicle-cli/src/commands/track.rs
cloudhead: Yes, `true` => established, `false` => already established (noop)

The implementation doesn't actually work yet (always returns true), but I'm fixing this now, so you can already depend on it.

xphoniex: does `false` always mean established?

cloudhead:
This implementation is actually different from the original. `track` was previously tracking a PeerId in the context of a project. In the above code, it tracks a project.

Since we don't actually have an API right now to track a specific peer, I see how you got confused. I'd leave it as-is for now (tracking the project), and once we have an API to track specific peers, we'll fix it. However this means the `id` should be optional and use the current project inside a project dir, ie. `rad track [<id>]`. Once we add peers, it should be changed to take two arguments I guess.

radicle-cli/src/commands/untrack.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        Ok((Options { id }, vec![]))
+    }
+}
+
+pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
+    let profile = ctx.profile()?;
+    let storage = &profile.storage;
+    let Doc { payload, .. } = storage.repository(options.id)?.project_of(profile.id())?;
+
```
cloudhead, line 70:
```suggestion
```

radicle-cli/src/commands/track.rs
cloudhead: Yes, same API as HashSet::insert/remove

PR #93, opened 2022-11-08
radicle: Collaborative Objects

FintanH:
While wiring up `radicle-cob` I came across simplifications and improvements that could be made to the library. See the commits for more details on each improvement.

The final commit represents all the wiring up. The commit message also details the process that went into it.


radicle/src/cob.rs
```diff
@@ -0,0 +1,121 @@
...
+    storage::git::Repository,
+};
+
+/// The `Author` of a [`create`] or [`update`].
+///
+/// **Note**: `Author` implements [`identity::Identity`], but since it
+/// is not content-addressed, the [`identity::Identity::content_id`]
+/// returns [`git2::Oid::zero`]. This means that if the `author` is
+/// set in the updates the history entries for those changes will
+/// contain the zero `Oid` for the `author` field.
```
cloudhead, line 19: I'm definitely missing some context here, let's chat about this at some point. I'd like to understand this `Identity` trait better and if we need to implement it given the caveat mentioned, or if there's a way to split it out or change the trait, eg. have content_id return an `Option<Oid>`.

radicle/src/cob.rs
```diff
@@ -0,0 +1,121 @@
...
+///
+/// The `repository` is the project this collaborative object is being
+/// stored under.
+///
+/// The `typename` is the type of objects to listed.
+pub fn list(
+    repository: &Repository,
+    typename: &TypeName,
+) -> Result<Vec<CollaborativeObject>, error::Retrieve> {
+    cob::list(repository, typename)
```
cloudhead, line 93: Eventually it might make sense to have this return an iterator somehow, as there could be thousands of items here.

radicle/src/cob.rs
```diff
@@ -0,0 +1,121 @@
...
+    repository: &Repository,
+    signer: S,
+    project: &Identity<Oid>,
+    args: Create<Author>,
+) -> Result<CollaborativeObject, error::Create>
+where
+    S: crypto::Signer,
+{
+    let namespace = *signer.public_key();
+    cob::create(repository, signer, project, &namespace, args)
```
cloudhead, line 64: Would it make sense for the `cob::update/create` functions to infer the namespace from the signer? It would at least prevent eg. `cob::create(repo, bob_signer, proj, alice_namespace, args)`

radicle/src/git.rs
```diff
@@ -106,6 +113,48 @@ pub mod refs {
...
+        ///
+        /// `refs/namespaces/<remote>/refs/cobs/<typename>/<object_id>`
+        ///
+        pub fn cob<'a>(
+            remote: &RemoteId,
+            typename: &cob::TypeName,
+            object_id: &cob::ObjectId,
+        ) -> Namespaced<'a> {
+            Qualified::from_components(
+                component!("cobs"),
```
cloudhead, line 127: I sort of recall seeing a ref somewhere that had a singular `cob` in it. Just want to flag that in case there's a mismatch somewhere.

radicle/src/git.rs
```diff
@@ -106,6 +113,48 @@ pub mod refs {
...
+        /// `refs/namespaces/*/refs/cobs/<typename>/<object_id>`
+        ///
+        pub fn cobs(typename: &cob::TypeName, object_id: &cob::ObjectId) -> PatternString {
+            refspec::pattern!("refs/namespaces/*")
+                .join(refname!("refs/cobs"))
+                .join(Component::from(typename))
+                .join(Component::from(object_id))
+        }
+
+        pub fn cob_suffix<R>(cob: &R) -> Option<(cob::TypeName, cob::ObjectId)>
```
cloudhead, line 145: There probably isn't a better place for this right now, but I think we should eventually move it to a better home. Perhaps the type it returns, if common enough could have a `from_refstr` function that would be this function. Eg. `TypedObjectId::from_refstr(cob)`

radicle/src/identity/project.rs
```diff
@@ -402,6 +402,22 @@ pub struct Identity<I> {
...
+    }
+
+    fn content_id(&self) -> Oid {
+        self.current
+    }
+
+    fn identifier(&self) -> Self::Identifier {
+        self.root
+    }
+}
```
cloudhead, line 419: Interesting :O

cloudhead: ... that both this type and `Author` are impls of `Identity`, given that this is a full doc while `Author` is just a pubkey. Something seems slightly off with regards to the intent of this trait, but I would have to understand the trait better to be sure. Just voicing an intuition..

radicle/src/storage/git.rs
```diff
@@ -651,6 +674,116 @@ impl WriteRepository for Repository {
...
+        &self,
+        identifier: &Self::Identifier,
+        typename: &cob::TypeName,
+        object_id: &cob::ObjectId,
+        change: &cob::Change,
+    ) -> Result<(), Self::UpdateError> {
+        self.backend.reference(
+            git::refs::storage::cob(identifier, typename, object_id).as_str(),
+            (*change.id()).into(),
+            false,
```
cloudhead, line 774: Are cobs essentially un-force-pushable?

cloudhead (approved):
Looks good to me! Left some questions/comments that are ok to address in future PRs if you think it's a good idea. I'm a little queasy about that Identity type impl on Author, but let's discuss it in person.

FintanH:
> I'm a little queasy about that Identity type impl on Author, but let's discuss it in person.

Same tbh, I'd be open to also just set that to `None` for the time being.

radicle/src/cob.rs
FintanH: I didn't want to do that because it being the namespace is specific to this particular storage layout. `radicle-cob` knowing about that doesn't seem clean to me. Otherwise, we could just write `radicle-cob` knowing the exact layout :shrug:

radicle/src/git.rs
FintanH: Mmmm ya, this could go in `radicle-cob` now that you mention it. I think generally a ref schema would always end in `<typename>/<object>`.

radicle/src/identity/project.rs
FintanH: Well, `Author` is the hack.

But perhaps I've messed this up. Should the `identifier` be `did:rad:<root>` or something like that? Actually... I don't think `identifier` is even used anymore :thinking: I think I'll need to investigate that. I'm pretty sure `content_id` is the only thing needed.

FintanH: Ya, I was able to remove that :)

radicle/src/cob.rs
cloudhead: I see I see hmmmm. let's leave it as is :+1:

radicle/src/git.rs
cloudhead: cool, yes agreed :+1:

radicle/src/identity/project.rs
cloudhead: Ah, great :+1:

radicle/src/git.rs
cloudhead: Yeah, seems like a useful type to have. Going to merge this and let you implement that in a separate PR

PR #94, opened 2022-11-09
radicle-cob: ObjectIdentifier

FintanH:


radicle-cob/src/object/collaboration.rs
```diff
@@ -25,28 +25,80 @@ pub use list::list;
...
 pub use update::{update, Update};
 
+/// The full object identifier for a [`CollaborativeObject`].
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ObjectIdentifier {
+    /// The [`TypeName`] for the given [`CollaborativeObject`].
+    pub name: TypeName,
+    /// The [`ObjectId`] for the given [`CollaborativeObject`].
+    pub object: ObjectId,
+}
```
cloudhead, line 35: Hmm, I feel like ObjectIdentifier stands for ObjectId; ie. they are too close and confusable. Some potential suggestions:

* `TypedObjectId`
* `TypedId`
* `TObjectId`

radicle-cob/src/change_graph.rs
```diff
@@ -15,7 +15,10 @@ use petgraph::{
 };
 
 use crate::{
-    change, object, signatures::Signatures, Change, CollaborativeObject, ObjectId, TypeName,
+    change,
+    object::{self, ObjectIdentifier},
+    signatures::Signatures,
+    Change, CollaborativeObject, ObjectId, TypeName,
```
slack-coder, line 21: Not particular to this PR, we should come to consensus on how imports are defined

radicle-cob/src/object/collaboration.rs
slack-coder: Or ObjectMetadata?  The TypeName is descriptive right?  Or wow does the type name affect the identity of the object?  This domain is new to me.

cloudhead: It basically describes the type of object refered to by the object-id, so like `(Issue, 123)`

FintanH: What about `CobId`, since it's the combination of the two that identifies a `CollaborativeObject`?

cloudhead: That could work, though there isn't as clear a relation betwen ObjectId and CobId :thinking:

cloudhead: Interestingly though, I haven't found that I "missed" that type while porting some of the code so far, so I wonder if we should scrap the type idea, and just put the function in some module hierarchy that makes sense?

slack-coder: @cloudhead I'm not quite understanding, could you make the idea more concrete?

PR #102, opened 2022-11-14
Adding transcoders to the wire protocol

dr-orlovsky:
This will enable Noise_XK (or any other form) of encoding for the peer connectivity.

The PR has an aim of getting approach and concept ACK, after which a real Noise encoder from the https://github.com/Internet2-WG/rust-cyphernet library will be integrated.


radicle-node/src/wire/transcoder.rs
```diff
@@ -0,0 +1,43 @@
...
+    type Error: std::error::Error;
+
+    /// Decodes data received from the remote peer and update the internal state
+    /// of the transcoder, either returning response which must be sent to the
+    /// remote (see [`DecodedData::Remote`]) or data which should be processed
+    /// by the local peer.
+    fn input(&mut self, data: &[u8]) -> Result<DecodedData, Self::Error>;
+
+    /// Encodes data before sending them to the remote peer.
+    fn output(&mut self, data: Vec<u8>) -> Result<Vec<u8>, Self::Error>;
```
cloudhead, line 26: Can we call these functions encode/decode? They both have input & output afaict.

radicle-node/src/wire.rs
```diff
@@ -423,51 +425,68 @@ impl Decode for node::Features {
...
+                log::trace!("performing handshake, sending to the remote peer more data");
+                self.handshake_queue.push_back((*addr, output));
+                return;
+            }
+            Err(err) => {
+                // TODO: Disconnect peer.
+                log::error!("invalid transcoder input. Details: {}", err);
+                return;
+            }
+        };
```
cloudhead, line 486: Wouldn't it be simpler if we got rid of `DecodedData`, and in the case of `DecodedData::Remote`, we just handle it internally in the transcoder? I'm a little confused on when/how the transcoder switches between local and remote, as for example in the case of a handshake, it would seem that a transition between `Remote -> Local` would happen once and then all further outputs would be `Local`?

radicle-node/src/parser.rs
```diff
@@ -4,16 +4,16 @@ use std::marker::PhantomData;
 use crate::service::message::Message;
 use crate::wire;
 
-/// Message stream decoder.
+/// Parser of Radicle messages from the network stream.
```
cloudhead, line 7: A "parser" is typically something that works with a grammar or language. Here, it's a binary stream, so I don't think parser makes sense as a term.

radicle-node/src/wire.rs
dr-orlovsky: We can't handle it in the transcoder since transcoder can't do a write - or doesn't maintains the queue of the events generated for the reactor. Here we are restricted by the specific reactor design from the nakamoto crate

cloudhead:
Concept ACK. This generally makes sense yes :+1:

I think some of the internal types/traits could be simplified or made stricter, but the general approach is the right one. From my understanding though, you plan on implementing both the wire decoder and the handshake as one module?

radicle-node/src/parser.rs
dr-orlovsky: My assumption was that the type system of P2P messages is a grammar. But I have no preference and happy to use any name & revert this commit.

dr-orlovsky:
> From my understanding though, you plan on implementing both the wire decoder and the handshake as one module?

I just do not see how they can be separated...

radicle-node/src/wire.rs
cloudhead: Sure, but we're keeping this Remote/Local distinction even after the handshake is over, instead of moving to a different transcoder. My suggestion would be to write it such that we start with eg. a `HandshakeTranscoder` for each peer, and then move on to a `WireTranscoder`, instead of having a single type that handles both. So it would be structured like:

```rust
struct State {
   negotiating: HashMap<Address, HandshakeTranscoder>, // Peers going through handshake
   negotiated: HashMap<Address, WireTranscoder>, // Peers after handshake
}
```

PR #103, opened 2022-11-14
Implement `rad-push`

cloudhead:

### DIFF
--- a/radicle-cli/src/commands/push.rs
+++ b/radicle-cli/src/commands/push.rs
@@ -1,0 +1,134 @@
+use std::ffi::OsString;
+use std::path::Path;
+
+use radicle::git;
+
+use crate::terminal as term;
+use crate::terminal::args::{Args, Error, Help};
+
+use anyhow::anyhow;
+
+pub const HELP: Help = Help {
+    name: "push",
+    description: env!("CARGO_PKG_DESCRIPTION"),
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad push [--all] [--[no-]sync] [<option>...]
+
+    By default, only the current branch is synced.
+
+Options
+
+    --all               Push all branches (default: false)
+    --sync              Sync after pushing to the "rad" remote (default: false)
+    --no-sync           Do not sync after pushing to the "rad" remote
+    --help              Print help
+
+Git options
+
+    -f, --force           Force push
+    -u, --set-upstream    Set upstream tracking branch
+
+"#,
+};
+
+#[derive(Default, Debug)]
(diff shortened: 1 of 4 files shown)


radicle-cli/src/commands/push.rs
```diff
@@ -0,0 +1,134 @@
...
+impl Args for Options {
+    fn from_args(args: Vec<OsString>) -> anyhow::Result<(Self, Vec<OsString>)> {
+        use lexopt::prelude::*;
+
+        let mut parser = lexopt::Parser::from_args(args);
+        let mut verbose = false;
+        let mut force = false;
+        let mut all = false;
+        let mut sync = false;
+        let mut set_upstream = false;
```
slack-coder, line 55: small suggestion, alphabetical order?

radicle-cli/src/commands/push.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    rad push [--all] [--[no-]sync] [<option>...]
+
+    By default, only the current branch is synced.
+
+Options
+
+    --all               Push all branches (default: false)
+    --sync              Sync after pushing to the "rad" remote (default: false)
+    --no-sync           Do not sync after pushing to the "rad" remote
```
FintanH, line 26: Why have both `sync` and `no-sync` flags? :thinking: Seems weird to have and feels like it should have precedence rules if it they both stay, i.e. which one trumps the other?

cloudhead: Good question - I usually have a `--no-*` version for many boolean flags as I think it's useful as self-documentation when you see the command written and to not force users to remember what the default is.

As far as I can tell this is pretty standard, eg. see this screenshot from the ripgrep manual:
image

radicle-cli/src/commands/push.rs
cloudhead: I don't think it's worth it for a list of 6 items?

radicle-cli/src/commands/push.rs
FintanH: Ya, so maybe we should follow suit and make `--no-sync` override any use of `--sync`.

cloudhead: https://github.com/radicle-dev/heartwood/issues/106

radicle-cli/src/commands/push.rs
slack-coder: Having clear ordering makes it easier to add items.  Newcomers have to figure out if there's a hidden order or not when adding fields.  Its stylistic and not a big deal, but I find it reduces a small amount of friction.

PR #105, opened 2022-11-15
Implement `rad-patch` command

cloudhead:
See commit. This is a port of the `radicle-cli` patch command.

The `sync` functionality is marked as `TODO` until we figure out what we want to do there. I didn't want to block this work from progressing because of that.

The other `TODO`s were there from the radicle-cli, so I kept them.

### DIFF
--- a/radicle-cli/src/commands/patch.rs
+++ b/radicle-cli/src/commands/patch.rs
@@ -1,0 +1,158 @@
+#[path = "patch/common.rs"]
+mod common;
+#[path = "patch/create.rs"]
+mod create;
+#[path = "patch/list.rs"]
+mod list;
+
+use std::ffi::OsString;
+use std::str::FromStr;
+
+use anyhow::anyhow;
+
+use radicle::cob::patch::PatchId;
+use radicle::prelude::*;
+
+use crate::terminal as term;
+use crate::terminal::args::{Args, Error, Help};
+use crate::terminal::patch::Comment;
+
+pub const HELP: Help = Help {
+    name: "patch",
+    description: "Work with radicle patches",
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad patch [<option>...]
+
+Create options
+
+    -u, --update [<id>]        Update an existing patch (default: no)
+        --[no-]sync            Sync patch to seed (default: sync)
+        --[no-]push            Push patch head to storage (default: true)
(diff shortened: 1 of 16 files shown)


radicle-cli/src/commands/patch.rs
```diff
@@ -0,0 +1,153 @@
...
+                }
+                Long("push") => {
+                    push = true;
+                }
+                Long("no-push") => {
+                    push = false;
+                }
+                Long("help") => {
+                    return Err(Error::Help.into());
+                }
```
slack-coder, line 121: What about ordering the arguments?  it looks pretty ad-hoc

radicle-cli/src/terminal/format.rs
```diff
@@ -17,15 +18,24 @@ pub fn node(node: &NodeId) -> String {
...
 }
 
+/// Format a timestamp.
+pub fn timestamp(time: &Timestamp) -> String {
+    let fmt = timeago::Formatter::new();
+    let now = Timestamp::now();
+    let duration = time::Duration::from_secs(now.as_secs() - time.as_secs());
+
+    fmt.convert(duration)
+}
```
slack-coder, line 37: What about a different method name like time_since(..) or similar?  The name and description appear a bit off from whats happening here.

slack-coder: Nevermind, I see the greater context.

radicle-cli/src/commands/patch.rs
slack-coder: Cool so always ordered by the struct's fields

radicle/src/storage/refs.rs
```diff
@@ -101,6 +101,17 @@ impl Refs {
         })
     }
 
+    /// Get a particular ref.
+    pub fn get(&self, name: &git::Qualified) -> Option<Oid> {
+        self.0.get(name.to_ref_string().as_refstr()).copied()
+    }
+
+    /// Get a particular head ref.
+    pub fn head(&self, name: &git::Component) -> Option<Oid> {
```
FintanH, line 110: I think you'd want `RefString` or `AsRef<RefStr>`, otherwise you can't do something like `finto/feature/patches`

cloudhead: Oh shit you're right!

cloudhead: Fixed.

FintanH (approved):

PR #107, opened 2022-11-16
Add licenses and contributor information

cloudhead:
Taken from other repos with some small edits.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# CONTRIBUTING
+
+Contributions are very welcome. When contributing code, please follow these
+simple guidelines.
+
+* Make sure you run `rustfmt` on your code. Also ensure all trailing whitespace is trimmed.
+* Run the tests with `cargo test --all`.
+* Before adding any code dependencies, check with the maintainers if this is okay.
+* Write properly formatted comments: they should be English sentences, eg:
+
+        // Return the current UNIX time.
+
+* Read the DCO and make sure all commits are signed off, using `git commit -s`.
+* Follow the guidelines when proposing code changes (see below).
+* Write properly formatted git commits (see below).
+
+Proposing changes
+-----------------
+When proposing changes via a pull-request or patch:
+
+* Isolate changes in separate commits to make the review process easier.
+* Don't make unrelated changes, unless it happens to be an obvious improvement to
+  code you are touching anyway ("boyscout rule").
+* Rebase on `master` when needed.
+* Keep your changesets small, specific and uncontroversial, so that they can be
+  merged more quickly.
+* If the change is substantial or requires re-architecting certain parts of the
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+* Follow the guidelines when proposing code changes (see below).
+* Write properly formatted git commits (see below).
+
+Proposing changes
+-----------------
+When proposing changes via a pull-request or patch:
+
+* Isolate changes in separate commits to make the review process easier.
+* Don't make unrelated changes, unless it happens to be an obvious improvement to
+  code you are touching anyway ("boyscout rule").
```
FintanH, line 23: What's "boyscout rule"? :smile:

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+A properly formed git commit subject line should always be able to complete the
+following sentence:
+
+     If applied, this commit will _____
+
+In addition, it should be capitalized and *must not* include a period.
+
+For example, the following message is well formed:
+
+     Add support for .gif files
```
FintanH, line 42: Is this for the message or the title? :thinking:

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+simple guidelines.
+
+* Make sure you run `rustfmt` on your code. Also ensure all trailing whitespace is trimmed.
+* Run the tests with `cargo test --all`.
+* Before adding any code dependencies, check with the maintainers if this is okay.
+* Write properly formatted comments: they should be English sentences, eg:
+
+        // Return the current UNIX time.
+
+* Read the DCO and make sure all commits are signed off, using `git commit -s`.
```
slack-coder, line 13: Does the DCO require outside contributors use their full legal (government) name for sign offs?

slack-coder:
LGTM

CONTRIBUTING.md
cloudhead: The message title :joy: 

I mean the "message" is the whole thing, ie. the short line and the description. But since the description is optional, that's a valid message as well as a valid title.

CONTRIBUTING.md
cloudhead: > Leave your code better than you found it.

FintanH: Ah! Makes sense :smile:

CONTRIBUTING.md
cloudhead: Hmm, we have a full example below.. But I'm not sure what part is confusing given that you do eg. `git commit --message "Add support..."`, I've never heard of the term "message" used for eg. only the description or only the title?

FintanH: Ah ya, fair enough the example is below

PR #109, opened 2022-11-16
Add TypeName & ObjectId parser

FintanH:
Added the `parse_refname` function for parsing a reference into its `TypeName` and `ObjectId`, if possible.

This replaces the `cob_suffix` function in `radicle`.

This change also fixes the `FromStr` instance of `ObjectId` to use `from_str` instead of `TryFrom` for bytes.


radicle-cob/src/object/collaboration.rs
```diff
@@ -53,3 +53,43 @@ impl CollaborativeObject {
         self.history.tips().into_iter().map(Oid::from).collect()
     }
 }
+
+/// Takes a `refname` and performs a best attempt to extract out the
+/// [`TypeName`] and [`ObjectId`] from it.
+///
```
slack-coder, line 59: What do you think about mentioning 'git' somewhere here to add the 'refname's context early on?  Might be helpful to others seeing the function and the names for the first time.

radicle-cob/src/object/collaboration.rs
```diff
@@ -53,3 +53,43 @@ impl CollaborativeObject {
...
+///   * The parsing of the [`ObjectId`] fails
+///   * The parsing of the [`TypeName`] fails
+pub fn parse_refstr<R>(name: &R) -> Option<(TypeName, ObjectId)>
+where
+    R: AsRef<git_ref_format::RefStr>,
+{
+    use git_ref_format::Qualified;
+    let name = name.as_ref();
+    let refs_cobs = match name.to_namespaced() {
+        None => Qualified::from_refstr(name)?,
```
slack-coder, line 85: Good to know the '?' can be used on options too :)

slack-coder: Do you see this being used much elsewhere outside of storage/git.rs where there is an expectation?

It may be better to return an error, saving the caller the need to write a 'badly formatted [something]' here.

radicle-cob/src/object/collaboration.rs
```diff
@@ -53,3 +53,43 @@ impl CollaborativeObject {
...
+        Some(ns) => ns.strip_namespace_recursive(),
+    };
+
+    let (_refs, _cobs, typename, mut object_id) = refs_cobs.non_empty_components();
+    let object = object_id
+        .next()
+        .and_then(|oid| oid.parse::<ObjectId>().ok())?;
+    let name = typename.parse::<TypeName>().ok()?;
+    Some((name, object))
+}
```
slack-coder, line 95: Maybe a small test to prove correctness?  It helps with review

cloudhead: Yes, agree with this, especially beacuse I think there is a bug in the code now that I look at it ;)

radicle-cob/src/object/collaboration.rs
cloudhead: This is a pretty low-level function, so I don't see it used in many places... I personally am fine with `Option` because I wouldn't handle the various errors differently, ie. it's either a valid cob path or it isn't; doesn't really matter which of the potential errors are returned.

FintanH: Ya, I'm inclined to leave it as `Option`. The function is well documented (and tested now) so if there's an issue we can refer to the docs :)

radicle-cob/src/object/collaboration.rs
FintanH: Well I've added a bunch of tests and no bugs showed up, so now I'm curious ;)

cloudhead: :eyes: was just noticing that you're using `ObjectId` when parsing, instead of `git2::Oid`, which is what I had to change it to for it to work!

slack-coder:
LGTM

radicle-cob/src/object/collaboration.rs
FintanH: Ya, the implementation for `FromStr` was wrong for `ObjectId` :) That's what I was pointing out when you were implementing the cobs. I guess I miscommunicated it :sweat_smile:

PR #110, opened 2022-11-16
Add `rad-merge` command

cloudhead:

### DIFF
--- a/radicle-cli/src/commands/merge.rs
+++ b/radicle-cli/src/commands/merge.rs
@@ -1,0 +1,385 @@
+use std::ffi::OsString;
+use std::fmt;
+use std::fmt::Write;
+use std::str::FromStr;
+
+use anyhow::{anyhow, Context};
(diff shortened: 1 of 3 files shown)


radicle-cli/src/commands/merge.rs
```diff
@@ -0,0 +1,385 @@
...
+    //
+    // TODO: Don't allow merging the same revision twice?
+    patches.merge(&patch_id, revision_id, head_oid.into(), &signer)?;
+
+    term::success!(
+        "Patch state updated, use {} to publish",
+        term::format::secondary("`rad push`")
+    );
+
+    Ok(())
```
slack-coder, line 280: Could be split into functions to improve clarity

cloudhead: Let's explore this idea in a pairing session at some point; I don't think it's an issue in the short term.

PR #111, opened 2022-11-16
Add `rad-review` command

cloudhead:

### DIFF
--- a/radicle-cli/src/commands/review.rs
+++ b/radicle-cli/src/commands/review.rs
@@ -1,0 +1,205 @@
+use std::ffi::OsString;
+use std::str::FromStr;
+
+use anyhow::{anyhow, Context};
+
+use radicle::cob;
+use radicle::cob::patch::{PatchId, RevisionIx, Verdict};
+use radicle::prelude::*;
+use radicle::rad;
+
+use crate::terminal as term;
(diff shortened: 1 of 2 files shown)


radicle-cli/src/commands/review.rs
```diff
@@ -0,0 +1,205 @@
...
+    let patch = patches
+        .get(&patch_id)?
+        .context(format!("couldn't find patch {} locally", patch_id))?;
+    let patch_id_pretty = term::format::tertiary(term::format::cob(&patch_id));
+    let revision_ix = options.revision.unwrap_or_else(|| patch.version());
+    let _revision = patch
+        .revisions
+        .get(revision_ix)
+        .ok_or_else(|| anyhow!("revision R{} does not exist", revision_ix))?;
+    let message = options.message.get(REVIEW_HELP_MSG);
```
slack-coder, line 154: The set of let statements are difficult to read.  What level of review would we like here?

I assume most of this are almost direct ports.  Might be best to review them for errors, and iterate after they are included?

cloudhead: I think I could improve that by having multiple sections with comments?

Yes the code is copy-pasted and fixed to compile. I don't expect an indepth review, more like if something looks terribly wrong I'd like to know :)

PR #113, opened 2022-11-18
Migrate API to heartwood

xphoniex:


radicle-httpd/src/api.rs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                "href": "/v1/node",
+                "rel": "node",
+                "type": "GET"
+            },
+            {
+                "href": "/v1/delegates/:did/projects",
+                "rel": "projects",
+                "type": "GET"
+            }
```
xphoniex, line 103: `/v1/delegates/:did/projects`

radicle-httpd/src/api/v1/delegates.rs
```diff
@@ -0,0 +1,70 @@
...
+    let projects = storage
+        .projects()?
+        .into_iter()
+        .filter_map(|id| {
+            let Ok(repo) = storage.repository(id) else { return None };
+            let Ok((_, head)) = repo.head() else { return None };
+            let Ok(Doc { payload, delegates, .. }) = repo.project_of(ctx.profile.id()) else { return None };
+
+            if !delegates.iter().any(|d| d.id == delegate) {
+                return None;
```
xphoniex, line 53: is usage of `project_of` correct here? and check delegates.

radicle-httpd/src/api/v1/delegates.rs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            let Ok(cobs) = Store::open(ctx.profile.public_key, &repo) else { return None };
+            let Ok(issues) = cobs.issues().count() else { return None };
+            let Ok(patches) = cobs.patches().count() else { return None };
+
```
xphoniex, line 59: like before, we're not interrupting the report in case of error and simply skipping problematic items. we could also have defaults at this stage, e.g. return `issues: 0, patches: 0` in case of error.

we can also use a regular loop and return error to api client, which I'm not a fan of. rather keep error loggings on server side.

radicle-httpd/src/api/v1/delegates.rs
```diff
@@ -0,0 +1,70 @@
...
+use crate::api::error::Error;
+use crate::api::Context;
+
+/// Project info.
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Info {
+    /// Project metadata.
+    #[serde(flatten)]
+    pub payload: Payload,
```
xphoniex, line 22: this struct is moved here, previously it was [here](https://github.com/radicle-dev/radicle-client-services/blob/master/http-api/src/project.rs#L20) and `head` was an `Option`, but [`repo.head()` ](https://github.com/radicle-dev/heartwood/pull/113/files#diff-8489cd4775842a934f20ffe06bc2e477f3e21b887fbb4573a88208cd08c4a6ccR49) no longer returns an `Option` so it was changed.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -0,0 +1,331 @@
...
+    Extension(ctx): Extension<Context>,
+    Path((project, sha, path)): Path<(Id, Oid, String)>,
+) -> impl IntoResponse {
+    let path = path.strip_prefix('/').ok_or(Error::NotFound)?.to_string();
+    let storage = &ctx.profile.storage;
+    let repo = storage.repository(project)?;
+    let tree = radicle_surf::object::tree(&repo.raw().into(), Some(Sha { sha }), Some(path))?;
+    let response = json!({
+        "path": &tree.path,
+        "entries": &tree.entries,
```
xphoniex, line 215: would be nice if `radicle_surf::Revision` implements `From<Oid>`

radicle/src/storage.rs
```diff
@@ -184,11 +185,12 @@ impl<V> From<Remotes<V>> for HashMap<RemoteId, Refs> {
...
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
 pub struct Remote<V = Verified> {
     /// ID of remote.
     pub id: PublicKey,
     /// Git references published under this remote, and their hashes.
+    #[serde(flatten)]
     pub refs: SignedRefs<V>,
     /// Whether this remote is a delegate for the project.
     pub delegate: bool,
```
xphoniex, line 196: check

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -0,0 +1,392 @@
...
+    Extension(ctx): Extension<Context>,
+    Path(project): Path<Id>,
+) -> impl IntoResponse {
+    let storage = &ctx.profile.storage;
+    let repo = storage.repository(project)?;
+    let cobs = Store::open(ctx.profile.public_key, &repo)?;
+    let patches = cobs.patches().all()?;
+
+    Ok::<_, Error>(Json(patches))
+}
```
xphoniex, line 321: Original implementation would return `Cob::new(...)` but we don't have that so I'm returning it as is, a `Vec<(ObjectId, Patch)>`.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -0,0 +1,392 @@
...
+    Extension(ctx): Extension<Context>,
+    Path((project, patch_id)): Path<(Id, ObjectId)>,
+) -> impl IntoResponse {
+    let storage = &ctx.profile.storage;
+    let repo = storage.repository(project)?;
+    let cobs = Store::open(ctx.profile.public_key, &repo)?;
+    let patch = cobs.patches().get(&patch_id)?;
+
+    Ok::<_, Error>(Json(patch))
+}
```
xphoniex, line 335: Also note `patch` here is of type `Option<Patch>`, want me to unwrap it? I prefer it with an `Option`, since it can return a `none` in json response.

radicle-httpd/src/api/v1/projects.rs
FintanH: I'm actually going to remove `radicle_surf::Revision` and use the `radicle_surf::git::Revision` trait instead. This means you can use the plain `Oid`, or anything that implements that trait, which a lot of types do :)

radicle-httpd/src/api/v1/projects.rs
cloudhead: I think we should return 404 here if the patch is `None`

radicle-httpd/src/api/v1/projects.rs
cloudhead: yeah that should work.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -0,0 +1,392 @@
...
+        }
+    }
+
+    Err(radicle_surf::object::Error::PathNotFound(
+        radicle_surf::file_system::Path::try_from("README").unwrap(),
+    ))?
+}
+
+/// Get project patches list.
+/// `GET /projects/:project/patches`
```
FintanH, line 310: Are these all patches whether open and closed?

A future suggestion is to also paginate these.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -0,0 +1,392 @@
...
+    let storage = &ctx.profile.storage;
+    let repo = storage.repository(project)?;
+    let cobs = Store::open(ctx.profile.public_key, &repo)?;
+    let patch = cobs.patches().get(&patch_id)?;
+
+    Ok::<_, Error>(Json(patch))
+}
+
+/// Get project issues list.
+/// `GET /projects/:project/issues`
```
FintanH, line 338: Same questions for issues

FintanH (commented):
Just leaving a few comments :)

radicle-httpd/src/api/v1/projects.rs
xphoniex: this is all for now, should we add pagination in this PR as well @cloudhead ?

cloudhead: Let's keep that for another PR :+1:

Cargo.toml
```diff
@@ -32,3 +32,6 @@ version = "0.3.0"
 git = "https://github.com/internet2-wg/rust-cyphernet"
 rev = "dee03a95abe4c964e5d9f8532c7dc76998dfeea7"
 version = "0.1.0"
+
+[patch.crates-io.radicle-git-ext]
+git = "https://github.com/radicle-dev/radicle-git"
```
cloudhead, line 37: We should be using the crates.io version of this, maybe ask fintan to publish a new version if it's out of date.

xphoniex: one of the packages in `radicle-git` (I think surf) depends on its local `radicle-git-ext` and so if we use crates.io version, we might get out of sync. 

either both has to be from crates.io or we pull them from git.

FintanH: We're almost ready to publish `radicle-surf` to crates again. I think once we've merged the current open PRs then we can release `0.10.0` or whatever it is

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -315,7 +305,23 @@ async fn patches_handler(
...
+        .all()?
+        .into_iter()
+        .map(|(id, patch)| {
+            json!({
+                "id": id,
+                "author": patch.author,
+                "title": patch.title,
+                "state": patch.state,
+                "target":patch.target,
+                "labels":patch.labels,
```
cloudhead, line 319: Some whitespace missing :)

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -324,12 +330,25 @@ async fn patches_handler(
...
+        .patches()
+        .get(&patch_id.into())?
+        .ok_or_else(|| Error::NotFound)?;
+    let patch = json!({
+        "id": patch_id,
+        "author": patch.author,
+        "title": patch.title,
+        "state": patch.state,
+        "target":patch.target,
+        "labels":patch.labels,
```
cloudhead, line 348: Same here

radicle-httpd/src/api/v1/projects.rs
xphoniex: ah, `rustfmt` doesn't like closures

xphoniex: yup, macros

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -124,3 +214,56 @@ async fn activity_handler(
...
+            if let Ok(commit) = commit {
+                Some((commit.author.name, commit.author.email))
+            } else {
+                None
+            }
+        })
+        .collect::<BTreeSet<_>>();
+
+    Ok(Stats {
+        branches,
```
xphoniex, line 391: @FintanH I borrowed this code from somewhere in surf I think, why did you not use `HashMap` instead of `BTreeSet`. Wouldn't it be faster?

radicle-httpd/src/api/v1/stats.rs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+/// Return the stats for the node.
+/// `GET /stats`
+async fn stats_handler(Extension(ctx): Extension<Context>) -> impl IntoResponse {
+    let storage = &ctx.profile.storage;
+    let projects = storage.projects()?.len();
+
+    Ok::<_, Error>(Json(
+        json!({ "projects": { "count": projects }, "users": { "count": 0 } }),
```
xphoniex, line 22: `"users": { "count": 0 }` atm. let's come up with a solution for this, or remove it entirely.

radicle-httpd/src/api/error.rs
```diff
@@ -73,16 +75,29 @@ pub enum Error {
...
+            Error::SiweParse(msg) => (StatusCode::BAD_REQUEST, Some(msg.to_string())),
+            Error::SiweVerification(msg) => (StatusCode::BAD_REQUEST, Some(msg.to_string())),
+            Error::Git2(e) => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Some(e.message().to_owned()),
+            ),
+            _ => {
+                tracing::error!("Error: {:?}", &self);
+
+                (StatusCode::INTERNAL_SERVER_ERROR, None)
```
xphoniex, line 92: lmk if there are any other variants which should be handled above.

radicle-httpd/src/api/v1/projects.rs
xphoniex: [added](https://github.com/radicle-dev/heartwood/pull/113/files#diff-85a73839d407aa508b9e78b3756b0e9b6e8d1087a0cafac6a8325a8cf427b65eR322)

radicle-httpd/Cargo.toml
```diff
@@ -16,18 +16,30 @@ logfmt = [
...
 tracing-logfmt = { version = "0.2", optional = true }
-axum = { version = "0.5.16", default-features = false }
-axum-server = { version = "0.4.2", default-features = false }
-hyper = { version = "0.14.17", default-features = false }
-tower-http = { version = "0.3.4", default-features = false, features = ["trace"] }
+tracing-subscriber = { version = "0.3", default-features = false, features = ["std", "ansi", "fmt"] }
+siwe = { version = "0.5" }
+ethers-core = { version = "1.0" }
+serde = { version = "1", features = ["derive"] }
+time = { version = "0.3.17" }
```
cloudhead, line 36: Let's try to keep this list alphabetical

radicle-httpd/src/api.rs
```diff
@@ -0,0 +1,161 @@
...
+impl Context {
+    pub fn new(profile: Arc<Profile>) -> Self {
+        Self {
+            profile,
+            sessions: Default::default(),
+        }
+    }
+
+    pub fn project_info(&self, id: Id) -> Result<project::Info, error::Error> {
+        use radicle::cob::issue::Issues;
```
cloudhead, line 48:
```suggestion
        use radicle::cob::issue::Issues;
        
```

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+#[derive(Serialize)]
+struct Stats {
+    branches: usize,
+    commits: usize,
+    contributors: usize,
+}
+
+fn stats<R: WriteRepository>(repo: &R) -> Result<Stats, Error> {
```
cloudhead, line 381:
```suggestion
fn stats<R: ReadRepository>(repo: &R) -> Result<Stats, Error> {
```
Should be enough?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -0,0 +1,401 @@
...
+            if let Ok(a) = a {
+                let seconds = a.committer.time.seconds();
+                if seconds > current_date - one_year_ago.num_seconds() {
+                    return Some(seconds);
+                }
+
+                None
+            } else {
+                None
+            }
```
cloudhead, line 219: You can remove the `else` branch and always return `None`

radicle-httpd/src/api.rs
xphoniex: ah, this was temporarily placed here. will move it up.

Cargo.lock
```diff
@@ -66,11 +66,23 @@ version = "1.0.66"
...
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
+
+[[package]]
+name = "arrayvec"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
+
```
cloudhead, line 80: Strange that there's duplicates introduced here. Can we see where it's coming from? Also `bitvec` and some others.

radicle-httpd/src/api/v1/projects.rs
FintanH: Do you have any evidence of one being faster than the other? :thinking:

radicle-httpd/src/api/v1/projects.rs
xphoniex: nope
```
error[E0599]: no method named `raw` found for reference `&R` in the current scope
   --> radicle-httpd/src/api/v1/projects.rs:247:25
    |
247 |     let branches = repo.raw().branches(Some(BranchType::Local))?.count();
    |                         ^^^ method not found in `&R`
    |
    = help: items from traits can only be used if the type parameter is bounded by the trait
help: the following traits define an item `raw`, perhaps you need to restrict type parameter `R` with one of them:
    |
246 | fn stats<R: ReadRepository + WriteRepository>(repo: &R) -> Result<Stats, Error> {
    |                            +++++++++++++++++
246 | fn stats<R: ReadRepository + git2::util::Binding>(repo: &R) -> Result<Stats, Error> {
    |                            +++++++++++++++++++++

error[E0599]: no method named `raw` found for reference `&R` in the current scope
   --> radicle-httpd/src/api/v1/projects.rs:250:42
    |
250 |     let contributors = History::new(repo.raw().into(), head)?
    |                                          ^^^ method not found in `&R`
    |
    = help: items from traits can only be used if the type parameter is bounded by the trait
help: the following traits define an item `raw`, perhaps you need to restrict type parameter `R` with one of them:
    |
246 | fn stats<R: ReadRepository + WriteRepository>(repo: &R) -> Result<Stats, Error> {
    |                            +++++++++++++++++
246 | fn stats<R: ReadRepository + git2::util::Binding>(repo: &R) -> Result<Stats, Error> {
    |                            +++++++++++++++++++++

For more information about this error, try `rustc --explain E0599`.
error: could not compile `radicle-httpd` due to 2 previous errors
```

cloudhead: Oh right, we're using `raw()`.

cloudhead (approved):
Nice work.

PR #114, opened 2022-11-18
Some CLI tests & fixes

cloudhead:
*See commits*


radicle-cli/src/terminal/io.rs
```diff
@@ -56,9 +56,10 @@ pub fn tip_args(args: fmt::Arguments) {
     );
 }
 
-pub fn width() -> usize {
-    let (_, rows) = console::Term::stdout().size();
-    rows as usize
+pub fn width() -> Option<usize> {
+    console::Term::stdout()
+        .size_checked()
+        .map(|(_, cols)| cols as usize)
```
slack-coder, line 62: What happens now if no width is returned? We continue as if its infinite?

radicle-cli/src/terminal/textbox.rs
```diff
@@ -39,8 +39,10 @@ impl fmt::Display for TextBox {
             .max()
             .unwrap_or(0)
             + 2;
-        if term::width() < width + 2 {
-            width = term::width() - 2
+        if let Some(max) = term::width() {
+            if max < width + 2 {
+                width = max - 2
+            }
```
slack-coder, line 45: It appears width is the subject therefor should it be

```
if let Some(max) = term::width() {
  if max < width + 2 {
    width = max - 2
  }
}
```

slack-coder:
I like the idea of trycmd

radicle-crypto/src/lib.rs
```diff
@@ -356,6 +356,23 @@ impl sqlite::Bindable for &PublicKey {
...
+    use std::env;
+
+    /// Generate a new keypair using OS randomness.
+    pub fn generate() -> KeyPair {
+        #[cfg(debug_assertions)]
+        if env::var("RAD_DEBUG").is_ok() {
+            // Generate a test keypair that is always the same.
+            // This is useful for debugging and testing, since the
+            // public key is known in advance.
+            return KeyPair::from_seed(Seed::new([0xff; 32]));
```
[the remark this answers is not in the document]
cloudhead: Notice that it's only when running with the `RAD_DEBUG` env var. So it won't be triggered during normal testing behavior. It's mainly meant for interactive testing. I also don't think we would use a keystore at all when testing multiple nodes, we could just use a `MockSigner`.

radicle-cli/src/terminal/io.rs
cloudhead: Yeah exactly. We just don't wrap lines. I think the reason this might be happening is when running the command not via a tty, eg. in the tests.

radicle-cli/src/terminal/textbox.rs
cloudhead: Yes, it tripped me up actually, but didn't think it worth changing. I'll make the change.

PR #115, opened 2022-11-18
node: try using static Vector limits

slack-coder:
Support clearly defining message constraints by introducing a new vector type.  The vector type takes a statically defined upper bound as a Type parameter.

Should we pursue using this?  The code is to express the idea and not meant as a final version, some clean up will be in order.

```
pub struct InventoryAnnouncement {
    /// Node inventory.
    pub inventory: BoundedVec<INVENTORY_LIMIT, Id>,
    /// Time of announcement.
    pub timestamp: Timestamp,
}
```

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -29,6 +29,7 @@ use crate::crypto::{Signer, Verified};
 use crate::git;
 use crate::identity::{Doc, Id};
 use crate::node;
+use crate::prelude::*;
 use crate::service::config::ProjectTracking;
 use crate::service::message::{Address, Announcement, AnnouncementMessage, Ping};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
@@ -66,6 +67,11 @@ pub const KEEP_ALIVE_DELTA: LocalDuration = LocalDuration::from_secs(30);
 pub const MAX_TIME_DELTA: LocalDuration = LocalDuration::from_mins(60);
 /// Maximum attempts to connect to a peer before we give up.
 pub const MAX_CONNECTION_ATTEMPTS: usize = 3;
+
+/// Maximum external address limit imposed by message size limits.
+pub use message::ADDRESS_LIMIT;
+/// Maximum inventory limit imposed by message size limits.
+pub use message::INVENTORY_LIMIT;
 
 /// A service event.
 #[derive(Debug, Clone)]
@@ -639,9 +645,11 @@ where
                     return Ok(false);
                 }
 
-                if let Err(err) =
-                    self.process_inventory(&message.inventory, *announcer, &message.timestamp)
-                {
+                if let Err(err) = self.process_inventory(
+                    message.inventory.as_slice(),
+                    *announcer,
+                    &message.timestamp,
+                ) {
                     error!("Error processing inventory from {}: {}", announcer, err);
 
                     if let Error::Fetch(storage::FetchError::Verify(err)) = err {
@@ -797,7 +805,7 @@ where
                 peer.state = session::State::Negotiated {
                     id,
                     since: self.clock.local_time(),
-                    addrs,
+                    addrs: addrs.unbound(),
                     ping: Default::default(),
                 };
             }
@@ -875,7 +883,7 @@ where
     /// Process a peer inventory announcement by updating our routing table.
     fn process_inventory(
         &mut self,
-        inventory: &Vec<Id>,
+        inventory: &[Id],
         from: NodeId,
         timestamp: &Timestamp,
     ) -> Result<(), Error> {
@@ -1269,7 +1277,14 @@ mod gossip {
         };
 
         let mut msgs = vec![
-            Message::init(*signer.public_key(), config.external_addresses.clone()),
+            Message::init(
+                *signer.public_key(),
+                config
+                    .external_addresses
+                    .clone()
+                    .try_into()
+                    .expect("external addresses are within the limit"),
+            ),
             Message::inventory(gossip::inventory(timestamp, inventory), signer),
             Message::subscribe(config.filter(), timestamp, Timestamp::MAX),
         ];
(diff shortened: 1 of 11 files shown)


FintanH:
I'm into representing these things at the type level :grin: 

It was also used in `radicle-link` and there was a helper crate for these things https://github.com/radicle-dev/radicle-link/blob/master/data/src/bounded.rs. You might want to use that.

dr-orlovsky:
I think I have guys what you need here: https://github.com/rust-amplify/rust-amplify/pull/155

@cloudhead was going to use that amplify crate in anyway as a dependency.

cloudhead:
Yeah at the very list we should be using `const` generics for the size limits, not regular generics. The `Confined` type seems pretty much what we need, if we're going to be depending on `amplify` anyway after the transport merge..

dr-orlovsky:
@cloudhead I have completed and released that implementation as a part of `amplify = 4.0.0-alpha.1`. You can see the whole final implementation here: https://github.com/rust-amplify/rust-amplify/blob/master/src/confinement.rs

The list of exported simplified newtype names to drop-in usage is this:
https://docs.rs/amplify/4.0.0-alpha.1/amplify/confinement/index.html#types

https://github.com/rust-amplify/rust-amplify/blob/master/src/confinement.rs#L677-L752

slack-coder:
I've revamped the custom type with some inspiration from radicle-links type.  I opted for the custom type because importing the module to use const generics was more difficult than expected.

A separate PR can be set up for changing the bounded vector type if and when necessary.

radicle-node/src/service.rs
```diff
@@ -1304,8 +1323,17 @@ mod gossip {
     }
 
     pub fn inventory(timestamp: Timestamp, inventory: Vec<Id>) -> InventoryAnnouncement {
+        type Inventory = BoundedVec<Id, INVENTORY_LIMIT>;
+
+        if inventory.len() > Inventory::max() {
+            log::error!(
+                "inventory announcement limit ({}) exceeded, other nodes will see only some of you projects",
```
cloudhead, line 1330:
```suggestion
                "inventory announcement limit ({}) exceeded, other nodes will see only some of your projects",
```

radicle-node/src/service/message.rs
```diff
@@ -497,9 +503,30 @@ impl ZeroBytes {
...
+                    .expect("size within bounds limit"),
+                timestamp: LocalTime::now().as_secs(),
+            },
+            &MockSigner::default(),
+        );
+        let mut buf: Vec<u8> = Vec::new();
+        assert!(
+            msg.encode(&mut buf).is_ok(),
+            "INVENTORY_LIMIT is too big to support message encoding",
+        );
```
cloudhead, line 528: Can you add an assert for decoding too?

radicle-node/src/tests.rs
```diff
@@ -598,7 +615,8 @@ fn test_inventory_relay() {
     let mut alice = Peer::new("alice", [7, 7, 7, 7], MockStorage::empty());
     let bob = Peer::new("bob", [8, 8, 8, 8], MockStorage::empty());
     let eve = Peer::new("eve", [9, 9, 9, 9], MockStorage::empty());
-    let inv = vec![];
+    // TODO: remove BoundedVec const integral when it becomes conferable
```
cloudhead, line 618:
```suggestion
```

radicle-node/src/wire/message.rs
```diff
@@ -18,7 +19,6 @@ pub enum MessageType {
     Ping = 10,
     Pong = 12,
 }
-
```
cloudhead, line 21: :eyes:

radicle-node/src/service.rs
```diff
@@ -1285,7 +1300,11 @@ mod gossip {
     pub fn node(timestamp: Timestamp, config: &Config) -> Option<NodeAnnouncement> {
         let features = node::Features::SEED;
         let alias = config.alias();
-        let addresses = config.external_addresses.clone();
+        let addresses: BoundedVec<_, ADDRESS_LIMIT> = config
+            .external_addresses
+            .clone()
+            .try_into()
+            .expect("verified config parameter size");
```
cloudhead, line 1307:
```suggestion
            .expect("external addresses are within the limit");
```

radicle-node/src/service.rs
```diff
@@ -1271,7 +1279,14 @@ mod gossip {
...
 
         let mut msgs = vec![
-            Message::init(*signer.public_key(), config.external_addresses.clone()),
+            Message::init(
+                *signer.public_key(),
+                config
+                    .external_addresses
+                    .clone()
+                    .try_into()
+                    .expect("verified config parameter size"),
```
cloudhead, line 1288:
```suggestion
            .expect("external addresses are within the limit");
```

radicle-node/src/service/message.rs
slack-coder: Sure, though does the property test for encoding/decoding cover this already?

https://github.com/radicle-dev/heartwood/blob/master/radicle-node/src/wire/message.rs#L405

slack-coder: In fact, the property test should cover both cases already right?

radicle-node/src/tests.rs
slack-coder: Looks like the constant was not needed.

slack-coder:
The code has been updated as suggested

radicle-node/src/service/message.rs
cloudhead: It doesn't (see unwrap) -- it only produces valid messages.

slack-coder: I went ahead and implemented it while waiting on the answer.

cloudhead: I must be missing something, I don't see where this is tested?

slack-coder: Indeed, its annoying.  Clicking on the file link of the comment points to a specific commit.   I thought I forgot to push the changes.  Maybe because the comment was on the specific commit?

Here are the relevant changes: https://github.com/radicle-dev/heartwood/pull/115/commits/0f7543799bd9c333344e20091a96bb3456638f3b

cloudhead: I saw those changes, but they don't test that decoding fails with a too big payload.

cloudhead:
```rust
        assert!(
            msg.encode(&mut buf).is_ok(),
            "INVENTORY_LIMIT is too big to support message encoding",
        );
```
The message here also doesn't make sense, it should say `"the inventory size is smaller than the limit"`

cloudhead (approved):
Perfect.

PR #119, opened 2022-11-22
Handshake accept/init

cloudhead:
Co-authored-by: Dr. Maxim Orlovsky <orlovsky@pandoraprime.ch>

### DIFF
--- a/radicle-node/src/service/message.rs
+++ b/radicle-node/src/service/message.rs
@@ -56,6 +56,19 @@ impl From<net::SocketAddr> for Address {
     }
 }
 
+impl Address {
+    pub fn to_socket_addr(&self) -> net::SocketAddr {
+        match self {
+            Self::Ipv4 { ip, port } => net::SocketAddrV4::new(*ip, *port).into(),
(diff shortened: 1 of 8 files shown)


radicle-node/src/service/message.rs
```diff
@@ -56,6 +56,19 @@ impl From<net::SocketAddr> for Address {
     }
 }
 
+impl Address {
+    pub fn to_socket_addr(&self) -> net::SocketAddr {
```
FintanH, line 60: suggestion: `impl From<Address> for net::SocketAddr`

cloudhead: Yeah will consider that; I think we might need an extra param in there, let's see.

dr-orlovsky (commented):
This looks ok except that I do not understand how this will work with the inbound connections. I mean I do not understand the inbound connection workflow in the current heartwood implementation - and while I see handing of `Link::Inbound` in the code of the PR, I do not see that `Address` cover the inbound connectivity - and how the cases of `Link::Inbound` will ever be called at all...

PR #122, opened 2022-11-23
node: Fix addresses sent to peers on conn.

slack-coder:
Send the node's configured external addresses instead of its listenning addresses.  External addresses are ones which the service is reachable from the outside world, the listening addresses can be the wrong context (local area network, or 0.0.0.0).


slack-coder:
The issue of populating the external-address list is still open.  As of now, it should be manually configured.

PR #127, opened 2022-11-28
rad-issue: Change `reaction` flag to `emoji`

xphoniex:
This is poor UX imo and I'd prefer it if we show the emojis in terminal somehow (like a TUI). Still the flag needs to be renamed to match the `help`.

### DIFF
--- a/radicle-cli/src/commands/issue.rs
+++ b/radicle-cli/src/commands/issue.rs
@@ -123,7 +123,7 @@ impl Args for Options {
                         reason: CloseReason::Solved,
                     });
                 }
-                Long("reaction") if op == Some(OperationName::React) => {
+                Long("emoji") if op == Some(OperationName::React) => {
                     if let Some(emoji) = parser.value()?.to_str() {
                         reaction =
                             Some(Reaction::from_str(emoji).map_err(|_| anyhow!("invalid emoji"))?);


sebastinez:
The idea behind `reaction` was for users to react to issue comments.
Not sure `emoji` is the right name for the flag..
Maybe `react` instead of `reaction`?

xphoniex:
> The idea behind `reaction` was for users to react to issue comments. So not sure `emoji` is the right name for the flag, what does `emoji` imply? That you can set an emoji onto a comment? Maybe `react` instead of `reaction`?

In the [help](https://github.com/radicle-dev/radicle-cli/blame/master/issue/src/lib.rs#L22) section of the original implementation, you've written the flag as `--emoji` and it's supposed to take a `<char>`. 

Also errors are mapped to msg [`"invalid emoji"`](https://github.com/radicle-dev/radicle-cli/blame/master/issue/src/lib.rs#L115) , and here user is asked to provide an [emoji](https://github.com/radicle-dev/radicle-cli/blame/master/issue/src/lib.rs#L155) (`"a reaction emoji must be provided"`) as well.

Let me know which one is the correct one (`emoji` or `reaction`) so I can rename all occurrences.

sebastinez:
Hey @xphoniex thanks for looking it up!
Yeah I think you are right we have to straighten things up, I would go with `reaction` everywhere.

PR #136, opened 2022-12-01
node: Document and test BoundedVec

slack-coder:
Borrow documentation from Rusts standard library to document BoundedVec's functions, and add documentation tests where appropriate.

And make it's deref trait return a slice instead of a Vec, which is more common in rust.

### DIFF
--- a/radicle-node/src/bounded.rs
+++ b/radicle-node/src/bounded.rs
@@ -6,40 +6,53 @@ pub enum Error {
     InvalidSize { expected: usize, actual: usize },
 }
 
+/// A vector with an upper limit on its size using type level constants.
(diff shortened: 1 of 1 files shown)


radicle-node/src/bounded.rs
```diff
@@ -6,25 +6,107 @@ pub enum Error {
...
+                expected: N,
+                actual: capacity,
+            });
+        }
+        Ok(Self {
+            v: Vec::with_capacity(capacity),
+        })
+    }
+
+    /// Return the maximum number of elements BoundedVec can contain.
```
cloudhead, line 68:
```suggestion
    /// Return the maximum number of elements [`BoundedVec`] can contain.
```

cloudhead (approved):

PR #139, opened 2022-12-04
Replace `quickcheck` with `qcheck`

cloudhead:
The latter is a fork of the former, which adds const generics and a bunch of other quality of life and performance improvements that aren't in the original crate due to it being unmaintained.

I'm maintaing the fork here: https://github.com/cloudhead/qcheck


FintanH (approved):
Nice :+1:

slack-coder:
@cloudhead where is the donate link? :)

PR #140, opened 2022-12-05
CLI testing framework

cloudhead:
I switched from `trycmd` to implementing something custom on top of `snapbox` directly, as `trycmd` was much too limiting.

The goal is to be able to write elaborate multi-user flows as if it were documentation, but have it run like a test. For now this is just a replacement of functionality of `trycmd`, but it will be pretty easy to add more custom functionality towards our needs.


FintanH (approved):
This is cool!

PR #141, opened 2022-12-05
node: Limit number of Refs in RefAnnouncements

### LINKED ISSUE #16: Check all lengths when decoding
To prevent OOM errors, we should check all lengths in the `wire` protocol. Currently, lengths are capped to 16bits, but for certain types of data, that could still be a lot, eg. when sending over public keys or hashes.

slack-coder:
Use the BoundedVec to limit the supported number of refs in RefAnnouncement messages.  The size (236) is the maximum supported by message wire size.

To reduce the complexity of the changes, BoundedVec is used again instead of using a new bounded Refs or BTreeMap.  This requires defining new Decode/Encode logic for Ref's element types, and workarounds for arbitrary instances.

closes #16

### DIFF
--- a/radicle-node/src/bounded.rs
+++ b/radicle-node/src/bounded.rs
@@ -18,6 +18,25 @@ impl<T, const N: usize> BoundedVec<T, N> {
         BoundedVec { v: Vec::new() }
     }
 
+    /// Build a `BoundedVec` by consuming from the given iterator up to its limit.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use radicle_node::bounded;
+    ///
+    /// let mut iter = (0..4).into_iter();
+    /// let bounded: bounded::BoundedVec<i32,3> = bounded::BoundedVec::collect_from(&mut iter);
+    ///
+    /// assert_eq!(bounded.len(), 3);
+    /// assert_eq!(iter.count(), 1);
+    /// ```
+    pub fn collect_from<I: Iterator<Item = T>>(iter: &mut I) -> Self {
+        BoundedVec {
+            v: iter.into_iter().take(N).collect(),
+        }
+    }
+
     /// Create a new `BoundedVec<T,N>` which takes upto the first N values of its argument, taking
     /// ownership.
     ///
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -73,6 +73,8 @@ pub const MAX_CONNECTION_ATTEMPTS: usize = 3;
 pub use message::ADDRESS_LIMIT;
 /// Maximum inventory limit imposed by message size limits.
 pub use message::INVENTORY_LIMIT;
+/// Maximum number of project git references imposed by message size limits.
+pub use message::REF_LIMIT;
 
 /// A service event.
 #[derive(Debug, Clone)]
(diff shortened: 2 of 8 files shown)


radicle-node/src/bounded.rs
```diff
@@ -18,6 +18,34 @@ impl<T, const N: usize> BoundedVec<T, N> {
...
+    /// ```
+    /// use radicle_node::bounded;
+    ///
+    /// let mut iter = (0..4).into_iter();
+    /// let bounded: bounded::BoundedVec<i32,3> = bounded::BoundedVec::collect_from(&mut iter);
+    ///
+    /// assert_eq!(bounded.len(), 3);
+    /// assert_eq!(iter.count(), 1);
+    /// ```
+    pub fn collect_from<I: Iterator<Item = T>>(iter: &mut I) -> Self {
```
FintanH, line 34: A couple of things:
1. This is the signature of `FromIterator::from_iter` so you can implement that and we can use `.collect()`
2. Perhaps it makes more sense to use:
```rust
BoundedVec { v: iter().into_iter().tak(Self::max()).collect() }
```
3. If we have a standalone function, i.e. not `from_iter`, perhaps we could return the rest of the iterator so that the caller can decide what to do with the rest of the iterator.

radicle-node/src/service.rs
```diff
@@ -905,12 +907,22 @@ where
...
+        type Refs = BoundedVec<Id, REF_LIMIT>;
+
         let node = self.node_id();
         let repo = self.storage.repository(id)?;
         let remote = repo.remote(&node)?;
         let peers = self.sessions.negotiated().map(|(_, _, p)| p);
-        let refs = remote.refs.into();
         let timestamp = self.clock.timestamp();
+
+        if remote.refs.len() > Refs::max() {
```
FintanH, line 918: Why do we not want to send multiple announcements instead? Dropping refs seems strange to me.

cloudhead: True, for InventoryAnnouncement we couldn't do that because it's supposed to be the entire inventory, but for refs I think we could? Would have to think through this a bit more..

radicle-node/src/bounded.rs
slack-coder: "This is the signature of FromIterator::from_iter so you can implement that and we can use .collect()"

Should collect() require collecting all items into the new collection?  I was unsure because of this.  One thing I miss about haskell is clearly defined rules and expectation of type classes.

FintanH: Oh wait, I'm looking at this again and realise you're taking an `&mut I`, which isn't the same as `FromIterator`! Sorry, was looking at this last night wait for a flight home :sweat_smile: This makes sense and looks good to me :+1:

radicle-node/src/wire.rs
```diff
@@ -266,6 +284,25 @@ impl Decode for Refs {
...
+    A: Decode,
+    B: Decode,
+{
+    fn decode<R: io::Read + ?Sized>(reader: &mut R) -> Result<Self, Error> {
+        let a = A::decode(reader)?;
+        let b = B::decode(reader)?;
+        Ok((a, b))
+    }
+}
+
```
cloudhead, line 305: Roundtrip test for this plz :)

radicle-node/src/bounded.rs
cloudhead: Can this function not take an `I` though?

slack-coder: > Can this function not take an owned `I` though?

Would this not make it more restrictive?

FintanH: I think it makes more sense to take `&mut` because, as @slack-coder pointed out to me, you're left with the unconsumed portion of the iterator. This is a good thing from my perspective. Say if we did do multiple ref announcements that are windowed by this `236` limit, then you would want to repeat the `BoundedVec` construction until the iterator is consumed.

cloudhead: Oh, gotcha :+1:

cloudhead (approved):

PR #142, opened 2022-12-05
cobs: allow for batched operations

FintanH:
Sometimes it is more efficient to perform multiple operations at once and store those in the collaborative object graph together.

Change the contents to be a non-empty sequence of byte sequences, i.e. NonEmpty<Vec<u8>>. These are stored as blobs in a git tree in order by their index in the non-empty sequence. They are, in turn, retrieved in order of their original index.

Change the conversion of an `EntryWithClock` to return a non-empty sequence of `Op`s, which are in turn used to apply many changes in each `from_history` implementation.

### DIFF
--- a/radicle-cob/Cargo.toml
+++ b/radicle-cob/Cargo.toml
@@ -17,6 +17,7 @@ git-commit = { version = "0.2" }
 git-ref-format = { version = "0.1" }
 git-trailers = { version = "0.1" }
 log = { version = "0.4.17" }
+nonempty = { version = "0.8.1", features = ["serialize"] }
 petgraph = { version = "0.5" }
 radicle-git-ext = { version = "0.2" }
 serde_json = { version = "1.0" }
--- a/radicle/src/cob/op.rs
+++ b/radicle/src/cob/op.rs
@@ -1,5 +1,6 @@
 use std::collections::BTreeMap;
 
+use nonempty::NonEmpty;
 use thiserror::Error;
 
 use radicle_cob::history::EntryWithClock;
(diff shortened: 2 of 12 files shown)


radicle-cob/Cargo.toml
```diff
@@ -17,6 +17,7 @@ git-commit = { version = "0.2" }
 git-ref-format = { version = "0.1" }
 git-trailers = { version = "0.1" }
 log = { version = "0.4.17" }
+nonempty = { version = "0.8.1", git = "https://github.com/cloudhead/nonempty.git", rev = "cdf611bab2bc37ff9c443067638324fb32a21f2e", features = ["serialize"] }
```
cloudhead, line 20: Published it :)

radicle/src/cob/op.rs
```diff
@@ -35,18 +36,31 @@ pub struct Op<A> {
...
+            .clone()
+            .try_map(|op| {
+                let action = serde_json::from_slice(&op)?;
+                let op = Op {
+                    action,
+                    author: *entry.actor(),
+                    clock,
+                    timestamp: entry.timestamp().into(),
+                };
+                clock = clock.tick();
```
cloudhead, line 60: `clock.tick()` should suffice, as it mutates the internal value as well.

cloudhead:
Looks good!

radicle-cob/Cargo.toml
slack-coder: Is there a reason for the revision to be specified?  It may impact how cargo uses semantic versioning to apply fixes.  When and if a fix comes out, say version 0.8.2, cargo could build the project with the latest patched version.

slack-coder:
Otherwise :+1:

radicle/src/cob/op.rs
FintanH: Ah, what's the reason for returning `Self` then? I found that a bit confusing

radicle-cob/Cargo.toml
FintanH: It will build with `0.8.2` but this says the minimum is `0.8.1` which is true since we're using methods that only exist there.

FintanH:
> I took the liberty to rebase and fix the nonempty dependency and clock.tick()

Thanks! Appreciate it :)

radicle-cob/Cargo.toml
slack-coder: @FintanH my concern is about the 'rev = "cdf611bab2bc37ff9c443067638324fb32a21f2e"' affecting this.  Would it not?

FintanH: Right, sorry that was temporary as I had opened a PR on `nonempty`. @cloudhead fixed this up for me before merging.

radicle/src/cob/op.rs
cloudhead: I often used it like `New { clock: old.clock.tick() }`, was just convenience

cloudhead: I think we can change that to only return but not mutate, since we use the DAG clock now.

PR #143, opened 2022-12-06
Move onto new reactor architecture

dr-orlovsky:


radicle-node/src/wire/new.rs
```diff
@@ -0,0 +1,273 @@
...
+            }
+            Some(Io::Connect(id, addr)) => {
+                let NetAddr { host, port } = *addr;
+                let socket_addr = match host {
+                    HostAddr::Ip(ip) => net::SocketAddr::new(ip, addr.port()),
+                    HostAddr::Dns(_) => todo!(),
+                    _ => self.proxy,
+                };
+
+                self.inner.attempted(&socket_addr);
```
cloudhead, line 245: We shouldn't call this here, it should only be called if we are able to create the socket. I guess it's replaced by `connecting` anyway, which we didn't have before?

radicle-node/src/wire/new.rs
```diff
@@ -0,0 +1,273 @@
...
+                    }
+                    Err(err) => {
+                        self.inner
+                            .disconnected(&id, &DisconnectReason::DialError(Arc::new(err)));
+                        return None;
+                    }
+                }
+            }
+            Some(Io::Disconnect(id, r)) => {
+                self.inner.disconnected(&id, &DisconnectReason::Protocol(r));
```
cloudhead, line 264: Same here we should only call this when something actually happened, and not in the same function that is reading from the inner service.

cloudhead (commented):
Concept ACK

radicle-node/Cargo.toml
```diff
@@ -7,14 +7,19 @@ edition = "2021"
 
 [features]
 test = ["radicle/test", "radicle-crypto/test", "qcheck"]
+new_arch = ["poll-reactor", "netservices"]
```
slack-coder, line 10: What about implementing it directly into the project in place of making it an optional feature?

cloudhead: It will be, but it's easier to review this way.

dr-orlovsky:
I have added business logic dealing with all possible exceptions (disconnecting unconnected peers etc) to the new Wire.

@cloudhead pls review https://github.com/radicle-dev/heartwood/pull/143/commits/4ee66a2644e555d7d4d1facfc4e0650ca4fddf54

dr-orlovsky:
Superceded by #162

PR #147, opened 2022-12-06
node: fix inconsistently failing test

slack-coder:
Fix the test for pruning a node's inventory which inconsistently and rarely fails.  It fails due to each test peers' clocks being slightly out of sync in the test.  One peer may have their clock skewed one second before the other.

The test confirms how pruning inventory entries relates to time.  Time therefor needs to be accurate to test against internal edge cases.

Note:

I would prefer to sync the clocks manually, the alternative of using an 'if' and 'elapsed' combination seems a bit much.

i.e.
```
bob.clock().set(alice.clock().local_time());
```

### DIFF
--- a/radicle-node/src/test/peer.rs
+++ b/radicle-node/src/test/peer.rs
@@ -15,7 +15,6 @@ use crate::identity::Id;
 use crate::node;
 use crate::prelude::NodeId;
 use crate::service;
-use crate::service::config::*;
 use crate::service::message::*;
 use crate::service::reactor::Io;
 use crate::service::*;
@@ -23,6 +22,7 @@ use crate::storage::git::transport::remote;
 use crate::storage::{RemoteId, WriteStorage};
 use crate::test::arbitrary;
 use crate::test::simulator;
+use crate::test::storage::MockStorage;
 use crate::{Link, LocalDuration, LocalTime};
 
 /// Service instantiation used for testing.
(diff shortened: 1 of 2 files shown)


FintanH:
One question this prompts for me is whether this is just fixing the test rather than fixing the actual problem in the code? I'm not sure of the answer because I'm not sure I understand the issue :)

slack-coder:
> One question this prompts for me is whether this is just fixing the test rather than fixing the actual problem in the code? I'm not sure of the answer because I'm not sure I understand the issue :)

The PR comment and commit message needed some tending to.  Is the new description clear?

FintanH:
I _think_ so. I suppose another way of phrasing my question is, "Can this clock skew happen in practice? And is the failure correct in that case?"

slack-coder:
The test is for confirming whether internal mechanisms work as expected using the public api.

FintanH (approved):

slack-coder:
> I _think_ so. I suppose another way of phrasing my question is, "Can this clock skew happen in practice? And is the failure correct in that case?"

Thanks for the review :+1:.  It does feel off using one peer's timestamp for another's message.  I'll try addressing this in this PR.  Upcoming tests may run into the same issue, so it is good to address.

slack-coder:
@FintanH I made some significant changes

cloudhead (approved):

cloudhead:
Nice.

FintanH:
Nice one :) Much better!

PR #149, opened 2022-12-06
dogfood the new radicle-surf API

keepsimple1:
This is to play around the new `radicle-surf` API and see if it can meet the existing user's needs. The goal is to seek feedbacks, not only about the usages in the diff, but also about what is not covered but necessary.

### DIFF
--- a/radicle-httpd/src/api/error.rs
+++ b/radicle-httpd/src/api/error.rs
@@ -38,17 +38,13 @@ pub enum Error {
     #[error(transparent)]
     GitProject(#[from] radicle::storage::git::ProjectError),
 
-    /// Surf commit error.
+    /// Surf directory error.
     #[error(transparent)]
-    SurfCommit(#[from] radicle_surf::source::commit::Error),
+    SurfDir(#[from] radicle_surf::fs::error::Directory),
 
-    /// Surf object error.
+    /// Surf error.
     #[error(transparent)]
-    SurfObject(#[from] radicle_surf::source::object::Error),
-
-    /// Surf git error.
-    #[error(transparent)]
-    SurfGit(#[from] radicle_surf::git::Error),
+    Surf(#[from] radicle_surf::Error),
 
     /// Git2 error.
     #[error(transparent)]
--- a/radicle-httpd/src/api/json.rs
+++ b/radicle-httpd/src/api/json.rs
@@ -1,0 +1,68 @@
+//! Utilities for building JSON responses of our API.
+
+use radicle_surf::{
+    object::{Blob, Tree},
+    Commit, Stats,
+};
+use serde_json::json;
+
+/// Returns JSON of a commit.
+pub(crate) fn commit(commit: &Commit) -> serde_json::Value {
+    json!({
+      "sha1": commit.id,
+      "author": {
+        "name": commit.author.name,
+        "email": commit.author.email
+      },
+      "summary": commit.summary,
+      "description": commit.description(),
+      "committer": {
+        "name": commit.committer.name,
+        "email": commit.committer.email
+      },
+      "committerTime": commit.committer.time.seconds()
+    })
+}
+
+/// Returns JSON for a blob with a given `path`.
+pub(crate) fn blob(blob: &Blob, path: &str) -> serde_json::Value {
+    json!({
+        "binary": blob.is_binary(),
+        "content": blob.content(),
+        "name": name_in_path(path),
+        "path": path,
+        "lastCommit": commit(blob.commit())
+    })
+}
+
+/// Returns JSON for a tree with a given `path` and `stats`.
+pub(crate) fn tree(tree: &Tree, path: &str, stats: &Stats) -> serde_json::Value {
+    let prefix = std::path::Path::new(path);
+    let entries = tree
+        .entries()
+        .iter()
+        .map(|entry| {
+            json!({
+                "path": prefix.join(entry.name()),
+                "name": entry.name(),
+                "lastCommit": serde_json::Value::Null,
+                "kind": if entry.is_tree() { "tree" } else { "blob" },
+            })
+        })
+        .collect::<Vec<_>>();
+    json!({
+        "entries": &entries,
+        "lastCommit": commit(tree.commit()),
+        "name": name_in_path(path),
+        "path": path,
+        "stats": stats,
+    })
+}
+
+/// Returns the name part of a path string.
+fn name_in_path(path: &str) -> &str {
+    match path.rsplit('/').next() {
+        Some(name) => name,
+        None => path,
+    }
+}
--- a/radicle-httpd/src/api/v1/projects.rs
+++ b/radicle-httpd/src/api/v1/projects.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashSet};
+use std::collections::BTreeMap;
 
 use axum::handler::Handler;
 use axum::http::{header, HeaderValue};
@@ -13,20 +13,15 @@ use tower_http::set_header::SetResponseHeaderLayer;
 use radicle::cob::issue::Issues;
 use radicle::cob::thread::{self, CommentId};
 use radicle::cob::Timestamp;
-use radicle::git::Oid;
 use radicle::identity::{Id, PublicKey};
 use radicle::node::NodeId;
-use radicle::storage::git::paths;
-use radicle::storage::{ReadRepository, WriteStorage};
-use radicle_surf::git::{Glob, History, Repository};
-use radicle_surf::source;
-use radicle_surf::source::object::blob::Blob;
-use radicle_surf::source::object::tree::Tree;
+use radicle::storage::{git::paths, ReadRepository, WriteStorage};
+use radicle_surf::{Glob, Oid, Repository};
 
 use crate::api::axum_extra::{Path, Query};
 use crate::api::error::Error;
 use crate::api::project::Info;
-use crate::api::{Context, PaginationQuery};
+use crate::api::{self, Context, PaginationQuery};
 
 const CACHE_1_HOUR: &str = "public, max-age=3600, must-revalidate";
 
@@ -147,7 +142,7 @@ async fn history_handler(
     };
 
     let headers = repo
-        .history(sha)?
+        .history(&sha)?
         .filter(|q| {
             if let Ok(q) = q {
                 if let (Some(since), Some(until)) = (since, until) {
@@ -166,14 +161,23 @@ async fn history_handler(
         })
         .skip(page * per_page)
         .take(per_page)
-        .filter_map(|commit| {
-            if let Ok(commit) = commit {
-                source::commit(&repo, commit.id).ok()
-            } else {
-                None
-            }
+        .map(|r| {
+            r.and_then(|c| {
+                let glob = Glob::all_heads().branches().and(Glob::all_remotes());
+                let branches: Vec<String> = repo
+                    .revision_branches(c.id, glob)?
+                    .iter()
+                    .map(|b| b.refname().to_string())
+                    .collect();
+                let diff = repo.diff_commit(c.id)?;
+                Ok(json!({
+                    "header": api::json::commit(&c),
+                    "diff": diff,
+                    "branches": branches
+                }))
+            })
         })
-        .collect::<Vec<_>>();
+        .collect::<Result<Vec<_>, _>>()?;
 
     let response = json!({
         "headers": headers,
@@ -195,9 +199,22 @@ async fn commit_handler(
 ) -> impl IntoResponse {
     let storage = &ctx.profile.storage;
     let repo = Repository::open(paths::repository(storage, &project))?;
-    let commit = source::commit(&repo, sha)?;
+    let commit = repo.commit(sha)?;
 
-    Ok::<_, Error>(Json(commit))
+    let diff = repo.diff_commit(commit.id)?;
+    let glob = Glob::all_heads().branches().and(Glob::all_remotes());
+    let branches: Vec<String> = repo
+        .revision_branches(commit.id, glob)?
+        .iter()
+        .map(|b| b.refname().to_string())
+        .collect();
+
+    let response = json!({
+      "header": api::json::commit(&commit),
+      "diff": diff,
+      "branches": branches
+    });
+    Ok::<_, Error>(Json(response))
 }
 
 /// Get project activity for the past year.
@@ -236,10 +253,9 @@ async fn tree_handler(
     let path = path.strip_prefix('/').ok_or(Error::NotFound)?.to_string();
     let storage = &ctx.profile.storage;
     let repo = Repository::open(paths::repository(storage, &project))?;
-    let path = if path.is_empty() { None } else { Some(&path) };
-    let tree = Tree::new(&repo, &sha, path)?;
-    let mut response = json!(tree);
-    response["stats"] = json!(stats(&repo, sha)?);
+    let tree = repo.tree(sha, &path)?;
+    let stats = repo.stats_from(&sha)?;
+    let response = api::json::tree(&tree, &path, &stats);
 
     Ok::<_, Error>(Json(response))
 }
@@ -313,9 +329,10 @@ async fn blob_handler(
     let path = path.strip_prefix('/').ok_or(Error::NotFound)?;
     let storage = &ctx.profile.storage;
     let repo = Repository::open(paths::repository(storage, &project))?;
-    let blob = Blob::new(&repo, &sha, &path)?;
+    let blob = repo.blob(sha, &path)?;
+    let response = api::json::blob(&blob, path);
 
-    Ok::<_, Error>(Json(blob))
+    Ok::<_, Error>(Json(response))
 }
 
 /// Get project readme.
@@ -336,8 +353,9 @@ async fn readme_handler(
     ];
 
     for path in paths {
-        if let Ok(blob) = Blob::new(&repo, &sha, &path) {
-            return Ok::<_, Error>(Json(blob));
+        if let Ok(blob) = repo.blob(sha, path) {
+            let response = api::json::blob(&blob, path);
+            return Ok::<_, Error>(Json(response));
         }
     }
 
@@ -437,35 +455,6 @@ impl<'a> FromIterator<(&'a CommentId, &'a thread::Comment)> for Comments {
 
         Comments(comments)
     }
-}
-
-#[derive(Serialize)]
-struct Stats {
-    branches: usize,
-    commits: usize,
-    contributors: usize,
-}
-
-fn stats(repo: &Repository, head: Oid) -> Result<Stats, Error> {
-    let branches = repo.branches(Glob::all_heads())?.count();
-
-    let mut commits = 0;
-    let contributors = History::new(repo, head)?
-        .filter_map(|commit| {
-            commits += 1;
-            if let Ok(commit) = commit {
-                Some((commit.author.name, commit.author.email))
-            } else {
-                None
-            }
-        })
-        .collect::<HashSet<_>>();
-
-    Ok(Stats {
-        branches,
-        commits,
-        contributors: contributors.len(),
-    })
 }
 
 #[cfg(test)]
@@ -723,16 +712,16 @@ mod routes {
             json!({
                 "entries": [
                   {
-                    "path": "README",
-                    "name": "README",
-                    "lastCommit": null,
-                    "kind": "blob"
-                  },
-                  {
                     "path": "dir1",
                     "name": "dir1",
                     "lastCommit": null,
                     "kind": "tree"
+                  },
+                  {
+                    "path": "README",
+                    "name": "README",
+                    "lastCommit": null,
+                    "kind": "blob"
                   }
                 ],
                 "lastCommit": {
@@ -778,7 +767,20 @@ mod routes {
                   "kind": "blob"
                 }
               ],
-              "lastCommit": null,
+              "lastCommit": {
+                "sha1": HEAD,
+                "author": {
+                  "name": "Alice Liddell",
+                  "email": "alice@radicle.xyz"
+                },
+                "summary": "Add another folder",
+                "description": "",
+                "committer": {
+                  "name": "Alice Liddell",
+                  "email": "alice@radicle.xyz"
+                },
+                "committerTime": 1673001014
+              },
               "name": "dir1",
               "path": "dir1",
               "stats": {
@@ -837,6 +839,42 @@ mod routes {
         let response = request(
             &app,
             format!("/projects/rad:z4FucBZHZMCsxTyQE1dfE2YR59Qbp/blob/{HEAD}/README"),
+        )
+        .await;
+
+        assert_eq!(response.status(), StatusCode::OK);
+        assert_eq!(
+            response.json().await,
+            json!({
+                "binary": false,
+                "content": "Hello World!\n",
+                "lastCommit": {
+                    "sha1": HEAD_1,
+                    "author": {
+                        "name": "Alice Liddell",
+                        "email": "alice@radicle.xyz"
+                    },
+                    "summary": "Initial commit",
+                    "description": "",
+                    "committer": {
+                        "name": "Alice Liddell",
+                        "email": "alice@radicle.xyz"
+                    },
+                    "committerTime": 1673001014
+                },
+                "name": "README",
+                "path": "README"
+            })
+        );
+    }
+
+    #[tokio::test]
+    async fn test_projects_readme() {
+        let tmp = tempfile::tempdir().unwrap();
+        let app = super::router(test::seed(tmp.path()));
+        let response = request(
+            &app,
+            format!("/projects/rad:z4FucBZHZMCsxTyQE1dfE2YR59Qbp/readme/{HEAD}"),
         )
         .await;
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1067,7 +1067,7 @@ dependencies = [
 [[package]]
 name = "git-ref-format"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-git?rev=79a94721366490053e2d8ac1c1afa14fb0c25f09#79a94721366490053e2d8ac1c1afa14fb0c25f09"
+source = "git+https://github.com/radicle-dev/radicle-git?rev=016ece66474edf45d2d304141515152e6f7b37a3#016ece66474edf45d2d304141515152e6f7b37a3"
 dependencies = [
  "git-ref-format-core",
  "git-ref-format-macro",
@@ -1076,7 +1076,7 @@ dependencies = [
 [[package]]
 name = "git-ref-format-core"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-git?rev=79a94721366490053e2d8ac1c1afa14fb0c25f09#79a94721366490053e2d8ac1c1afa14fb0c25f09"
+source = "git+https://github.com/radicle-dev/radicle-git?rev=016ece66474edf45d2d304141515152e6f7b37a3#016ece66474edf45d2d304141515152e6f7b37a3"
 dependencies = [
  "serde",
  "thiserror",
@@ -1085,7 +1085,7 @@ dependencies = [
 [[package]]
 name = "git-ref-format-macro"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-git?rev=79a94721366490053e2d8ac1c1afa14fb0c25f09#79a94721366490053e2d8ac1c1afa14fb0c25f09"
+source = "git+https://github.com/radicle-dev/radicle-git?rev=016ece66474edf45d2d304141515152e6f7b37a3#016ece66474edf45d2d304141515152e6f7b37a3"
 dependencies = [
  "git-ref-format-core",
  "proc-macro-error",
@@ -2229,7 +2229,7 @@ dependencies = [
 [[package]]
 name = "radicle-git-ext"
 version = "0.2.0"
-source = "git+https://github.com/radicle-dev/radicle-git?rev=79a94721366490053e2d8ac1c1afa14fb0c25f09#79a94721366490053e2d8ac1c1afa14fb0c25f09"
+source = "git+https://github.com/radicle-dev/radicle-git?rev=016ece66474edf45d2d304141515152e6f7b37a3#016ece66474edf45d2d304141515152e6f7b37a3"
 dependencies = [
  "git-ref-format",
  "git2",
@@ -2325,12 +2325,12 @@ dependencies = [
 [[package]]
 name = "radicle-std-ext"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-git?rev=79a94721366490053e2d8ac1c1afa14fb0c25f09#79a94721366490053e2d8ac1c1afa14fb0c25f09"
+source = "git+https://github.com/radicle-dev/radicle-git?rev=016ece66474edf45d2d304141515152e6f7b37a3#016ece66474edf45d2d304141515152e6f7b37a3"
 
 [[package]]
 name = "radicle-surf"
 version = "0.8.0"
-source = "git+https://github.com/radicle-dev/radicle-git?rev=79a94721366490053e2d8ac1c1afa14fb0c25f09#79a94721366490053e2d8ac1c1afa14fb0c25f09"
+source = "git+https://github.com/radicle-dev/radicle-git?rev=016ece66474edf45d2d304141515152e6f7b37a3#016ece66474edf45d2d304141515152e6f7b37a3"
 dependencies = [
  "anyhow",
  "base64",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,8 @@ version = "0.1.0"
 
 [patch.crates-io.radicle-git-ext]
 git = "https://github.com/radicle-dev/radicle-git"
-rev = "79a94721366490053e2d8ac1c1afa14fb0c25f09"
+rev = "016ece66474edf45d2d304141515152e6f7b37a3"
 
 [patch.crates-io.git-ref-format]
 git = "https://github.com/radicle-dev/radicle-git"
-rev = "79a94721366490053e2d8ac1c1afa14fb0c25f09"
+rev = "016ece66474edf45d2d304141515152e6f7b37a3"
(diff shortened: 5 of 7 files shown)


radicle-httpd/src/api/v1/projects.rs
```diff
@@ -131,7 +131,7 @@ async fn history_handler(
     };
 
     let storage = &ctx.profile.storage;
-    let repo = storage.repository(project)?;
+    let repo = Repository::open(storage.path())?;
```
keepsimple1, line 134: It is confusing that we have two `Repository` types, one from `storage` and one from` radicle-surf`. Hopefully we can merge them somehow.

cloudhead: `storage.path()` does not point to a repository though. I guess you'd have to do:
`let repo = Repository::from(repo.raw())` if you had a `From<git2::Repository>` instance.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -349,7 +348,7 @@ async fn issues_handler(
 /// `GET /projects/:project/issues/:id`
 async fn issue_handler(
     Extension(ctx): Extension<Context>,
-    Path((project, issue_id)): Path<(Id, Oid)>,
+    Path((project, issue_id)): Path<(Id, radicle::git::Oid)>,
```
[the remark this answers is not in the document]
cloudhead: Why can't they be substituted?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -224,16 +227,15 @@ async fn activity_handler(
...
     Extension(ctx): Extension<Context>,
-    Path((project, sha, path)): Path<(Id, Oid, String)>,
+    Path((_, sha, path)): Path<(Id, Oid, String)>,
 ) -> impl IntoResponse {
     let path = path.strip_prefix('/').ok_or(Error::NotFound)?.to_string();
     let storage = &ctx.profile.storage;
-    let repo = storage.repository(project)?;
-    let tree = radicle_surf::object::tree(&repo.raw().into(), Some(Sha { sha }), Some(path))?;
+    let repo = Repository::open(storage.path())?;
+    let sub_dir = repo.directory(sha, &path)?;
```
cloudhead, line 235: I find the use of `directory` here really confusing, (and `file` in other places), since these are not directories on the file system, but trees inside the git object database..

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -375,11 +374,12 @@ struct Stats {
     contributors: usize,
 }
 
-fn stats<R: WriteRepository>(repo: &R) -> Result<Stats, Error> {
-    let branches = repo.raw().branches(Some(BranchType::Local))?.count();
-    let (_, head) = repo.head()?;
+fn stats(repo: &Repository) -> Result<Stats, Error> {
+    let branches = repo.branches(Glob::all_heads())?.count();
```
cloudhead, line 378: Is this equivalent? ie. it won't pick up remotes?

radicle-httpd/src/api/v1/projects.rs
keepsimple1: I see. `storage` can have multiple resources, where each resource is a git repo, and `storage.path()` is the parent directory of the actual git repo. Is that the case? I will check what's the proper way to open the repo here.

radicle-httpd/src/api/v1/projects.rs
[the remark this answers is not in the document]
keepsimple1: For some reason, rustc / cargo didn't like it.  For example, if I removed `readicle::git::` here, I would get the following error:

```
error[E0277]: the trait bound `ObjectId: From<radicle_surf::git::Oid>` is not satisfied
   --> radicle-httpd/src/api/v1/projects.rs:356:15
    |
356 |         .get(&issue_id.into())?
    |               ^^^^^^^^ ---- required by a bound introduced by this call
    |               |
    |               the trait `From<radicle_surf::git::Oid>` is not implemented for `ObjectId`
    |
    = help: the following other types implement trait `From<T>`:
              <ObjectId as From<&radicle::git::Oid>>
              <ObjectId as From<&radicle::git::git2::Oid>>
              <ObjectId as From<radicle::git::Oid>>
              <ObjectId as From<radicle::git::git2::Oid>>
    = note: required for `radicle_surf::git::Oid` to implement `Into<ObjectId>`
```

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -224,16 +227,15 @@ async fn activity_handler(
...
     let path = path.strip_prefix('/').ok_or(Error::NotFound)?.to_string();
     let storage = &ctx.profile.storage;
-    let repo = storage.repository(project)?;
-    let tree = radicle_surf::object::tree(&repo.raw().into(), Some(Sha { sha }), Some(path))?;
+    let repo = Repository::open(storage.path())?;
+    let sub_dir = repo.directory(sha, &path)?;
     let response = json!({
-        "path": &tree.path,
-        "entries": &tree.entries,
-        "info": &tree.info,
```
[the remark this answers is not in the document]
keepsimple1: My first check was it seems not really needed. I will list the actual content, and if needed, we can add it.

radicle-httpd/src/api/v1/projects.rs
keepsimple1: correct. It won't pick up remotes.

radicle-httpd/src/api/v1/projects.rs
cloudhead: correct!

radicle-httpd/src/api/v1/projects.rs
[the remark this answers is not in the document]
cloudhead: There may be two versions imported somehow, I'll take a look.

radicle-httpd/src/api/v1/projects.rs
cloudhead: I think there's some cognitive dissonance there; eg. if it's a directory, I'd expect that I can interact with it like a directory, eg. move it or place files inside of it, or delete it.

radicle-httpd/src/api/v1/projects.rs
FintanH: We have a function for this https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/src/git/repo.rs#L272-L288

radicle-httpd/src/api/v1/projects.rs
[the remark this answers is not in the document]
FintanH: I removed the `Info` type and flattened it. The serialization for `Tree` is now [this](https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/src/source/object/tree.rs#L87-L101). So `Directory` isn't good enough in this case since it's missing information that would usually be displayed.

radicle-httpd/src/api/v1/projects.rs
FintanH: Ya, it's probably best to do the `From<git2::Repository>`. The other option which I mentioned waaaaay back was to design surf using traits :) But that would be a lot more work now.

keepsimple1: I've updated the diff to grab the correct (hopefully) repo path based on the `storage` and the `project`. 

I'm trying to stick with getting the path string because: "Write programs to handle text streams, because that is a universal interface." (UNIX).  ;-)

radicle-httpd/src/api/v1/projects.rs
keepsimple1: Good to know that! This is a read-only directory.  We can document that or name it better. However now I'm a bit concerned that maybe we should get rid of `Directory` and `File`, and only use `Tree` and `Blob`? 

For some reason, I automatically felt comfortable with the read-only nature of `Directory` and `File` provided by `radicle-surf` API. Didn't realize others would feel so differently.

radicle-httpd/src/api/v1/projects.rs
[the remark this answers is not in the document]
keepsimple1: Looks like it was because the root `Cargo.toml` has a patch section for `radicle-git-ext`. I've updated the diff to use the same version for the patch. Now the problem is gone.

radicle-httpd/src/api/v1/projects.rs
keepsimple1: Btw, what is our strategy for writing test for `radicle-httpd`?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -131,7 +130,7 @@ async fn history_handler(
     };
 
     let storage = &ctx.profile.storage;
-    let repo = storage.repository(project)?;
+    let repo = Repository::open(storage_repo_path(storage, &project))?;
```
cloudhead, line 133: We try to avoid aliasing as it creates an extra step for the reader to figure out what is being called; I think if you import up to `git::paths`, it should be clear enough: `Repository::open(paths::repository(storage, &project))`

cloudhead: I would also almost prefer to move this logic inside context, since it's so common, eg. we could have a function `ctx.surf(&project)`, which would do exactly this.

radicle-httpd/src/api/v1/projects.rs
cloudhead: Well, it's still not quite a read-only directory, since it doesn't exist on the file system; ie. I can't `cd` into it, and I can't access it via `std::fs` or check its file-system path. It's some kind of "virtual" directory called a "tree", so I'd just name it what it is to avoid confusing it with a regular directory.. Especially that Git does also work with regular directories, in working copies, and surf may want to provide an API for that at some point.

radicle-httpd/src/api/v1/projects.rs
cloudhead: We have tests that we need to port from client-services, see here for eg.: https://github.com/radicle-dev/radicle-client-services/blob/master/http-api/src/v1/projects.rs#L977

radicle-httpd/src/api/v1/projects.rs
keepsimple1: Sounds good, I will remove the aliasing. 

> I would also almost prefer to move this logic inside context, since it's so common, eg. we could have a function ctx.surf(&project), which would do exactly this.

This seems to be an optimization that can be done separately.

radicle-httpd/src/api/v1/projects.rs
keepsimple1: I think the reason I liked the name `directory` (and `file`) is that I felt they naturally fit with the motivation/ideas of `radicle-surf`:

"radicle-surf is a library to describe a Git repository as a file system. It aims to provide an easy-to-use API to browse a repository via the concept of files and directories for any given revision." ([README](https://github.com/radicle-dev/radicle-git/tree/main/radicle-surf)).

It is surely possible to only use `Tree` (and `Blob`) in `radicle-surf` and let surf's users to convert them into `Directory` and `File` when presenting to the end user. But I thought it's a good idea for `radicle-surf` to present `Directory` and `File` in its API already.

I'm a bit surprised to see you mention ` I can't cd into it, and I can't access it via std::fs or check its file-system path. `.  as that's not what I expected of `Directory` from such a library. I was wondering if other users think the same way. I will try to ask around more.

Regarding the possible future of supporting working copies, I think it would need some additional design and it's difficult for me to predict anything: will it support write? will it be part of surf? My take is that I don't think having `Directory` will prevent developing that feature.

cloudhead: > "radicle-surf is a library to describe a Git repository as a file system. It aims to provide an easy-to-use API to browse a repository via the concept of files and directories for any given revision." ([README](https://github.com/radicle-dev/radicle-git/tree/main/radicle-surf)).

That was written a long time ago, when surf was multi-vcs, which means that concepts like Blob or Tree would not work, since not all VCSs have those.

> I'm a bit surprised to see you mention I can't cd into it, and I can't access it via std::fs or check its file-system path.. as that's not what I expected of Directory from such a library. I was wondering if other users think the same way. I will try to ask around more.

To flip it around: I *wouldn't* expect that an object called "Directory" would have a SHA-1; yet in this case it does. However you put it, the expectations are broken due to the mismatch between Directories and Trees. When working with bare repositories, I would search the docs for "blob" and "tree", given that's the object I want to work with and be very surprised if I don't get any results in a git-oriented library. If I saw "directory" in the docs, I would expect that to be used for actual file-system objects, eg. repository directories, user config directories or the working copy directories.

cloudhead: I find it really strange that we would shy away from naming something what it actually is when there is a specific and well-known term used for that thing. That alone I think would create confusion: "why did they call it File? Is it different in some way from a Blob? Does it only work for blobs in working copies?" The only good reason to pick a different name in this case is if it is *not* intended to behave like a Blob/Tree.

keepsimple1: I think it's a good discussion that helps me think more about the names. (As they say, naming is hard ;-)) . I spent a bit time to read the git [book](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects). My take is that, in git world:

1. `Tree` and `Blob` are Git internals, even though they are very familiar terms for developers of `radicle-git` / `hearwood`.

2. `Tree` and `Blob` are contents.  `Tree` is the content of a virtual directory, and `Blob` is the content of a file. This means that if two directories have the same content, they share the same tree even when they are different directories.  Likewise, two files can share the same blob.

For example in a simple test: 
```
$ git cat-file -p branch_b^{tree}
040000 tree 494c44e7fe186a29d2f360a6d5a4071a47a8041c    dir_1
040000 tree 494c44e7fe186a29d2f360a6d5a4071a47a8041c    dir_2
100644 blob a5bce3fd2565d8f458555a0c6f42d0504a848bd5    test1
100644 blob f83bb2683ace630ef6e8dcfde2aedde6948c90b4    test2
100644 blob f83bb2683ace630ef6e8dcfde2aedde6948c90b4    test3
```

This seems to me, logically a `Directory` is a container that has a `Tree` (its content) and other info (mainly its path). `File` is a container that has a `Blob` (its content) and other info (file name). 

And that's how I understand the difference between (`Directory`, `File`) and (`Tree`, `Blob`). 

3.  If we expose `Tree` and `Blob` in radicle-surf's public API, I hope they map to what they mean in git. Currently they are not, for example  [`Blob`](https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/src/source/object/blob.rs#L36) contains a `File`. I will play around this area in radicle-surf to see if I can do anything.

cloudhead: > This seems to me, logically a Directory is a container that has a Tree (its content) and other info (mainly its path). File is a container that has a Blob (its content) and other info (file name).

This is true in a sense but also confuses things: git doesn't have directories or files in its object database, and the file system doesn't have trees/blobs. So saying that a directory "contains" a tree is not right, unless you mean it in a very abstract sense. I'd say more that a tree is *associated* with a directory.

FintanH: > I'd say more that a tree is associated with a directory.

Or a tree can materialize a directory, i.e. when you checkout a commit the associated tree is rendered into a file system directory.

FintanH: Another note that I'll repeat here is that the httpd currently requires additional information for these types:
* What was the last commit for a Blob or Tree
* The kind of the Blob content, i.e. plain text or binary
* The entries for the current Tree

These are all computable so we either provide utility functions for calculating those or we provide the data itself.

keepsimple1: I have updated https://github.com/radicle-dev/radicle-git/pull/79 to try out using `Tree` (and later will finish `Blob` part). Will update here after that.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -226,17 +224,16 @@ async fn activity_handler(
...
-    let repo = storage.repository(project)?;
-    let tree = radicle_surf::object::tree(&repo.raw().into(), Some(Sha { sha }), Some(path))?;
+    let repo = Repository::open(paths::repository(storage, &proj))?;
+    let tree = repo.tree(sha, &path)?;
     let response = json!({
-        "path": &tree.path,
-        "entries": &tree.entries,
-        "info": &tree.info,
-        "stats": stats(&repo)?,
+        "path": &path,
```
keepsimple1, line 234: If I understand correctly, we can use the input `path` here.

radicle-httpd/src/api/v1/projects.rs
keepsimple1: I have updated to use `Tree` and `Blob` here.  They are content based, same as in git.  `Tree` itself does not contain a path, and `Blob` itself does not contain a name.  For `path` and `name`, the caller should already have them.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -275,24 +272,28 @@ async fn remote_handler(
...
     let path = path.strip_prefix('/').ok_or(Error::NotFound)?;
     let storage = &ctx.profile.storage;
-    let repo = storage.repository(project)?;
-    let blob = radicle_surf::blob::blob(&repo.raw().into(), Some(Sha { sha }), path)?;
+    let repo = Repository::open(paths::repository(storage, &proj))?;
+    let blob = repo.blob(sha, &path)?;
+    let response = json!({
+        "path": &path,
+        "content": blob.content(),
+    });
```
keepsimple1, line 284: The actual JSON response might be different, depending on what `httpd` team wanted. This is just to show that I think `httpd` should define its JSON output format independent of `radicle-surf` API serialization. 

If there is some data needed by `httpd` JSON but missing from `radicle-surf` API, we can always fix that.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -303,14 +304,16 @@ async fn readme_handler(
     ];
 
     for path in paths {
-        if let Ok(blob) = radicle_surf::blob::blob(&repo.raw().into(), Some(Sha { sha }), path) {
-            return Ok::<_, Error>(Json(blob));
+        if let Ok(blob) = repo.blob(sha, path) {
+            let response = json!({
+                "path": &path,
+                "readme": blob.content(),
```
keepsimple1, line 310: Again this is to show an example / possibility, not saying the field should be named "readme".

radicle-httpd/src/api/v1/projects.rs
cloudhead: Yeah I mostly agree with that, though I think that a good JSON format can work for more than one use-case, so it's a good aim to have.

radicle-httpd/src/api/v1/projects.rs
cloudhead: Yeah we don't send over binary data, and when it's plain text we convert to utf8 string and send that over.

keepsimple1: What if a file is big? say 10GB, or 1GB? would we do streaming instead?

radicle-httpd/src/api/v1/projects.rs
keepsimple1: btw, is there a spec for `radicle-httpd` API? (like in swagger format or something similar?)

radicle-httpd/src/api/v1/projects.rs
cloudhead: Hmm, not sure. The UI is not meant to handle files like that really, so I'd say that beyond a certain size we wouldn't display it.

FintanH: Git in general isn't very good with large files, so I don't _think_ we have to try worry about it too much

radicle-httpd/src/api/v1/projects.rs
cloudhead: No, but this is something we intended to create as things stabilize a little!

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -226,17 +224,37 @@ async fn activity_handler(
...
+    let entries = tree
+        .entries()
+        .iter()
+        .map(|entry| EntryInfo {
+            name: entry.name(),
+            path: prefix.join(entry.name()),
+            kind: if entry.is_tree() { "tree" } else { "blob" },
+            last_commit: entry.commit(),
+        })
+        .collect::<Vec<EntryInfo>>();
```
keepsimple1, line 252: This is how I think `radicle-httpd` can use `radicle-surf` API without relying on surf's serialization.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -226,17 +224,37 @@ async fn activity_handler(
...
+    }
+
+    let prefix = std::path::Path::new(&path);
+    let entries = tree
+        .entries()
+        .iter()
+        .map(|entry| EntryInfo {
+            name: entry.name(),
+            path: prefix.join(entry.name()),
+            kind: if entry.is_tree() { "tree" } else { "blob" },
```
FintanH, line 249: Might be nicer if surf provides a `fn kind` that has a `Display` and/or `Serialize`

FintanH (commented):
Looking good from my perspective :)

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -236,10 +253,30 @@ async fn tree_handler(
...
+
+    let prefix = std::path::Path::new(&path);
+    let entries = tree
+        .entries()
+        .iter()
+        .map(|entry| {
+            json!( {
+                "path": prefix.join(entry.name()),
+                "name": entry.name(),
+                "lastCommit": serde_json::Value::Null,
```
keepsimple1, line 266: This is to match the current unit test. We can replace it with the actual value and update the unit test.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -722,17 +774,17 @@ mod routes {
...
                     "path": "dir1",
                     "name": "dir1",
                     "lastCommit": null,
                     "kind": "tree"
+                  },
+                  {
+                    "path": "README",
+                    "name": "README",
+                    "lastCommit": null,
+                    "kind": "blob"
```
keepsimple1, line 787: Now `repo.tree()` returns a tree with entries sorted.  The sort always lists the directories (trees) first.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -778,7 +830,20 @@ mod routes {
                   "kind": "blob"
                 }
               ],
-              "lastCommit": null,
+              "lastCommit": {
```
keepsimple1, line 833: We should have this info in the output, not sure why it didn't.  This is also an example of the actual lastCommit value to replace a null.

keepsimple1:
I have rebased the diff to the master branch, and updated the `radicle-git` ref to the latest (1/12/2023).  I've also updated the JSON creation for the `api/v1/projects.rs` avoid using the serialization of `radicle-surf` types, except `Diff`.

I think going forward, `radicle-surf` can provide default serializations for its public types like `Diff`.  A user can choose to use the default serializations or create their own JSON directly like in this PR.

With this update, I am aiming to land the changes. Please review.  Thanks!

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -166,14 +161,23 @@ async fn history_handler(
...
+                let branches: Vec<String> = repo
+                    .revision_branches(c.id, glob)?
+                    .iter()
+                    .map(|b| b.refname().to_string())
+                    .collect();
+                let diff = repo.diff_commit(c.id)?;
+                Ok(json!({
+                "header": commit_json(&c),
+                "diff": diff,
+                "branches": branches
```
FintanH, line 176:
```suggestion
                    "header": commit_json(&c),
                    "diff": diff,
                    "branches": branches
```

radicle-httpd/src/api/v1/projects.rs
FintanH: Is this comment referring to https://github.com/radicle-dev/heartwood/pull/149/files#diff-85a73839d407aa508b9e78b3756b0e9b6e8d1087a0cafac6a8325a8cf427b65eR266?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -401,6 +450,38 @@ async fn issue_handler(
...
+    match path.rfind('/') {
+        Some(idx) => {
+            if idx == path.len() - 1 {
+                ""
+            } else {
+                &path[idx + 1..]
+            }
+        }
+        None => path,
+    }
```
FintanH, line 464: I think this can be written as:
```suggestion
    match path.rsplit('/').next() {
        Some(name) => name,
        None => path,
    }
```

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -401,6 +450,38 @@ async fn issue_handler(
...
+      },
+      "summary": last_commit.summary,
+      "description": last_commit.description(),
+      "committer": {
+        "name": last_commit.committer.name,
+        "email": last_commit.committer.email
+      },
+      "committerTime": last_commit.committer.time.seconds()
+    })
+}
```
FintanH, line 483:
```suggestion
fn commit_json(commit: &Commit) -> serde_json::Value {
    json!({
      "sha1": commit.id,
      "author": {
        "name": commit.author.name,
        "email": commit.author.email
      },
      "summary": commit.summary,
      "description": commit.description(),
      "committer": {
        "name": commit.committer.name,
        "email": commit.committer.email
      },
      "committerTime": commit.committer.time.seconds()
    })
}
```

FintanH (commented):
Looking good

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -236,10 +253,30 @@ async fn tree_handler(
...
+        })
+        .collect::<Vec<_>>();
+
+    let last_commit = tree.commit();
+    let response = json!({
+        "entries": &entries,
+        "lastCommit": commit_json(last_commit),
+        "name": name_in_path(&path),
+        "path": &path,
+        "stats": repo.stats()?,
```
cloudhead, line 278: This looks incorrect; what commit will it use for the stats?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -236,10 +253,30 @@ async fn tree_handler(
...
-    let mut response = json!(tree);
-    response["stats"] = json!(stats(&repo, sha)?);
+    let tree = repo.tree(sha, &path)?;
+
+    let prefix = std::path::Path::new(&path);
+    let entries = tree
+        .entries()
+        .iter()
+        .map(|entry| {
+            json!( {
```
cloudhead, line 263:
```suggestion
            json!({
```

radicle-httpd/src/api/v1/projects.rs
FintanH: Ohhhh, was there a conversation about this before... I think you want to use `stats_from` here https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/src/repo.rs#L334-L353

cloudhead: Yeah there was :)

radicle-httpd/src/api/v1/projects.rs
xphoniex: don't use plain hash, use `HEAD` like elsewhere

radicle-httpd/src/api/error.rs
```diff
@@ -38,17 +38,13 @@ pub enum Error {
...
 
-    /// Surf object error.
+    /// Surf error.
     #[error(transparent)]
-    SurfObject(#[from] radicle_surf::source::object::Error),
-
-    /// Surf git error.
-    #[error(transparent)]
-    SurfGit(#[from] radicle_surf::git::Error),
+    SurfGit(#[from] radicle_surf::Error),
```
xphoniex, line 47: change `SurfGit` to `Surf`

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -236,10 +253,30 @@ async fn tree_handler(
     let path = path.strip_prefix('/').ok_or(Error::NotFound)?.to_string();
     let storage = &ctx.profile.storage;
     let repo = Repository::open(paths::repository(storage, &project))?;
-    let path = if path.is_empty() { None } else { Some(&path) };
-    let tree = Tree::new(&repo, &sha, path)?;
-    let mut response = json!(tree);
-    response["stats"] = json!(stats(&repo, sha)?);
+    let tree = repo.tree(sha, &path)?;
+
```
xphoniex, line 257: remove empty line 257, and use `Path::new` instead of `std::path::Path::new`

radicle-httpd/src/api/v1/projects.rs
xphoniex: is `fn commit_json` temporary? this doesn't belong here.

xphoniex (commented):
Nice work @keepsimple1 , can you make sure this passes e2e tests from [`radicle-interface`](https://github.com/radicle-dev/radicle-interface/) as well, before merging?

Terminal 1 (must have `radicle-httpd` in PATH)

```
./scripts/run-httpd-with-fixtures -b --non-interactive
```

Terminal 2
```
HEARTWOOD=true npm run test:e2e --  --project chromium
```

radicle-httpd/src/api/v1/projects.rs
keepsimple1: Got it, will do.

radicle-httpd/src/api/v1/projects.rs
keepsimple1: I missed that, thanks!

radicle-httpd/src/api/v1/projects.rs
xphoniex: why? what will the default serialization be used for then? cc @FintanH

keepsimple1:
Addressed all review comments except passing e2e tests of `radicle-interface`. Working on the e2e test.

radicle-httpd/src/api/v1/projects.rs
keepsimple1: The rationale is to keep `radicle-surf` and its users decoupled (or, loosely coupled).  It is up to the user to use the default serializations or not. For example, in this case, the current JSON output in `radicle-httpd` uses `sha1` as the key name for commit, while the default serialization would have `id` as the key name.

This diff is to show how the user can decide its own JSON output. Otherwise, whenever `radicle-surf` changes its serialization for its `commit` type, it will immediately break the users JSON output.

xphoniex: > The rationale is to keep `radicle-surf` and its users decoupled

Ok, but why?

> Otherwise, whenever `radicle-surf` changes its serialization for its `commit` type, it will immediately break the users JSON output.

- What other applications are consuming `surf`?
- And, even if that is the reason, you changing the scheme will break the custom serializer living in `httpd`, no?

keepsimple1: > Ok, but why?

I've explained in more details in https://github.com/radicle-dev/heartwood/issues/172 . 

> * What other applications are consuming `surf`?

I don't know at this point, but I hope there will be more than `radicle-httpd` in the future ;-) .
 
> * And, even if that is the reason, you changing the scheme will break the custom serializer living in `httpd`, no?

What do you mean by `changing the scheme` ?

keepsimple1:
Thanks a lot for your help @sebastinez @xphoniex !  Now this PR passed the e2e test of `radicle-interface`.  (It failed earlier because of `readme_handler` changes. I fixed it now, and added a new unit test to cover the readme.)

radicle-httpd/src/api/v1/projects.rs
FintanH: > > The rationale is to keep `radicle-surf` and its users decoupled
> 
> Ok, but why?

I think this is a good move, to be honest. The `httpd` code is what wants to serve up JSON in the first place, so having it define its own schemas for that data is important. Surf is then just the tool for grabbing the required data, which it should be consistent in doing.

> And, even if that is the reason, you changing the scheme will break the custom serializer living in httpd, no?

Going forward, I don't think `httpd` should rely on any schema surf has defined.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -401,6 +440,43 @@ async fn issue_handler(
...
+/// Returns JSON for a blob with a given `path`.
+fn blob_json(blob: &Blob, path: &str) -> serde_json::Value {
+    json!({
+        "binary": blob.is_binary(),
+        "content": blob.content(),
+        "name": name_in_path(path),
+        "path": path,
+        "lastCommit": commit_json(blob.commit())
+    })
+}
```
cloudhead, line 478: Can you move these two functions to `json::commit` and `json::blob` by creating a `json` module in here?

keepsimple1: Will do.

xphoniex: put it into its own file `/api/json.rs`

keepsimple1: Done.

cloudhead:
LGTM.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -166,14 +161,23 @@ async fn history_handler(
...
+                    .revision_branches(c.id, glob)?
+                    .iter()
+                    .map(|b| b.refname().to_string())
+                    .collect();
+                let diff = repo.diff_commit(c.id)?;
+                Ok(json!({
+                    "header": api::json::commit(&c),
+                    "diff": diff,
+                    "branches": branches
+                }))
```
FintanH, line 177: Not a blocker, but maybe all `json!` calls should go into the `json` module now

FintanH (approved):
Nice work :ok_hand:

radicle-httpd/src/api/json.rs
```diff
@@ -0,0 +1,68 @@
+//! Utilities for building JSON responses of our API.
```
xphoniex, line 1: we don't use comments like this, might wanna remove it.

cloudhead: We definitely welcome module documentation though :)

radicle-httpd/src/api/v1/projects.rs
cloudhead: aye, this can be cleaned up later.

cloudhead (approved):

radicle-httpd/src/api/json.rs
xphoniex: Yup, another case of keeping everything uniform.

xphoniex (approved):

PR #155, opened 2022-12-08
Manage project delegates iff threshold = 1

FintanH:
This series allows the adding, removing, and listing of delegates.

It's limited to projects that have a threshold of 1 since we need to be able to gather signatures for an identity that needs a higher threshold.

The update of an identity with a higher threshold will be tackled in a separate series.

### DIFF
--- a/radicle-cli/src/commands/delegate.rs
+++ b/radicle-cli/src/commands/delegate.rs
@@ -1,0 +1,141 @@
+use std::ffi::OsString;
+use std::str::FromStr;
+
+use anyhow::{anyhow, Context as _};
+
+use radicle::identity::Id;
+use radicle_crypto::PublicKey;
+
+use crate::terminal as term;
+use crate::terminal::args::{Args, Error, Help};
+
+#[path = "delegate/add.rs"]
+mod add;
+#[path = "delegate/list.rs"]
+mod list;
+#[path = "delegate/remove.rs"]
+mod remove;
+
+pub const HELP: Help = Help {
+    name: "delegate",
+    description: "Manage the delegates of an identity",
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad delegate (add|remove) <public key> [--to <id>]
+    rad delegate list [<id>]
+
+    The `add` and `remove` commands are limited to managing delegates
+    where the `threshold` for the quorum is exactly `1`. Otherwise,
+    the verification of the document will not be able to gather enough
+    signatures to pass the quorum.
+
+Options
+
+    --help              Print help
+"#,
+};
+
+#[derive(Debug, Default, PartialEq, Eq)]
+pub enum OperationName {
+    Add,
+    Remove,
+    #[default]
+    List,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum Operation {
+    Add { id: Option<Id>, key: PublicKey },
+    Remove { id: Option<Id>, key: PublicKey },
+    List { id: Option<Id> },
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct Options {
+    pub op: Operation,
+}
+
+impl Args for Options {
+    fn from_args(args: Vec<OsString>) -> anyhow::Result<(Self, Vec<OsString>)> {
+        use lexopt::prelude::*;
+
+        let mut parser = lexopt::Parser::from_args(args);
+        let mut id: Option<Id> = None;
+        let mut op: Option<OperationName> = None;
+        let mut key: Option<PublicKey> = None;
+
+        while let Some(arg) = parser.next()? {
+            match arg {
+                Long("help") => {
+                    return Err(Error::Help.into());
+                }
+                Long("to") => {
+                    id = Some(parser.value()?.parse::<Id>()?);
+                }
+                Value(val) if op.is_none() => match val.to_string_lossy().as_ref() {
+                    "a" | "add" => op = Some(OperationName::Add),
+                    "r" | "remove" => op = Some(OperationName::Remove),
+                    "l" | "list" => op = Some(OperationName::List),
+
+                    unknown => anyhow::bail!("unknown operation '{}'", unknown),
+                },
+                Value(val) if op.is_some() => {
+                    let val = val.to_string_lossy();
+
+                    match op {
+                        Some(OperationName::Add) | Some(OperationName::Remove) => {
+                            if let Ok(val) = PublicKey::from_str(&val) {
+                                key = Some(val);
+                            } else {
+                                return Err(anyhow!("invalid Public Key '{}'", val));
+                            }
+                        }
+                        Some(OperationName::List) => {
+                            if let Ok(val) = Id::from_str(&val) {
+                                id = Some(val);
+                            } else {
+                                return Err(anyhow!("invalid Project ID '{}'", val));
+                            }
+                        }
+                        None => continue,
+                    }
+                }
+                _ => return Err(anyhow!(arg.unexpected())),
+            }
+        }
+
+        let op = match op.unwrap_or_default() {
+            OperationName::List => Operation::List { id },
+            OperationName::Add => Operation::Add {
+                id,
+                key: key.ok_or_else(|| anyhow!("a delegate key must be provided"))?,
+            },
+            OperationName::Remove => Operation::Remove {
+                id,
+                key: key.ok_or_else(|| anyhow!("a delegate key must be provided"))?,
+            },
+        };
+
+        Ok((Options { op }, vec![]))
+    }
+}
+
+pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
+    let profile = ctx.profile()?;
+    let storage = &profile.storage;
+
+    match options.op {
+        Operation::Add { id, key } => add::run(&profile, storage, get_id(id)?, key)?,
+        Operation::Remove { id, key } => remove::run(&profile, storage, get_id(id)?, &key)?,
+        Operation::List { id } => list::run(&profile, storage, get_id(id)?)?,
+    }
+
+    Ok(())
+}
+
+fn get_id(id: Option<Id>) -> anyhow::Result<Id> {
+    id.or_else(|| radicle::rad::cwd().ok().map(|(_, id)| id))
+        .context("Couldn't get ID from either command line or cwd")
+}
--- a/radicle-cli/src/commands/delegate/add.rs
+++ b/radicle-cli/src/commands/delegate/add.rs
@@ -1,0 +1,51 @@
+use anyhow::Context as _;
+use radicle::{
+    prelude::{Did, Id},
+    storage::{WriteRepository as _, WriteStorage},
+    Profile,
+};
+use radicle_crypto::PublicKey;
+
+use crate::terminal as term;
+
+pub fn run<S>(profile: &Profile, storage: &S, id: Id, key: PublicKey) -> anyhow::Result<()>
+where
+    S: WriteStorage,
+{
+    let signer = term::signer(profile)?;
+    let me = signer.public_key();
+
+    let mut project = storage
+        .get(&profile.public_key, id)?
+        .context("No project with such ID exists")?;
+
+    let repo = storage.repository(id)?;
+
+    if !project.is_delegate(me) {
+        return Err(anyhow::anyhow!(
+            "'{}' is not a delegate of the project, only a delegate may add this key",
+            me
+        ));
+    }
+
+    if project.threshold > 1 {
+        return Err(anyhow::anyhow!("project threshold > 1"));
+    }
+
+    if project.delegate(&key) {
+        project.sign(&signer).and_then(|(_, sig)| {
+            project.update(
+                signer.public_key(),
+                "Updated payload",
+                &[(signer.public_key(), sig)],
+                repo.raw(),
+            )
+        })?;
+        term::info!("Added delegate '{}'", Did::from(key));
+        term::success!("Update successful!");
+        Ok(())
+    } else {
+        term::info!("the delegate for '{}' already exists", key);
+        Ok(())
+    }
+}
--- a/radicle-cli/src/commands/delegate/remove.rs
+++ b/radicle-cli/src/commands/delegate/remove.rs
@@ -1,0 +1,54 @@
+use anyhow::Context as _;
+use radicle::{
+    prelude::Id,
+    storage::{WriteRepository as _, WriteStorage},
+    Profile,
+};
+use radicle_crypto::PublicKey;
+
+use crate::terminal as term;
+
+pub fn run<S>(profile: &Profile, storage: &S, id: Id, key: &PublicKey) -> anyhow::Result<()>
+where
+    S: WriteStorage,
+{
+    let signer = term::signer(profile)?;
+    let me = signer.public_key();
+
+    let mut project = storage
+        .get(&profile.public_key, id)?
+        .context("No project with such ID exists")?;
+
+    let repo = storage.repository(id)?;
+
+    if !project.is_delegate(me) {
+        return Err(anyhow::anyhow!(
+            "'{}' is not a delegate of the project, only a delegate may remove this key",
+            me
+        ));
+    }
+
+    if project.threshold > 1 {
+        return Err(anyhow::anyhow!("project threshold > 1"));
+    }
+
+    match project.rescind(key)? {
+        Some(delegate) => {
+            project.sign(&signer).and_then(|(_, sig)| {
+                project.update(
+                    signer.public_key(),
+                    "Updated payload",
+                    &[(signer.public_key(), sig)],
+                    repo.raw(),
+                )
+            })?;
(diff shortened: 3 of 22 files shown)


radicle-cli/src/commands/delegate.rs
```diff
@@ -0,0 +1,122 @@
...
+
+pub const HELP: Help = Help {
+    name: "delegate",
+    description: "Manage the delegates of an identity",
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad delegate add [<id>] --key <public key>
+    rad delegate remove [<id>] --key <public key>
```
cloudhead, line 24: So I would flip these around since the `id` is implied if we're in a project dir, whereas the key is always required:

`rad delegate (add|remove) <public-key> [--repo <id>]`

radicle-cli/src/commands/delegate.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use radicle::identity::Id;
+use radicle_crypto::PublicKey;
+
+use crate::terminal as term;
+use crate::terminal::args::{Args, Error, Help};
+
+mod add;
+mod list;
+mod remove;
```
cloudhead, line 14: Probably these should be under a `delegate/` folder

radicle/src/identity/project.rs
```diff
@@ -117,6 +117,19 @@ impl Doc<Verified> {
         false
     }
 
+    pub fn remove(&mut self, key: &crypto::PublicKey) -> Result<Option<Did>, DocError> {
```
cloudhead, line 120: `undelegate`?

radicle-cli/src/commands/add.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    let mut project = storage
+        .get(&profile.public_key, id)?
+        .context("No project with such ID exists")?;
+
+    let repo = storage.repository(id)?;
+
+    if !project.is_delegate(me) && key != *me {
+        return Err(anyhow::anyhow!("'{}' is not a current delegate of the project and does not have permission to add another person's key. However, they may add their own.", me));
+    }
```
cloudhead, line 22: We probably don't want this right? This lets anyone add their own key, even if they aren't a delegate :)

cloudhead:
Looking good. By the way since we have a separate `threshold` property, we can actually create valid updates with any number of delegates, as long as `threshold = 1`.

radicle-cli/src/commands/delegate.rs
FintanH: Ya, that's fair. Originally I also had the `--name` so it wasn't as clear then :) Thanks!

radicle/src/identity/project.rs
FintanH: `abdicate`? :eyes: Or other synonyms https://www.merriam-webster.com/thesaurus/abdicate I kind of like `rescind` too :smile:

radicle-cli/src/commands/delegate.rs
FintanH: Whoops, think this was an IDE bug when adding the files from my Emacs. Guess it doesn't play well with that `path` config

radicle-cli/src/commands/add.rs
FintanH: So, initially, I thought the same. But is there any harm in them adding themselves as a proposal? If the current delegates don't agree then the document shouldn't verify.

FintanH:
> Looking good. By the way since we have a separate `threshold` property, we can actually create valid updates with any number of delegates, as long as `threshold = 1`.

Wait... does that mean I can wipe out a delegate set by removing them and setting the threshold to 1?

FintanH:
> > Looking good. By the way since we have a separate `threshold` property, we can actually create valid updates with any number of delegates, as long as `threshold = 1`.
> 
> Wait... does that mean I can wipe out a delegate set by removing them and setting the threshold to 1?

Actually, was able to confirm that one can't do this because the threshold can't be reached with the new delegate.

FintanH:
I'm curious why the quorum is checked during the `load` of an untrusted identity though (see https://github.com/radicle-dev/heartwood/blob/master/radicle/src/identity/project.rs#L441-L448). I would have expected it to be in the `verified` method.

radicle/src/identity/project.rs
cloudhead: I'm ok with `rescind`, though we lose the symmetry with `delegate` :shrug:

FintanH: Well I actually looked up antonyms of delegate, as in "to delegate", and abdicate was what came up :) So the symmetry is there ;)

radicle-cli/src/commands/add.rs
cloudhead: Yes but here it won't be a proposal, it's going to write that update to your *actual* identity branch for the project. I think for proposals we have to handle them quite differently, and in coordination with `rad patch` :thinking:

cloudhead:
> Wait... does that mean I can wipe out a delegate set by removing them and setting the threshold to 1?

If the threshold is `1`, yes, any of the delegates can do that. All edits to the document follow the same rule of requiring `threshold` signatures to pass.

cloudhead:
> I'm curious why the quorum is checked during the `load` of an untrusted identity though (see https://github.com/radicle-dev/heartwood/blob/master/radicle/src/identity/project.rs#L441-L448). I would have expected it to be in the `verified` method.

I think for two reasons:
1. It made sense due to the load function going through the entire history. Otherwise we'd have to traverse the history twice.
2. "Trusted" here is about signatures/peers, whereas "Verified" is around it actually matching the ID, given that the signatures are already checked. Basically, a history can be valid in the sense that all updates are valid, but that's not so useful until you check that it's the identity you requested. Hence the distinction here between "trusted" and "verified".

radicle-cli/src/commands/add.rs
FintanH: Oh right, you did mention that :monocle_face: Hmmmm does this CLI command make any sense then, in hindsight? I'm thinking it might be worth seeing what you have in the identity RIP before going any deeper here, wdyt?

cloudhead: It does make sense, as I said for a `threshold = 1` situation, where any delegate can update the doc with this command.

FintanH:
> > I'm curious why the quorum is checked during the `load` of an untrusted identity though (see https://github.com/radicle-dev/heartwood/blob/master/radicle/src/identity/project.rs#L441-L448). I would have expected it to be in the `verified` method.
> 
> I think for two reasons:
> 
>     1. It made sense due to the load function going through the entire history. Otherwise we'd have to traverse the history twice.
> 
>     2. "Trusted" here is about signatures/peers, whereas "Verified" is around it actually matching the ID, given that the signatures are already checked. Basically, a history can be valid in the sense that all updates are valid, but that's not so useful until you check that it's the identity you requested. Hence the distinction here between "trusted" and "verified".

In my mind, verification is the process of checking the signatures (i.e. _verifying_ the signatures), checking the quorum, and checking that `previous == parent`, which includes the `Id` I think.

cloudhead:
> > > I'm curious why the quorum is checked during the `load` of an untrusted identity though (see https://github.com/radicle-dev/heartwood/blob/master/radicle/src/identity/project.rs#L441-L448). I would have expected it to be in the `verified` method.
> > 
> > 
> > I think for two reasons:
> > ```
> > 1. It made sense due to the load function going through the entire history. Otherwise we'd have to traverse the history twice.
> > 
> > 2. "Trusted" here is about signatures/peers, whereas "Verified" is around it actually matching the ID, given that the signatures are already checked. Basically, a history can be valid in the sense that all updates are valid, but that's not so useful until you check that it's the identity you requested. Hence the distinction here between "trusted" and "verified".
> > ```
> 
> In my mind, verification is the process of checking the signatures (i.e. _verifying_ the signatures), checking the quorum, and checking that `previous == parent`, which includes the `Id` I think.

Yeah I'm not opposed to that, but I'm not sure how you'd write the API in that case in a clean way. I guess you could just load all commits into a Vec without verifying them, and then traverse that vec again on `verify`? It's redundant work though and requires keeping all commits in memory, but would work I guess. The other issue thing is it wouldn't't let you load the current identity unless you know the ID before hand.

FintanH:
> > > > I'm curious why the quorum is checked during the `load` of an untrusted identity though (see https://github.com/radicle-dev/heartwood/blob/master/radicle/src/identity/project.rs#L441-L448). I would have expected it to be in the `verified` method.
> > > 
> > > 
> > > I think for two reasons:
> > > ```
> > > 1. It made sense due to the load function going through the entire history. Otherwise we'd have to traverse the history twice.
> > > 
> > > 2. "Trusted" here is about signatures/peers, whereas "Verified" is around it actually matching the ID, given that the signatures are already checked. Basically, a history can be valid in the sense that all updates are valid, but that's not so useful until you check that it's the identity you requested. Hence the distinction here between "trusted" and "verified".
> > > ```
> > 
> > 
> > In my mind, verification is the process of checking the signatures (i.e. _verifying_ the signatures), checking the quorum, and checking that `previous == parent`, which includes the `Id` I think.
> 
> Yeah I'm not opposed to that, but I'm not sure how you'd write the API in that case in a clean way. I guess you could just load all commits into a Vec without verifying them, and then traverse that vec again on `verify`? It's redundant work though and requires keeping all commits in memory, but would work I guess. The other issue is it wouldn't't let you load the current identity unless you know the ID before hand.

I'm not advocating that we change it (yet :)) In what scenario would one not know the `ID` beforehand? It seems like the usage is always using `storage::Repository` where it has access to `Id`.

cloudhead:
> I'm not advocating that we change it (yet :)) In what scenario would one not know the ID beforehand? It seems like the usage is always using storage::Repository where it has access to Id.

Hmm true, it's more a matter of whether we want to couple it with storage/repository or not. I think if we find a nice way to pass the id down, then it should work. I think originally that wasn't so straightforward, but maybe it's worth trying now.

radicle-cli/src/commands/add.rs
FintanH: So it's completely limited to the case where `threshold = 1`. As soon as one bumps that the new document is verified but if they make any changes it'll say that the quorum is not reached, right? How useful is that if it only covers a single case? :thinking:

cloudhead: well no, you can just keep the threshold at `1` for ever. You get the github admin model then, where everyone in the delegate set is an "admin".

cloudhead: I'd argue this is the most common case actually that we're able to handle, ie. how most ACLs are setup on github right now.

FintanH: Eww :upside_down_face:

FintanH:
@cloudhead marking this as ready since the case is limited to `threshold = 1`. I can follow up with the other scenario with gathering signatures through a patch :)

radicle/src/identity/project.rs
```diff
@@ -20,42 +25,189 @@ pub enum ProjectError {
...
+            DefaultBranch,
+        }
+
+        struct ProjectVisitor;
+
+        impl<'de> Visitor<'de> for ProjectVisitor {
+            type Value = Project;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("Project payload")
```
cloudhead, line 58: Should we use the type id here for the string to be more specific? ie. `xyz.radicle.project`

radicle/src/identity/doc.rs
```diff
@@ -174,6 +178,24 @@ impl Doc<Verified> {
...
+    pub fn rescind(&mut self, key: &crypto::PublicKey) -> Result<Option<Did>, DocError> {
+        let delegate = Did::from(key);
+        let (matches, delegates) = self.delegates.iter().partition(|d| **d == delegate);
+        match NonEmpty::from_vec(delegates) {
+            Some(delegates) => {
+                self.delegates = delegates;
+                if self.threshold > self.delegates.len() {
+                    return Err(DocError::Threshold(
+                        self.threshold,
+                        "the thresholds exceeds the new amount of delegates after removal",
```
cloudhead, line 190:
```suggestion
                        "the thresholds exceeds the new delegate count after removal",
```
"amount" made it sound like we're measuring something :)

radicle/src/identity/doc.rs
```diff
@@ -174,6 +178,24 @@ impl Doc<Verified> {
...
+                    ));
+                }
+                Ok(matches.is_empty().not().then_some(delegate))
+            }
+            None => Err(DocError::Delegates("cannot remove the last delegate")),
+        }
+    }
```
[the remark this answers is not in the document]
FintanH: Thanks :blush:

radicle/src/identity/project.rs
FintanH: Ya, that makes sense!

radicle/src/identity/project.rs
```diff
@@ -20,42 +25,168 @@ pub enum ProjectError {
...
+        &self.name
+    }
+
+    pub fn description(&self) -> &String {
+        &self.description
+    }
+
+    pub fn default_branch(&self) -> &BranchName {
+        &self.default_branch
     }
```
slack-coder, line 190: Would inlining remove the overhead of the function call?  The methods seem added purely for safeguarding against changes.

```
#[inline]
pub fn default_branch(&self) -> &BranchName {
    &self.default_branch
}
```

radicle/src/identity/project.rs
```diff
@@ -20,42 +25,189 @@ pub enum ProjectError {
...
+            Err(ProjectError::DefaultBranch(
                 "default branch cannot exceed 255 bytes",
-            ));
+            ))
+        } else {
+            Ok(())
         }
-        Ok(())
     }
 }
```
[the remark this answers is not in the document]
slack-coder: Naive suggestion, what about doing something like:

```
project.update{|p| p.name = ...}
```

Its overkill too, but its also expressive and easy to read whats being updated.

radicle/src/identity/project.rs
FintanH: Ya, that makes sense to me!

radicle/src/identity/project.rs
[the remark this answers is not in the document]
FintanH: As opposed to the `update` I've changed it to?

What would that look like, because I'm not sure I follow :)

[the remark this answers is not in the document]
cloudhead: Yeah so like `p.update(|p| p.name = "foo")` instead of `p.update(Some("foo"), None, None)`

radicle-cli/src/commands/delegate/add.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let mut project = storage
+        .get(&profile.public_key, id)?
+        .context("No project with such ID exists")?;
+
+    let repo = storage.repository(id)?;
+
+    if !project.is_delegate(me) && key != *me {
+        return Err(anyhow::anyhow!("'{}' is not a current delegate of the project and does not have permission to add another person's key. However, they may add their own.", me));
+    }
```
cloudhead, line 26: This should be changed now to only check if you're a delegate and just say you're not a delegate as the error.

radicle/src/identity/project.rs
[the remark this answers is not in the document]
FintanH: Oh, I'm with you. Sure.

radicle-cli/src/commands/delegate/add.rs
cloudhead: Also the `key != me` should be removed now. We do probably want to make sure that the key being added isn't already in the delegate set, ie. `!project.is_delegate(key)`

radicle-cli/src/commands/delegate/remove.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    let mut project = storage
+        .get(&profile.public_key, id)?
+        .context("No project with such ID exists")?;
+
+    let repo = storage.repository(id)?;
+
+    if !project.is_delegate(me) && key != me {
+        return Err(anyhow::anyhow!("'{}' is not a current delegate of the project and does not have permission to remove another person's key. However, they may add their own.", me));
+    }
```
cloudhead, line 26: Same here

radicle/src/identity/project.rs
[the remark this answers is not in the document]
FintanH: It makes validating the new values a bit awkward, actually. Not impossible, but I'd probably reintroduce the validate functions again to make them reusable.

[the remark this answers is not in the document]
cloudhead: Hmm how come? It's just a matter of running validation on the payload every time something changes, no?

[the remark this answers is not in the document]
cloudhead: Let's merge as-is and I can propose changes we can discuss.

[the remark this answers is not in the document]
FintanH: Sounds good :+1:

PR #156, opened 2022-12-08
httpd: Conform response from issue handlers to previous model

xphoniex:


radicle-httpd/src/api/v1/projects.rs
```diff
@@ -360,14 +362,49 @@ async fn issue_handler(
         "id": issue_id,
         "author": issue.author(),
         "title": issue.title(),
-        "status": issue.status(),
-        "discussion": issue.comments().collect::<Vec<_>>(),
+        "state": issue.status(),
```
cloudhead, line 365: We should probably call this status to match the method name and underlying type?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -360,14 +362,49 @@ async fn issue_handler(
...
+    id: PublicKey,
+}
+
+#[derive(Serialize)]
+struct Comment {
+    author: Author,
+    body: String,
+    reactions: [String; 0],
+    timestamp: Timestamp,
+    reply_to: Option<PublicKey>,
```
cloudhead, line 384: This should be `Option<CommentId>`, but we still need to define what `CommentId` should be on the frontend. Perhaps just an index into the discussion list..

radicle-httpd/src/api/v1/projects.rs
xphoniex: `state` is what is expected, it'd end up like:
```
"state": { "status": "open|close", "reason": "" }
```

radicle-httpd/src/api/v1/projects.rs
xphoniex: fixed [here](https://github.com/radicle-dev/heartwood/compare/9cb50027918aee4a79a93b35645c0f6dbd698ddd..55bfe8f4b7fda0a9208b59f4cf45439ea8b44f1e#diff-85a73839d407aa508b9e78b3756b0e9b6e8d1087a0cafac6a8325a8cf427b65eR385)

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -360,14 +362,49 @@ async fn issue_handler(
...
+}
```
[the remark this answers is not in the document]
xphoniex: fixed [here](https://github.com/radicle-dev/heartwood/compare/9cb50027918aee4a79a93b35645c0f6dbd698ddd..55bfe8f4b7fda0a9208b59f4cf45439ea8b44f1e#diff-85a73839d407aa508b9e78b3756b0e9b6e8d1087a0cafac6a8325a8cf427b65eR379)

radicle-httpd/src/api/v1/projects.rs
cloudhead: Oh I see, then I should rename the method and type.

xphoniex: here's the [type](https://github.com/radicle-dev/radicle-interface/blob/master/src/issue.ts#L21-L28) on interface for reference.

cloudhead (approved):

PR #159, opened 2022-12-09
Assign issues using rad command + improve client example tests

slack-coder:
Introduce two changes:

 - Support command tool example tests by improving idempotence (https://github.com/radicle-dev/heartwood/pull/159/commits/9c5c0e89fdecb7a7c2ae98d039c7ddca67564f18)
 - Add assign/unassign subcommands (https://github.com/radicle-dev/heartwood/pull/159/commits/458fcaeff676c3faa61e181b1cecca8ff06e4c19)

Issue #148

### DIFF
--- a/radicle-cli/examples/rad-issue.md
+++ b/radicle-cli/examples/rad-issue.md
@@ -1,0 +1,34 @@
+Project 'todo' items are called 'issue's.  They can be inspected and modified
+using the 'issue' subcommand.
+
+Let's say the new car you are designing with your peers has a problem with its flux capacitor.
+
+```
+$ rad issue new --title "flux capacitor underpowered" --description "Flux capacitor power requirements exceed current supply"
+```
+
+The issue is now listed under our project.
+
+```
+$ rad issue list
+de81d97d7fe07a80bfb339200c6af862d4526b6a "flux capacitor underpowered"
+```
+
+Great! Now we've documented the issue for ourselves and others.
+
+Just like with other project management systems, the issue can be assigned to
+others to work on.  This is to ensure work is not duplicated.
+
+Let's assign ourselves to this one.
+
+```
+$ rad assign de81d97d7fe07a80bfb339200c6af862d4526b6a z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi
+$ rad issue list
+de81d97d7fe07a80bfb339200c6af862d4526b6a "flux capacitor underpowered" z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi
+```
+
+Note: this can always be undone with the 'unassign' subcommand.
+
+```
+$ rad unassign de81d97d7fe07a80bfb339200c6af862d4526b6a z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi
+```
--- a/radicle-cli/src/commands/assign.rs
+++ b/radicle-cli/src/commands/assign.rs
@@ -1,0 +1,96 @@
+use std::ffi::OsString;
+use std::str::FromStr;
+
+use anyhow::anyhow;
+
+use crate::terminal as term;
+use crate::terminal::args;
+use radicle::cob;
+use radicle::cob::issue;
+use radicle::storage::WriteStorage;
+
+pub const HELP: args::Help = args::Help {
+    name: "assign",
+    description: "assign an issue",
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad assign <issue> <peer>
+
+Options
+
+    --help      Print help
+"#,
+};
+
+#[derive(Debug)]
+pub struct Options {
+    pub id: issue::IssueId,
+    pub peer: cob::ActorId,
+}
+
+impl args::Args for Options {
+    fn from_args(args: Vec<OsString>) -> anyhow::Result<(Self, Vec<OsString>)> {
+        use lexopt::prelude::*;
+
+        let mut parser = lexopt::Parser::from_args(args);
+        let mut id: Option<issue::IssueId> = None;
+        let mut peer: Option<cob::ActorId> = None;
+
+        while let Some(arg) = parser.next()? {
+            match arg {
+                Long("help") => {
+                    return Err(args::Error::Help.into());
+                }
+                Value(ref val) => {
+                    if id.is_none() {
+                        let val = val.to_string_lossy();
+                        let Ok(val) = issue::IssueId::from_str(&val) else {
+                            return Err(anyhow!("invalid issue ID '{}'", val));
+                        };
+
+                        id = Some(val);
+                    } else if peer.is_none() {
+                        let val = val.to_string_lossy();
+                        let Ok(val) = cob::ActorId::from_str(&val) else {
+                            return Err(anyhow!("invalid peer ID '{}'", val));
+                        };
+
+                        peer = Some(val);
+                    } else {
+                        return Err(anyhow!(arg.unexpected()));
+                    }
+                }
+                _ => {
+                    return Err(anyhow!(arg.unexpected()));
+                }
+            }
+        }
+
+        Ok((
+            Options {
+                id: id.unwrap(),
+                peer: peer.unwrap(),
+            },
+            vec![],
+        ))
+    }
+}
+
+pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
+    let profile = ctx.profile()?;
+    let signer = term::signer(&profile)?;
+    let storage = &profile.storage;
+    let (_, id) = radicle::rad::cwd()?;
+    let repo = storage.repository(id)?;
+    let mut issues = issue::Issues::open(*signer.public_key(), &repo)?;
+
+    let mut issue = issues.get_mut(&options.id).map_err(|err| match err {
+        cob::store::Error::NotFound(_, _) => anyhow!("issue not found '{}'", options.id),
+        _ => err.into(),
+    })?;
+    issue.assign(vec![options.peer], &signer)?;
+
+    Ok(())
+}
--- a/radicle-cli/src/commands/issue.rs
+++ b/radicle-cli/src/commands/issue.rs
@@ -235,7 +235,17 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
         Operation::List => {
             for result in issues.all()? {
                 let (id, issue, _) = result?;
-                println!("{} {}", id, issue.title());
+
+                let assigned: String = issue
+                    .assigned()
+                    .map(|p| p.to_string())
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                if assigned.is_empty() {
+                    println!("{} \"{}\"", id, issue.title().escape_default());
+                } else {
+                    println!("{} {:?} {}", id, issue.title().escape_default(), &assigned,);
+                }
             }
         }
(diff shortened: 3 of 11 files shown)


FintanH:
Why wouldn't this be a subcommand for `issue` out of interest?

cloudhead:
> Why wouldn't this be a subcommand for `issue` out of interest?

In the future I'd like to be able to use it to assign users to other types of cobs. I do think we should have an alias under issue though, ie. `rad issue assign` should assign to issues only, vs. `rad assign` can assign to other supported cobs.

FintanH:
> > Why wouldn't this be a subcommand for `issue` out of interest?
> 
> In the future I'd like to be able to use it to assign users to other types of cobs. I do think we should have an alias under issue though, ie. `rad issue assign` should assign to issues only, vs. `rad assign` can assign to other supported cobs.

How would the "id" of the cob be parsed then? :thinking:

cloudhead:
> How would the "id" of the cob be parsed then? thinking

The plan is to have some kind of way to distinguish different cob types, eg. `patch:<oid>` or some kind of symbol eg. `%<oid>` for patches and `!<oid>` for issues.

radicle-crdt/src/lib.rs
```diff
@@ -40,6 +40,7 @@ pub trait Semilattice: Sized {
     }
 }
 
+// TODO: why does T have to be a semilattice?
```
FintanH, line 43: How would you merge them otherwise? :)

radicle-cli/src/commands/assign.rs
```diff
@@ -0,0 +1,99 @@
...
+use radicle::storage::WriteStorage;
+
+pub const HELP: Help = Help {
+    name: "assign",
+    description: "Manage issues",
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad assign <issue> @<alias>
```
FintanH, line 22: What's the `<alias>` for? And wouldn't you also need to parse the `ActorId`? :thinking:

radicle-cli/examples/rad-issue.md
```diff
@@ -0,0 +1,32 @@
...
+Let's assign ourselves to this one.
+
+```
+$ rad assign 522dcca93e8e172fcbd0d661326527949f2967ed z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi
+```
+
+Note: this can always be undone with the 'unassign' subcommand.
+
+```
+$ rad unassign 522dcca93e8e172fcbd0d661326527949f2967ed z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi
```
slack-coder, line 31: @cloudhead how can I make these issue and peer ID's in the example reproducible?

cloudhead: Check `radicle-cli/tests/commands.rs`

However, for COB IDs it might require something more, but at least ensure the user public key is stable, as that's one of the inputs to the COB ID.

cloudhead: Looking at the code, it seems like the ObjectId of the cob is the root commit `oid`. This means commit timestamps probably affects the ID. This is a bit tricky, as we'll need a way to force the timestamp to a certain value to make it deterministic.

FintanH:
Would it not also make sense to list the current assignees?

radicle-cli/src/commands/assign.rs
slack-coder: The alias is shorthand for another peer (radicle user).  Support will be added in upcoming changes, its been left out here.  For now, its the ActorID you mentioned.

slack-coder:
> Would it not also make sense to list the current assignees?

I agree, I'll add it as part of this PR for completeness.  It won't be pretty until support for using aliases and/or agree on how ID's can be abbreviated (like git's commit IDs can be).  

Maybe Actor ID's should follow the same rules as git uses for commits?

slack-coder:
Squashed and rebased to amend git commit messages.

radicle-cli/src/commands/issue.rs
```diff
@@ -235,7 +235,14 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
         Operation::List => {
             for result in issues.all()? {
                 let (id, issue, _) = result?;
-                println!("{} {}", id, issue.title());
+                let mut line = format!("{} \"{}\"", id, issue.title().escape_default());
+
+                let assigned: Vec<String> = issue.assigned().map(|p| p.to_string()).collect();
+                if !assigned.is_empty() {
+                    line = format!("{} {}", line, &assigned.join(", "));
+                }
```
cloudhead, line 243: It's clearer if you set `line` only once and keep it immutable, eg. `let line = if assigned { ... } else { ... };`

cloudhead:
This looks ready, modulo that one thing with `line`. I think I'd also prefer the word "determinism" instead of "idempotence" in the commit message, as this isn't about doing a repeated action but about pre-determining the result before doing the action.

Otherwise :+1:

slack-coder:
"I'd also prefer the word "determinism" instead of "idempotence" in the commit message"

Indeed, one important difference is idempotent methods can have an effect on state the first time its run.

radicle-cli/src/commands/issue.rs
```diff
@@ -235,7 +235,22 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+
+                let assigned: String = issue
+                    .assigned()
+                    .map(|p| p.to_string())
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                if assigned.is_empty() {
+                    println!("{} \"{}\"", id, issue.title().escape_default());
+                } else {
+                    println!(
```
FintanH, line 247: Actually, how come you're using `println` here instead of the terminal code?

radicle-cob/src/backend/git/change.rs
```diff
@@ -235,13 +236,22 @@ where
...
+
         let mut headers = commit::Headers::new();
         headers.push(
             "gpgsig",
             &String::from_utf8(crypto::ssh::ExtendedSignature::from(signature).to_armored())?,
         );
-        let author = commit::Author::try_from(&author)?;
+        let mut author = commit::Author::try_from(&author)?;
+
+        if let Ok(s) = std::env::var(git::RAD_COMMIT_TIME) {
```
FintanH, line 248: I don't wanna hold up this PR but I'm wondering if there is a way to ensure that we're using a debug build for this? I'm not sure how the testing framework works and if it uses the debug target of the binary. Would be good to make sure this doesn't run in released code :)

FintanH (approved):
LGTM just a few thoughts :)

radicle-cob/src/backend/git/change.rs
cloudhead: Yeah, we do exactly that when checking `RAD_DEBUG`:

```rust
        #[cfg(debug_assertions)]
        if env::var("RAD_DEBUG").is_ok() {
            // Generate a test keypair that is always the same.
            // This is useful for debugging and testing, since the
            // public key is known in advance.
            return KeyPair::from_seed(Seed::new([0xff; 32]));
        }
        KeyPair::generate()
```
All that's needove the if.ed is that pragma above the `if`.

radicle-cli/src/commands/issue.rs
cloudhead: Yea probably a good idea. I'd also prefer if we used `{:?}` for printing a quoted string, as escape sequences are hard to read.

slack-coder: Is there a better way to print quoted strings without operating through debug output?  Escaping quotes may be better for reading, but using '{:?}' is using debug code out of the context of debugging.

radicle-cob/src/backend/git/change.rs
slack-coder: Will do this.  Maybe then we could panic on a badly formatted env var?  Since its in debug mode, we can assume the intent on setting it.

radicle-cli/src/commands/issue.rs
slack-coder: > Actually, how come you're using `println` here instead of the terminal code?

Using the terminal crate's Table formatter now.

cloudhead: > Is there a better way to print quoted strings without operating through debug output? Avoiding escaping quotes may be better for reading, but using '{:?}' is using debug code out of the context of debugging.

Hmm not really. We probably won't be using quotes anyway once we design this properly, so it's not something to worry about long term. If you prefer, you can simply print it unquoted for now.

radicle-cob/src/backend/git/change.rs
cloudhead: Yeah

radicle-cli/src/commands/issue.rs
cloudhead: I've given this more thought, and I think that if we want quoted strings, using `{:?}` is the right solution, because:
1. It does exactly that: it quotes the string and escapes any `"` characters
2. Given that it's part of `String`'s public API, we can rely on it not breaking
3. `{:?}` is a standard formatting option and `std::fmt::Debug` is a standard trait. Just because it's called `Debug` doesn't mean it should not be used in production. In this case, the debug output is exactly what we're looking for (if we choose to display quoted strings).

slack-coder:
All points have been addressed :+1:

radicle-cob/src/backend/git/change.rs
slack-coder: Added the panic B)

radicle-cli/src/commands/issue.rs
slack-coder: Ok, I'll use this more consistently moving forwards.

radicle-cli/src/commands/assign.rs
```diff
@@ -0,0 +1,96 @@
...
+    let repo = storage.repository(id)?;
+    let mut issues = issue::Issues::open(*signer.public_key(), &repo)?;
+
+    let mut issue = issues.get_mut(&options.id).map_err(|err| match err {
+        cob::store::Error::NotFound(_, _) => anyhow!("issue not found '{}'", options.id),
+        _ => err.into(),
+    })?;
+    issue.assign(vec![options.peer], &signer)?;
+
+    Ok(())
```
FintanH, line 95: What's our general policy on finishing with `term::success`? I noticed that we use it some places like `track` or `patch`, but not in `issue`.

Not a blocker for this, but I'm curious :)

FintanH (approved):
Nice work :ok_hand:

radicle-cli/src/commands/assign.rs
slack-coder: I am curious too, maybe we can codify these in the crates documentation?

cloudhead: "finished" commands will usually end with `term::success`, but the issue commands haven't been polished yet, hence it's just the functionality and none of the visuals.

radicle-cli/src/commands/issue.rs
```diff
@@ -233,10 +233,22 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
-                println!("{} {}", id, issue.title());
+
+                let assigned: String = issue
+                    .assigned()
+                    .map(|p| p.to_string())
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                t.push([
+                    id.to_string(),
+                    format!("{:?}", issue.title().escape_default().to_string()),
```
cloudhead, line 247: `issue.title()` should be enough?

PR #162, opened 2022-12-10
Completed integration of the new reactor architecture

dr-orlovsky:
Supersedes 143


radicle-node/src/wire/protocol.rs
```diff
@@ -0,0 +1,315 @@
...
+        while let Some(ev) = self.inner.next() {
+            match ev {
+                Io::Write(node_id, msgs) => {
+                    log::debug!("Sending {} messages to {}", msgs.len(), node_id);
+                    let (fd, _) = self
+                        .sessions
+                        .iter()
+                        .find(|(_, id)| **id == node_id)
+                        .expect("sending message to the peer which is not connected");
+                    return Some(reactor::Action::Send(*fd, msgs));
```
cloudhead, line 260: Could we rename this `Send` variant? It's confusing for syntax highlighters :) Maybe `SendMessage`?

radicle-node/src/wire/protocol.rs
```diff
@@ -0,0 +1,315 @@
...
+    fn marshall(&self, mut writer: impl Write) -> Result<usize, Self::Error> {
+        self.encode(&mut writer).map_err(wire::Error::from)
+    }
+}
+
+type Queue<N> = VecDeque<reactor::Action<NetAccept<Session<N>>, NetTransport<Session<N>, Message>>>;
+
+pub struct Wire<R, S, W, G, N: Negotiator> {
+    inner: service::Service<R, S, W, G>,
+    inner_queue: Queue<N>,
```
cloudhead, line 48: I don't understand what `inner_queue` is for: it doesn't seem to queue messages from the inner service as I originally thought. It seems like we push & pop it within this module :thinking:

radicle-node/src/wire/protocol.rs
```diff
@@ -0,0 +1,315 @@
...
+use std::io::{Read, Write};
+use std::net;
+use std::os::unix::io::AsRawFd;
+use std::os::unix::prelude::RawFd;
+use std::sync::Arc;
+use std::time::Duration;
+
+use cyphernet::addr::{Addr as _, HostAddr, PeerAddr};
+use cyphernet::crypto::Ecdh;
+use nakamoto_net::{DisconnectReason, Link, LocalTime};
```
cloudhead, line 11: What would it take to drop the nakamoto dependency now? Is it just these three types? Could we put them in `netservices`?

radicle-node/src/wire/protocol.rs
```diff
@@ -0,0 +1,315 @@
...
+
+type Queue<N> = VecDeque<reactor::Action<NetAccept<Session<N>>, NetTransport<Session<N>, Message>>>;
+
+pub struct Wire<R, S, W, G, N: Negotiator> {
+    inner: service::Service<R, S, W, G>,
+    inner_queue: Queue<N>,
+    handshakes: HashMap<RawFd, Link>,
+    sessions: HashMap<RawFd, NodeId>,
+    // We use vec and not set since the same node may have multiple `N` sessions and has to
+    // disconnect N-1 times (instead of disconnecting a single session)
```
cloudhead, line 52: This comment doesn't seem to match the field?

radicle-node/src/wire/protocol.rs
```diff
@@ -0,0 +1,315 @@
...
+    }
+}
+
+type Queue<N> = VecDeque<reactor::Action<NetAccept<Session<N>>, NetTransport<Session<N>, Message>>>;
+
+pub struct Wire<R, S, W, G, N: Negotiator> {
+    inner: service::Service<R, S, W, G>,
+    inner_queue: Queue<N>,
+    handshakes: HashMap<RawFd, Link>,
+    sessions: HashMap<RawFd, NodeId>,
```
cloudhead, line 50: Can I combine these two maps?

cloudhead: There's a bunch of places where we modify the session for eg. but not the handshake..

radicle-node/src/wire/protocol.rs
```diff
@@ -0,0 +1,318 @@
...
+                        "New session from the same peer {}, closing previous session",
+                        node_id
+                    );
+                    self.disconnect(
+                        *fd,
+                        DisconnectReason::DialError(
+                            io::Error::from(io::ErrorKind::AlreadyExists).into(),
+                        ),
+                    );
+                    return;
```
cloudhead, line 163: Is it possible that more than one session exists here for the same NodeId, that needs to be disconnected?

radicle-node/src/wire/protocol.rs
dr-orlovsky: It is used when the Wire needs to produce an action for the `Reactor` due to its own internal logic. Nothing to do with the `Service`

radicle-node/src/wire/protocol.rs
dr-orlovsky: 1. I need to understand better the way `LocalTime` works, maybe it deserves dedicated crate or can be moved into `amplify` if it useful in multiple projects.

2. I have something like `Link` in the `netservices`, at least I was doing it. If it is still there, we can switch on it.

3. Disconnect reason is strange thing for me, since it appeared I was able to deal w/o it in the new reactor and netservices, so potentially we can avoid it here as well.

radicle-node/src/wire/protocol.rs
dr-orlovsky: AFAIR objects get removed from `handshakes` at the moment when it is completed. I.e. this field is requried to track not fully connected peers. The `session`s are otherwise tracking completed handshakes and are "permanent". 

So while you can add `Link` to the `sessions` to have `HashMap<RawFd, (NodeId, Link)>` you still need `handshakes: HashMap<RawFd, Link>` since there is no `NodeId` for the incoming handshakes and you need to track handshakes which are not "full sessions" yet.

I do not see the commit you are referring to - github doesn't show it to me, probably you did a force-push.

radicle-node/src/wire/protocol.rs
dr-orlovsky: The rule is "one session per node id" always. Since it doesn't make sense otherwise and the proper business logic can't be constructed without that assumption (each session has a different noise encoder state, and for outcoming messages addressing some NodeId it can't be decided which session to use).

Thus, if a new connection happens with the peer with the same node id, either the connection must not happen (possible for outgoing connections) or the previous session has to be closed (for incoming connection) at the moment the handshake is done.

Again, github UI doesn't show me the code around you comment, so if it is not the case of what's happened, I need to do a fix.

radicle-node/src/wire/protocol.rs
cloudhead: Yup, either works for me.

radicle-node/src/wire/protocol.rs
cloudhead: Ahh, gotcha. So it's kind of like a "short circuit" where reactor inputs are processed directly by Wire.

dr-orlovsky: Correct

PR #163, opened 2022-12-11
Use the new COB batching capabilities

cloudhead:
See commits.

@FintanH we forgot to set the clock correctly in the previous change, so have a look at that.

### DIFF
--- a/radicle/src/cob/issue.rs
+++ b/radicle/src/cob/issue.rs
@@ -1,11 +1,13 @@
 use std::ops::{ControlFlow, Deref};
 use std::str::FromStr;
 
+use nonempty::NonEmpty;
 use once_cell::sync::Lazy;
-use radicle_crdt::clock;
-use radicle_crdt::{LWWReg, LWWSet, Max, Semilattice};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+
+use radicle_crdt::clock;
+use radicle_crdt::{LWWReg, LWWSet, Max, Semilattice};
 
 use crate::cob;
 use crate::cob::common::{Author, Reaction, Tag};
@@ -285,7 +287,7 @@ impl<'a, 'g> IssueMut<'a, 'g> {
     ) -> Result<OpId, Error> {
         let cob = self
             .store
-            .update(self.id, msg, action.clone(), signer)
+            .update(self.id, msg, NonEmpty::new(action.clone()), signer)
             .map_err(Error::Store)?;
         let clock = cob.history().clock().into();
         let timestamp = cob.history().timestamp().into();
@@ -363,7 +365,9 @@ impl<'a> Issues<'a> {
         let title = title.into();
         let description = description.into();
         let action = Action::Title { title };
-        let (id, issue, clock) = self.raw.create("Create issue", action, signer)?;
+        let (id, issue, clock) = self
+            .raw
+            .create("Create issue", NonEmpty::new(action), signer)?;
         let mut issue = IssueMut {
             id,
             clock,
--- a/radicle/src/cob/issue.rs
+++ b/radicle/src/cob/issue.rs
@@ -396,7 +396,7 @@ impl<'a> Issues<'a> {
                 tx.tag(tags.to_owned(), []);
             })?;
         // Just a sanity check that our clock is advancing as expected.
-        assert_eq!(clock.get(), 2);
+        debug_assert_eq!(clock.get(), 2);
 
         Ok(IssueMut {
             id,
--- a/radicle/src/cob/issue.rs
+++ b/radicle/src/cob/issue.rs
@@ -1,7 +1,6 @@
 use std::ops::{ControlFlow, Deref};
 use std::str::FromStr;
 
-use nonempty::NonEmpty;
 use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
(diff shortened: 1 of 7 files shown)


radicle/src/cob/issue.rs
```diff
@@ -355,26 +384,26 @@ impl<'a> Issues<'a> {
...
-        let (id, issue, clock) = self.raw.create("Create issue", action, signer)?;
-        let mut issue = IssueMut {
+        let (id, issue, clock) =
+            Transaction::initial("Create issue", &mut self.raw, signer, |tx| {
+                tx.edit(title);
+                tx.comment(description);
+                tx.tag(tags.to_owned(), []);
+            })?;
+        // Just a sanity check that our clock is advancing as expected.
+        assert_eq!(clock.get(), 2);
```
FintanH, line 399: I think you can do `debug_assert!` so it only runs in debug mode.

radicle/src/cob/patch.rs
```diff
@@ -710,27 +754,24 @@ impl<'a> Patches<'a> {
...
-
-        Ok(patch)
+        let (id, patch, clock) =
+            Transaction::initial("Create patch", &mut self.raw, signer, |tx| {
+                tx.revision(base, oid);
+                tx.edit(title, description, target);
+                tx.tag(tags.to_owned(), []);
+            })?;
+        // Just a sanity check that our clock is advancing as expected.
+        assert_eq!(clock.get(), 2);
```
FintanH, line 772: Same about `debug_assert!` here

radicle/src/cob/store.rs
```diff
@@ -99,11 +100,17 @@ where
...
-        action: T::Action,
+        message: &str,
+        actions: impl Into<NonEmpty<T::Action>>,
         signer: &G,
     ) -> Result<CollaborativeObject, Error> {
-        let changes = NonEmpty::new(encoding::encode(&action)?);
+        let changes = actions
+            .into()
+            .iter()
+            .map(encoding::encode)
```
FintanH, line 110: I added `try_map` to `NonEmpty` so you can use that instead :)

radicle/src/cob/store.rs
```diff
@@ -124,11 +131,17 @@ where
...
-        action: T::Action,
+        message: &str,
+        actions: impl Into<NonEmpty<T::Action>>,
         signer: &G,
     ) -> Result<(ObjectId, T, Lamport), Error> {
-        let contents = NonEmpty::new(encoding::encode(&action)?);
+        let contents = actions
+            .into()
+            .iter()
+            .map(encoding::encode)
```
FintanH, line 141: Same goes for here

radicle/src/cob/store.rs
```diff
@@ -187,6 +200,114 @@ where
     }
 }
 
+/// Allows operations to be batched atomically.
+#[derive(Debug)]
+pub struct Transaction<T: FromHistory> {
+    actor: ActorId,
+    start: Lamport,
+    clock: Option<Lamport>,
+    actions: Vec<T::Action>,
```
FintanH, line 209: Does making this `NonEmpty` make the rest of the code awkward to use?

FintanH (commented):
Nice catch on the clocks :)

radicle/src/cob/issue.rs
```diff
@@ -387,7 +416,7 @@ impl<'a> Issues<'a> {
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 #[serde(tag = "type", rename_all = "camelCase")]
 pub enum Action {
-    Title { title: String },
+    Edit { title: String },
```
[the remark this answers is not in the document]
cloudhead: I explained it in one of the last commits, but it's to be consistent with the `Patch` actions, where you can edit a couple of other things as well.

The way I've been thinking about it is that the `Edit` action edits the issue metadata, which for now is only the title, but in the future could be other things that are not worth having a dedicated action for.

Comment editing would go under the `Thread` action, since the comments are part of the Thread CRDT.

Does that make sense?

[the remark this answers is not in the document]
FintanH: > I explained it in one of the last commits, but it's to be consistent with the Patch actions, where you can edit a couple of other things as well.

Ah cool, sorry that I missed that :) 

> Does that make sense?

Yup, thanks for the explanation!

radicle/src/cob/store.rs
cloudhead: Yeah the problem is you need a different API then, eg. you can't do:
```rust
store.transaction(|tx| {
  let id = tx.foo();
  tx.bar(id);
})
```
Because it would start out empty. You also can't just pass in a list, because you want something that could be monadic here, eg. where state is threaded through so that you can refer to `OpId`s within the tx.

The more annoying thing actually is the `clock`, which also needs an "empty" state for "initial" transactions, ie. when a COB is being created. In that scenario, the first clock value is 0, so we'd have to set it to -1 as a starting value, which isn't possible, hence the `Option`.

FintanH: Right, I had a feeling as such. Makes sense :)

radicle/src/cob/store.rs
```diff
@@ -137,10 +134,7 @@ where
     ) -> Result<(ObjectId, T, Lamport), Error> {
         let contents = actions
             .into()
-            .iter()
-            .map(encoding::encode)
-            .collect::<Result<Vec<_>, _>>()?
-            .try_into()
+            .try_map(|e| encoding::encode(&e))
             .expect("the collection is always non-empty");
```
FintanH, line 138: Is this `expect` needed now? I wouldn't think so. I think it's working on the `Result` type now.

cloudhead: Oh right, because the list is NonEmpty to start with!

FintanH: :brain: big brain strats

PR #164, opened 2022-12-12
Verify identity root commit signatures

cloudhead:
In RIP-2, we specify that identity document root commits must include the signature of all founding delegates.

This commit adds the required signatures during document creation and verifies them on load.

### DIFF
--- a/radicle/src/identity.rs
+++ b/radicle/src/identity.rs
@@ -10,7 +10,6 @@ use thiserror::Error;
 use crate::crypto;
 use crate::crypto::{Signature, Verified};
 use crate::git;
-use crate::storage::git::trailers;
 use crate::storage::{ReadRepository, RemoteId};
 
 pub use crypto::PublicKey;
@@ -39,10 +38,6 @@ pub enum IdentityError {
     MissingRootSignatures,
     #[error("commit signature for {0} is invalid: {1}")]
     InvalidSignature(PublicKey, crypto::Error),
-    #[error("commit message for {0} is invalid")]
-    InvalidCommitMessage(Oid),
-    #[error("commit trailers for {0} are invalid: {1}")]
-    InvalidCommitTrailers(Oid, trailers::Error),
     #[error("threshold not reached: {0} signatures for a threshold of {1}")]
     ThresholdNotReached(usize, usize),
     #[error("identity document error: {0}")]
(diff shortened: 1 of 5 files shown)


radicle/src/identity.rs
```diff
@@ -101,50 +98,45 @@ impl Identity<Untrusted> {
...
 
-        let mut trusted = trusted.verified()?;
-        let mut current = root;
-        let mut signatures = Vec::new();
+        // Every identity founder must have signed the root document.
+        for founder in &root.doc.delegates {
+            if !root.sigs.iter().any(|(k, _)| k == &**founder) {
+                return Err(IdentityError::MissingRootSignatures);
+            }
+        }
```
FintanH, line 109: I believe you can do something like this:
```suggestion
        if !root.doc.delegates.all(|founder| root.sigs.contains_key(founder)) {
            return Err(IdentityError::MissingRootSignatures)
        }
```

radicle/src/identity/doc.rs
```diff
@@ -106,11 +112,28 @@ impl Deref for Payload {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DocAt {
+    /// The commit at which this document exists.
+    pub commit: Oid,
+    /// The document blob at this commit.
+    pub blob: Oid,
+    /// The parsed document.
+    pub doc: Doc<Verified>,
+    /// The validated commit signatures.
+    pub sigs: Vec<(PublicKey, Signature)>,
```
FintanH, line 125: Why wouldn't this be `HashMap` as well?

radicle/src/identity/doc.rs
```diff
@@ -122,6 +145,11 @@ impl<V> Doc<V> {
         repo.reference_oid(remote, &git::refs::storage::IDENTITY_BRANCH)
             .map_err(DocError::from)
     }
+
+    pub fn blob_at<R: ReadRepository>(commit: Oid, repo: &R) -> Result<git2::Blob, DocError> {
```
FintanH, line 149: I'm thinking it might be better to have these as helper functions rather than on `Doc` since they don't use `self` or return `Self`. It means we don't need to use it by specifying a `V` for `Doc` on calling the functions.

radicle/src/identity.rs
FintanH: Ah, you would need to change `sigs` to a `HashMap`

FintanH (approved):
Nice :+1:

radicle/src/identity/doc.rs
cloudhead: Interesting, yes we can look into that :+1:

PR #165, opened 2022-12-13
Update `radicle-git` to `79a94721366490053e2d8ac1c1afa14fb0c25f09`

xphoniex:


FintanH:
There's some overlap with @keepsimple1's exploration here https://github.com/radicle-dev/heartwood/pull/149

Ideally, we finalise the details on the API and publish a version of surf rather than relying on the revision.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -230,14 +232,11 @@ async fn tree_handler(
...
-        "path": &tree.path,
-        "entries": &tree.entries,
-        "info": &tree.info,
-        "stats": stats(&repo)?,
-    });
+    let repo = Repository::open(paths::repository(storage, &project))?;
+    let path = if path == "" { None } else { Some(&path) };
+    let tree = Tree::new(&repo, &sha, path)?;
+    let mut response = json!(tree);
+    response["stats"] = json!(repo.stats()?);
```
xphoniex, line 239: `Tree::new(..)` won't take an empty path so I have to do this. However the response from API has broken the interface.

JSON:
```
{
    "entries": [{
        "path": "",
        "name": "README",
        "lastCommit": null,
        "kind": "blob"
    }, {
        "path": "",
        "name": "test_dir",
        "lastCommit": null,
        "kind": "directory"
    }],
    "lastCommit": {
        "sha1": "4f30784e586d09c0fb8d6c38ea35d6fd255ae10a",
        "author": {
            "name": "xphoniex",
            "email": "xphoniex@users.noreply.github.com"
        },
        "summary": "Initial commit",
        "description": "",
        "committer": {
            "name": "xphoniex",
            "email": "xphoniex@users.noreply.github.com"
        },
        "committerTime": 1670943683
    },
    "name": "",
    "path": "",
    "stats": {
        "commits": 1,
        "branches": 1,
        "contributors": 1
    }
}
```

Browser:
Screenshot from 2022-12-13 19-06-20

It's because both entries have the same path. What's the issue?

cc @FintanH

FintanH: Ah, was `"path"` the full file path originally? The change we made was that `path` was where the file was located and `name` was the name of the blob. So `full_path = path ++ name`. This can be changed in the serialize instance for `TreeEntry` on the surf side.

xphoniex: I should be able to update the PR to the new hash once you change it on the surf side.

cloudhead:
> There's some overlap with @keepsimple1's exploration here #149
> 
> Ideally, we finalise the details on the API and publish a version of surf rather than relying on the revision.

Yeah, so #149 as I understand has some additional changes that are not even in master yet. I think it would be good if we could at least update to the current master, and then afterwards update again with other changes? The current situation is making our CI fail every other time due to `cargo doc` failing because of conflicting files..

FintanH:
> > There's some overlap with @keepsimple1's exploration here #149
> > Ideally, we finalise the details on the API and publish a version of surf rather than relying on the revision.
> 
> Yeah, so #149 as I understand has some additional changes that are not even in master yet. I think it would be good if we could at least update to the current master, and then afterwards update again with other changes? The current situation is making our CI fail every other time due to `cargo doc` failing because of conflicting files..

Ya for sure! I was just making sure @xphoniex knew there was similar work, just in case they missed it :)

keepsimple1:
I'm working on get the additional `radicle-surf` changes in its master.  Will update here once they are in. Thanks.

xphoniex:
> I'm working on get the additional `radicle-surf` changes in its master. Will update here once they are in. Thanks.

We can merge once [this](https://github.com/radicle-dev/heartwood/pull/165#discussion_r1047338509) is fixed.  Can you help with this?

keepsimple1:
I've updated https://github.com/radicle-dev/heartwood/pull/149 to use the patch from https://github.com/radicle-dev/radicle-git/pull/79 .  The main things: 

1. `repo.tree()` supports `path` being empty.
2. `TreeEntry` has `lastCommit` included.
3.  `Tree` itself does not include `path`. I propose the caller uses the `path` from its input parameters. This also avoids a potential inconsistency between the input path and the path in Tree (if Tree were to include `path`).
4. `TreeEntry` has `name` included.

PR #167, opened 2022-12-13
Start adding support for replies in COBs

cloudhead:
We add a new thread action: `initialize`, which initializes a thread with a "root" comment and doesn't require a parent to be specified.

The comment actions can then be made to always require a `reply_to`, which should be the initial comment, or some other comment.


FintanH (commented):
LGTM :)

PR #171, opened 2022-12-19
List your issues using 'issue list' and show issue info via 'issue show'

slack-coder:
Add support for listing a users assigned issue via 'issue list', and show issue information with 'issue show <issue id>'

These changes depend on the earlier pull request (https://github.com/radicle-dev/heartwood/pull/159)

### DIFF
--- a/radicle-cli/src/commands/issue.rs
+++ b/radicle-cli/src/commands/issue.rs
@@ -43,17 +43,31 @@ pub struct Metadata {
 #[derive(Debug, PartialEq, Eq)]
 pub enum OperationName {
     Create,
-    Default,
-    Show,
-    State,
-    React,
     Delete,
     List,
+    None,
+    React,
+    Show,
+    State,
 }
 
 impl Default for OperationName {
     fn default() -> Self {
-        Self::Default
+        Self::None
+    }
+}
+
+/// Command line Peer argument.
+#[derive(Debug, PartialEq, Eq)]
+pub enum OptPeer {
+    Any,
+    CurrentUser,
+    Peer(cob::ActorId),
+}
+
+impl Default for OptPeer {
+    fn default() -> Self {
+        OptPeer::Any
     }
 }
 
@@ -63,7 +77,6 @@ pub enum Operation {
         title: Option<String>,
         description: Option<String>,
     },
-    Default,
     Show {
         id: IssueId,
     },
@@ -79,7 +92,7 @@ pub enum Operation {
         reaction: Reaction,
     },
     List {
-        assignee: Option<cob::ActorId>,
+        assignee: OptPeer,
     },
 }
 
@@ -95,7 +108,7 @@ impl Args for Options {
         let mut parser = lexopt::Parser::from_args(args);
         let mut op: Option<OperationName> = None;
         let mut id: Option<IssueId> = None;
-        let mut assignee: Option<cob::ActorId> = None;
+        let mut assignee: OptPeer = OptPeer::default();
         let mut title: Option<String> = None;
         let mut reaction: Option<Reaction> = None;
         let mut description: Option<String> = None;
@@ -141,13 +154,13 @@ impl Args for Options {
 
                     unknown => anyhow::bail!("unknown operation '{}'", unknown),
                 },
-                Value(val) if op == Some(OperationName::List) && assignee.is_none() => {
+                Value(val) if op == Some(OperationName::List) && assignee == OptPeer::Any => {
                     let val = val.to_string_lossy();
                     let Ok(val) = cob::ActorId::from_str(&val) else {
                             return Err(anyhow!("invalid peer ID '{}'", val));
                     };
 
-                    assignee = Some(val)
+                    assignee = OptPeer::Peer(val)
                 }
                 Value(val) if op.is_some() => {
                     let val = val
@@ -166,8 +179,10 @@ impl Args for Options {
         }
 
         let op = match op.unwrap_or_default() {
+            OperationName::None => Operation::List {
+                assignee: OptPeer::CurrentUser,
+            },
             OperationName::Create => Operation::Create { title, description },
-            OperationName::Default => Operation::Default,
             OperationName::Show => Operation::Show {
                 id: id.ok_or_else(|| anyhow!("an issue id must be provided"))?,
             },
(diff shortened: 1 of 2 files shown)


radicle-cli/src/commands/issue.rs
```diff
@@ -232,13 +257,18 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
-        Operation::List => {
+        Operation::List { assignee } => {
             let mut t = term::Table::new(term::table::TableOptions::default());
             for result in issues.all()? {
                 let (id, issue, _) = result?;
+                let assigned: Vec<_> = issue.assigned().collect();
+
+                if Some(true) == assignee.map(|a| !assigned.contains(&&a)) {
+                    continue;
+                }
```
cloudhead, line 281: We don't need to call `issuee.assigned()` if `assignee` is `None`, so I'd rewrite this with that in mind.

radicle-cli/src/commands/issue.rs
```diff
@@ -68,12 +70,14 @@ pub enum Operation {
...
-    List,
+    List {
+        assignee: Option<cob::ActorId>,
+    },
 }
 
 #[derive(Debug)]
 pub struct Options {
-    pub op: Operation,
+    pub op: Option<Operation>,
```
cloudhead, line 85: I find this change complicates a lot of the logic. I would keep it simple and default to showing all issues, which I think is a sensible default (ie. assignee: `None`)

radicle-cli/src/commands/issue.rs
slack-coder: Hmm its used a little further below to be displayed to the users.  Or did I miss something?

The assignee is intended to be whom to list issues for.  The None part of the option is to specify an 'all'.  Its an optional argument.

radicle-cli/src/commands/issue.rs
```diff
@@ -232,13 +255,19 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
                 )?;
             }
         }
-        Operation::List => {
+        Operation::Default => panic!("default should be set earlier"),
```
slack-coder, line 271: The text needs to be updated, suggestion: "default operation must be set to another"

radicle-cli/src/commands/issue.rs
```diff
@@ -18,11 +20,12 @@ pub const HELP: Help = Help {
     usage: r#"
 Usage
 
+    rad issue
     rad issue new [--title <title>] [--description <text>]
     rad issue state <id> [--closed | --open | --solved]
     rad issue delete <id>
     rad issue react <id> [--emoji <char>]
-    rad issue list
+    rad issue list <assignee>
```
FintanH, line 28:
```suggestion
    rad issue list [<assignee>]
```

FintanH, line 28: `show` is missing from the list

radicle-cli/src/commands/issue.rs
```diff
@@ -171,13 +196,24 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
     let repo = storage.repository(id)?;
     let mut issues = Issues::open(*signer.public_key(), &repo)?;
 
-    match options.op {
+    let op = match options.op {
+        Operation::Default => Operation::List {
+            assignee: Some(*profile.id()),
+        },
+        _ => options.op,
+    };
```
FintanH, line 204: Why do we need the `Default`? Can't `List` stay as the default operation?

radicle-cli/src/commands/issue.rs
```diff
@@ -171,13 +196,24 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+
+    match op {
         Operation::Create {
             title: Some(title),
             description: Some(description),
         } => {
             issues.create(title, description, &[], &signer)?;
         }
+        Operation::Show { id } => {
+            let issue = issues.get(&id)?.context("No issue with such ID exists")?;
```
FintanH, line 214: Reads a bit better to me
```suggestion
            let issue = issues.get(&id)?.context("No issue for the given ID exists")?;
```

radicle-cli/src/commands/issue.rs
```diff
@@ -245,3 +299,17 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+    println!("title: {}", issue.title());
+    println!("state: {}", issue.state());
+
+    let tags: Vec<String> = issue.tags().cloned().map(|t| t.into()).collect();
+    println!("tags: {}", tags.join(", "));
+
+    let assignees: Vec<String> = issue.assigned().map(|a| a.to_string()).collect();
+    println!("assignees: {}", assignees.join(", "));
+
+    println!("{}", issue.description().unwrap_or(""));
```
FintanH, line 313: Should these be `terminal::info`?

radicle-cli/src/commands/issue.rs
slack-coder: The behavior is defined as one of the 'dogfooding cli' issues https://github.com/radicle-dev/heartwood/issues/148

radicle-cli/src/commands/issue.rs
slack-coder: Updated to use info(..), but not entirely sure when and when it should be preferred.

radicle-cli/src/commands/issue.rs
slack-coder: :+1:

radicle-cli/src/commands/issue.rs
FintanH: Sorry, what's that got to do with this `Default` thing?

slack-coder: Maybe I misunderstood the question, the task is to make 'rad issue' list only your assigned issues, which requires your profile ID.

Fetching the profile information happens in run(..), and it didnt feel right fetching it when processing the command's arguments.

FintanH: Ah, I think I see now. I was thinking that if the `assignee` was `None` you could default to the profile. But I see now that there are three distinct functionalities.

I'd suggest capturing the listing functionality in a function and not doing this pre-match to also avoid the panic below.

slack-coder: How do you feel about this approach? https://github.com/radicle-dev/heartwood/pull/171/commits/422e28265f6a3fca1ef120390041d91a8b970891

FintanH (approved):
Chef's kiss :kissing_closed_eyes: :ok_hand:

PR #173, opened 2022-12-21
Improve terminal's Table

slack-coder:
Add a little documentation and make it easier to use by accepting `&str` over `String`.

### DIFF
--- a/radicle-cli/src/commands/self.rs
+++ b/radicle-cli/src/commands/self.rs
@@ -78,33 +78,30 @@ fn all(profile: &Profile) -> anyhow::Result<()> {
     let mut table = term::Table::default();
 
     let node_id = profile.id();
-    table.push([String::from("ID"), term::format::tertiary(node_id)]);
+    table.push(["ID", &term::format::tertiary(node_id)]);
 
     let ssh_short = ssh::fmt::fingerprint(node_id);
-    table.push([
-        String::from("Key (hash)"),
-        term::format::tertiary(ssh_short),
-    ]);
+    table.push(["Key (hash)", &term::format::tertiary(ssh_short)]);
 
     let ssh_long = ssh::fmt::key(node_id);
-    table.push([String::from("Key (full)"), term::format::tertiary(ssh_long)]);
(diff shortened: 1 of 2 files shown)


FintanH (commented):
I'm not sure the change to `&str` makes sense because you're allocating to a `String` anyway. But maybe it could take an array of `impl ToString`?

slack-coder:
> I'm not sure the change to `&str` makes sense because you're allocating to a `String` anyway. But maybe it could take an array of `impl ToString`?

Performance is low priority for the Table, but its now less verbose and easier to use.  At least that was the goal.

radicle-cli/src/commands/ls.rs
```diff
@@ -53,10 +53,10 @@ pub fn run(_options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
         table.push([
-            term::format::bold(name),
-            term::format::tertiary(id),
-            term::format::secondary(head),
-            term::format::italic(description),
+            &term::format::bold(name),
+            &term::format::tertiary(id),
+            &term::format::secondary(head),
+            &term::format::italic(description),
         ]);
```
FintanH, line 60: I don't think the `&` are needed anymore
```suggestion
            term::format::bold(name),
            term::format::tertiary(id),
            term::format::secondary(head),
            term::format::italic(description),
        ]);
```

radicle-cli/src/terminal/patch.rs
```diff
@@ -56,8 +56,8 @@ pub fn list_commits(commits: &[git::raw::Commit]) -> anyhow::Result<()> {
             .summary_bytes()
             .unwrap_or_else(|| commit.message_bytes());
         table.push([
-            term::format::secondary(term::format::oid(commit.id())),
-            term::format::italic(String::from_utf8_lossy(message)),
+            &term::format::secondary(term::format::oid(commit.id())),
+            &term::format::italic(String::from_utf8_lossy(message)),
```
FintanH, line 60: Same here :)

FintanH (approved):
Nice :+1:

FintanH (approved):

PR #174, opened 2022-12-21
CI job to publish CLI binaries

adaszko:
So far this job successfully produces (see https://github.com/radicle-dev/heartwood/actions/runs/3750372267):

```
radicle-cli-aarch64-apple-darwin.tar.gz
radicle-cli-aarch64-unknown-linux-musl.tar.gz
radicle-cli-x86_64-apple-darwin.tar.gz
radicle-cli-x86_64-unknown-linux-musl.tar.gz
radicle-cli_amd64.deb
```

To do before merging:
 * [ ] Replace `upload-artifacts` action with `action-gh-release` (it’s there because it’s much easier to test that way)
 * [ ] Remove the `on: push: adaszko/release-binaries` fragment (also needed for testing)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,126 @@
+name: Release
+on:
+  push:
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+"
+
+jobs:
+  release-x86_64-linux-binaries:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v3
+      - name: Configure build cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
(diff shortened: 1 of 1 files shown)


cloudhead:
Nice.

xphoniex:
Can we also add `radicle-httpd`?

adaszko:
> Can we also add `radicle-httpd`?

See https://github.com/radicle-dev/heartwood/pull/196

xphoniex:
> > Can we also add `radicle-httpd`?
> 
> See #196

Why not release `httpd` here on github as well, instead of pushing to gcloud storage?

.github/workflows/release.yml
```diff
@@ -0,0 +1,136 @@
+name: Release
+on:
+  push:
+    branches:
+      - adaszko/release-binaries
```
cloudhead, line 5: :eye:

PR #175, opened 2022-12-22
Conceptual worker thread implementation

dr-orlovsky:


radicle-node/src/service/reactor.rs
```diff
@@ -33,7 +33,8 @@ pub struct Fetch {
     repo: Id,
     /// Namespaces to fetch.
     namespaces: Namespaces,
-    // TODO: Needs session also.
+    /// Connection we are fetching from
+    remote: NodeId,
```
cloudhead: This is already passed into `Io::Fetch`

radicle-node/src/service.rs
```diff
@@ -884,8 +884,9 @@ where
                     }
                 }
             }
-            (session::State::Connected { .. }, Message::Upgrade { .. }) => {
-                todo!();
+            (session::State::Connected { .. }, Message::Upgrade { repo }) => {
+                // All we need is to instruct the transport to handover to the worker
+                self.reactor.fetch(*remote, repo, Namespaces::default());
```
cloudhead, line 905: We're going to need the direction of the fetch also right? Think I forgot to add that.

radicle-node/src/service.rs
```diff
@@ -156,9 +158,13 @@ pub enum Command {
     UntrackNode(NodeId, chan::Sender<bool>),
     /// Query the internal service state.
     QueryState(Arc<QueryState>, chan::Sender<Result<(), CommandError>>),
+    /// Worker has completed fetch job
+    FetchCompleted(Id, NetTransport<NoiseXk<G>, Message>),
+    /// Worker has failed fetch job
+    FetchFailed(Id, storage::Error, NetTransport<NoiseXk<G>, Message>),
```
cloudhead, line 164: These should be methods on `Service`, as they aren't RPC calls but rather events that happened.

cloudhead:
Superceded by #198

PR #176, opened 2022-12-23
Disable enforcing git UID = repo file UID by git

adaszko:


cloudhead:
Can we move the `gitconfig` to `.docker/gitconfig`? It would document the fact that it's only relevant for docker?

xphoniex:
This doesn't fix the issue of [`"config value 'safe.directory' was not found"`](https://github.com/radicle-dev/radicle-interface/actions/runs/3782403622/jobs/6430111819#step:12:13), it only changes the error into [`"repository path '/app/radicle/storage/<rad>/' is not owned by current user"`](https://github.com/radicle-dev/radicle-interface/actions/runs/3782782147/jobs/6430799760#step:11:290).

PR #188, opened 2022-12-30
Tidy up some COB things

cloudhead:
Some renames that were discussed, plus a new `OpId` type that is safer.

### DIFF
--- a/radicle-cob/src/object/collaboration/create.rs
+++ b/radicle-cob/src/object/collaboration/create.rs
@@ -20,8 +20,8 @@ pub struct Create {
 }
 
 impl Create {
-    fn create_spec(&self) -> change::Create<git_ext::Oid> {
-        change::Create {
+    fn template(&self) -> change::Template<git_ext::Oid> {
+        change::Template {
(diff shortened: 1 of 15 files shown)


radicle-cob/src/object/collaboration/create.rs
```diff
@@ -20,8 +20,8 @@ pub struct Create {
 }
 
 impl Create {
-    fn create_spec(&self) -> change::Create<git_ext::Oid> {
-        change::Create {
+    fn create_spec(&self) -> change::Template<git_ext::Oid> {
```
FintanH, line 23:
```suggestion
    fn template(&self) -> change::Template<git_ext::Oid> {
```

FintanH (approved):
LGTM :)

There's an empty file `radicle-cob/src/backend/git/objects.rs`, btw

radicle-cob/src/object/collaboration/create.rs
cloudhead: Good point

PR #189, opened 2022-12-30
cli: Avoid trailing space in `patch list`

slack-coder:
Trailing spaces on terminal output provide no use.  As they are hard to see and difficult to handle, remove them from `patch list`.

An alternative is to make term::print() trim trailing spaces instead.  This is avoided because it violates the principle of least surprise.

### DIFF
--- a/radicle-cli/src/commands/patch/list.rs
+++ b/radicle-cli/src/commands/patch/list.rs
@@ -39,7 +39,7 @@ pub fn run(
         }
     }
     term::blank();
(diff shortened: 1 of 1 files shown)


cloudhead:
Where is the trailing space coming from?

slack-coder:
> Where is the trailing space coming from?

"term::format::badge_positive()"

cloudhead:
Oh that's deliberate, since the colors are reversed, it wouldn't look correct if you removed that space.

PR #190, opened 2023-01-02
Setup for testing history permutations

cloudhead:
A few changes are made to make testing history permutations possible, and then the first test of that kind is written.

See individual commits for more detail.

### DIFF
--- a/radicle-cob/src/history/entry.rs
+++ b/radicle-cob/src/history/entry.rs
@@ -49,6 +49,7 @@ pub struct Entry {
     /// The actor that authored this entry.
     pub(super) actor: PublicKey,
     /// The content-address for the resource this entry lives under.
+    /// If the resource was updated, this should point to its latest version.
     pub(super) resource: Oid,
     /// The child entries for this entry.
     pub(super) children: Vec<EntryId>,
@@ -87,7 +88,7 @@ impl Entry {
         self.children.iter()
     }
 
-    /// The `Oid` of the resource this change lives under.
+    /// The current `Oid` of the resource this change lives under.
     pub fn resource(&self) -> Oid {
         self.resource
     }
--- a/radicle-cob/src/history/entry.rs
+++ b/radicle-cob/src/history/entry.rs
@@ -131,11 +131,11 @@ pub struct EntryWithClock {
     pub clock: Clock,
 }
 
-impl From<Entry> for EntryWithClock {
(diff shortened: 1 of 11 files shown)


radicle-cob/src/history/entry.rs
```diff
@@ -48,8 +48,6 @@ pub struct Entry {
     pub(super) id: EntryId,
     /// The actor that authored this entry.
     pub(super) actor: PublicKey,
-    /// The content-address for the resource this entry lives under.
-    pub(super) resource: Oid,
```
FintanH, line 52: I saw in the commit message you say this is redundant, why is that?

radicle-dag/src/lib.rs
```diff
@@ -130,6 +130,21 @@ impl<K: Eq + Copy + Hash, V> Dag<K, V> {
...
+        for k in other.tips.into_iter() {
+            self.tips.insert(k);
+        }
+        for k in other.roots.into_iter() {
+            self.roots.insert(k);
+        }
+        for (k, v) in other.graph.into_iter() {
+            self.graph.insert(k, v);
+        }
+    }
```
FintanH, line 146: This makes me wonder if there's a better use for a `merge_with` that uses a comparison function on the keys, essentially allowing the selection of a later key in a graph if we're involving clocks and such.

FintanH (approved):
Nice, looks good. Just a couple of comments from my side

radicle-cob/src/history/entry.rs
cloudhead: Ahh, actually this is a mistake! I thought it was the repo id, which never changes, but it's actually the latest document oid. I'll revert and add a bit more info to the docstring.

radicle-dag/src/lib.rs
cloudhead: Hmm yeah I was thinking *should Dag be a CRDT?*

If it were, we'd basically constrain keys to `Semilattice`, and then all inserts would merge. But I'm not sure if there's a use-case, because the way we use the DAG is that keys are `Oid` and therefore should only be overwritten if they point to the same commit. So in practice values should never be changed after a merge, only added.

FintanH: Makes sense! Thanks for the explanation :)

FintanH (approved):

PR #192, opened 2023-01-03
WIP: identity patching

FintanH:
This is a WIP draft of a COB that can handle patching an identity by gathering signatures from delegates to replace a previous document with a proposed document.


cloudhead:
Ahh interesting! You don't think it makes sense to use the Patch COB for this?

FintanH:
> Ahh interesting! You don't think it makes sense to use the Patch COB for this?

There were a couple of reasons I thought the `Patch` COB wasn't quite right for this, but maybe it could be reworked. It was a useful exercise for me to work through this at least, anyway :) So...

The main goal of this new COB is to gather the `Signature`s so that we can make a valid `Identity` commit with the latest `Doc`. The `Patch` does not actually have `Signature`s for `Action::Review` -- which incidentally might be a good thing for verifiable reviews. Even if it did, we would probably want the `Signature` over the latest `Oid` in the case of code review, but we want the `Signature` over the `Doc`, in our new case -- that, however, could be fixed by changing the signing to use the `Oid` instead, which is actually what Link was doing.

The other thing that made me avoid it is `Action::Merge`. As I said above, the aim was to gather signatures but I was aiming to avoid committing the final result until all of the signatures are gathered. The merge action seems to allow doing that beforehand. The way I was thinking about this was that `Identity` would have another method, something like:
```rust
pub fn merge(&self, proposal: Proposal) -> Result<Identity, Error> {
  /* 1. Checks that proposal.previous.current === self.current
     2. Checks that proposal.passes_quorum()
     3. Commit proposal.proposed
     4. Create the new Identity
  */
}
```

Thoughts? :)

cloudhead:
I see! I need to think about this, but there is one benefit in using the patch cob, which is that UIs we build would be re-usable quite easily. Perhaps it's worth thinking about an "identity" mode for patches, though I also don't want to have a bunch of if-statements everywhere, so needs some good design. I think by parametrizing one or two things, we should be able to make the patch cob work in this scenario. Eg. maybe the merge rule is customizable using some trait or closure; and perhaps the signing function, ie. what is signed and how the signatures are stored is also customizable.

cloudhead:
Great, yes there's value in just sketching this out before we commit to a design.

Where are we signing raw bytes? I'm confused :D

FintanH:
> Where are we signing raw bytes? I'm confused :D

Specifically for signing the `Doc` we use the `bytes` instead of the revision's `oid`:
https://github.com/radicle-dev/heartwood/blob/e49226c50023e012657c022009d58489054d6f99/radicle/src/identity/doc.rs#L210-L215

cloudhead:
Ah in the doc, yes. We can definitely sign the blob oid instead.

cloudhead:
I wonder though, whether we sign the bytes or oid though, it's still not a valid commit signature :thinking: -- which means we're forced to add the sig to the trailer.. Would it make sense to create a commit signature instead?

FintanH:
> I wonder though, whether we sign the bytes or oid though, it's still not a valid commit signature thinking -- which means we're forced to add the sig to the trailer.. Would it make sense to create a commit signature instead?

The `Oid` is the blob's `Oid` though not the commit :) But also realised that it doesn't make any sense for my point about `Patch` above :facepalm:

cloudhead:
Yeah, I know it's the blob's that's why I'm suggesting if the commit should be signed instead to be more git native..

FintanH:
> Yeah, I know it's the blob's that's why I'm suggesting if the commit should be signed instead to be more git native..

I'm not sure I follow :) The signature needs to be included in the identity's commit, no? So what would be the point of signing the commit?

cloudhead:
Commit signatures are included in commits via the gpgsig header :)

FintanH:
> Commit signatures are included in commits via the gpgsig header :)

Ya, but I believe that changes the SHA of the commit. I tested that quickly using the `git-commit` tests

cloudhead:
Ok, yeah that makes sense since the sha-1 is just taken on the entire commit payload. This does mean though that we can't use commit header signatures (gpgsig) for identity docs, as stated in the spec? Because those sign the entire commit. It means we're forced to using trailer signatures, which is what we currently do.

If we used commit header signatures, it would "work", but would be an abuse of git, because they wouldn't actually be signing the commit. So this means RIP#2 needs to be updated to use trailer sigs, right?

FintanH:
> Ok, yeah that makes sense since the sha-1 is just taken on the entire commit payload. This does mean though that we can't use commit header signatures (gpgsig) for identity docs, as stated in the spec? Because those sign the entire commit. It means we're forced to using trailer signatures, which is what we currently do.
> 
> If we used commit header signatures, it would "work", but would be an abuse of git, because they wouldn't actually be signing the commit. So this means RIP#2 needs to be updated to use trailer sigs, right?

Nowhere in our code are we signing the commit though. The signature is currently on the `Doc` payload. That's why I've gone down this route of "gathering signatures" via the `Proposal` COB, and then committing the final result with the signatures attached.

I'm not sure if there's any benefit to signing the commit itself, but perhaps I'm missing something.

FintanH:
Added a CLI subcommand for interacting with proposals. Here's a taster:

image

image


There's something strange going on with the `TextBox` and the ansii colouring in the diff output.

cloudhead:
> Nowhere in our code are we signing the commit though. The signature is currently on the `Doc` payload. That's why I've gone down this route of "gathering signatures" via the `Proposal` COB, and then committing the final result with the signatures attached.

Yes, that's correct, it's just the RIP that needs changing to match what's in the code. Now if we want to sign the git-object-hash of the blob instead for some reason, we can make that change as well.

FintanH:
> Looks cool, the UI; I'd anyway move away from TextBox as I have done with the inspect command, because they don't really add anything but make parsing the output really annoying, and break if there's line-wrapping involved..

Fair enough! Is there/Will there be a way of separating out the parts of the output? I liked how the boxes separate out the diff, acceptances, rejections, and quorum :)

FintanH:
Added publishing to the CLI command :) Ignore the first `list` output :sweat_smile: 

image

cloudhead:
> Fair enough! Is there/Will there be a way of separating out the parts of the output? I liked how the boxes separate out the diff, acceptances, rejections, and quorum :)

Yeah good question. There's no out-of-the-box way, but you can use headers like I use in patch, eg. `YOU PROPOSED`, with a blank line around it. There's the `terminal::io::headline` function which does that and bolds..

cloudhead:
Btw I'm still of the opinion that this should be part of the patch command, as it's essentially identical, and will need all the same sub-commands. Only if we think it's sufficiently different should it merit its  own subcommand and crdt..

FintanH:
> > Fair enough! Is there/Will there be a way of separating out the parts of the output? I liked how the boxes separate out the diff, acceptances, rejections, and quorum :)
> 
> Yeah good question. There's no out-of-the-box way, but you can use headers like I use in patch, eg. `YOU PROPOSED`, with a blank line around it. There's the `terminal::io::headline` function which does that and bolds..

Oh sweet, I'll take a look :ok_hand:

FintanH:
> Btw I'm still of the opinion that this should be part of the patch command, as it's essentially identical, and will need all the same sub-commands. Only if we think it's sufficiently different should it merit its own subcommand and crdt..

I can try playing around with that today idea today but I'm sceptical. I think they _look_ the same but have different outcomes. We'll see :)

FintanH:
> > Btw I'm still of the opinion that this should be part of the patch command, as it's essentially identical, and will need all the same sub-commands. Only if we think it's sufficiently different should it merit its own subcommand and crdt..
> 
> I can try playing around with that today idea today but I'm sceptical. I think they _look_ the same but have different outcomes. We'll see :)

The more I dig into this, trying to abstract out the parts that are different are pain points. I get into cases where I need some existential loading of an identity patch versus a code patch. That doesn't even get as far as changing the logic for `Patch` creation in the CLI. Whereas the proposal code I've written so far is lightweight and self-contained. 

I'm not sure what you had in mind for using `Patch`, so maybe I might have gone off-piste and got lost under an avalanche of generic type trickery.

@cloudhead how do you feel about proceeding? Because I'm not sure putting more time into this is worth it without some alignment on the approach going forward :)

cloudhead:
Ok, I'm fine with proceeding with a separate CRDT / CLI if you think doing otherwise would add too much complexity to Patch. I honestly haven't thought it through deeply, but it seemed like very similar workflows. Perhaps there are other points of convergence at a higher level, eg. if we implement code review for patches, we can re-use those UI components for identity change reviews, while keeping the backends different. I'm more interested in offering similar UX for both patch and identity than to have them implemented with the same datatypes. I wouldn't want to re-implement all our reviewing tools for both, basically.

So I'd say let's proceed with your original approach for now. I would however move the commands to `rad id *`, as `rad proposal` and `rad patch` are too similar.

Perhaps we could have something like `rad id edit` be the equivalent of `rad patch open`.

FintanH:
Sorry about the delay. Worked on the UI and have some screenshots:

help and list

show new identity proposal

accept proposal

reject proposal

show published proposal

I'll clean up and push what I have :)

radicle-crdt/src/gate.rs
```diff
@@ -0,0 +1,178 @@
+use crate::Semilattice;
+
+/// An optional value that once it is `closed`, it cannot be opened
+/// again.
```
cloudhead, line 4: I think this is `Redactable`?

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,680 @@
...
+    thread::{self, Thread},
+    Author, OpId,
+};
+
+/// The logical clock we use to order operations to patches.
+pub use clock::Lamport as Clock;
+
+/// Type name of a patch.
+pub static TYPENAME: Lazy<TypeName> =
+    Lazy::new(|| FromStr::from_str("xyz.radicle.identity.proposal").expect("type name is valid"));
```
cloudhead, line 33:
```suggestion
    Lazy::new(|| FromStr::from_str("xyz.radicle.id.proposal").expect("type name is valid"));
```
More familiar, since we use `rad/id` ref and `rad id` sub-command?

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,680 @@
...
+    Publish,
+    Redact {
+        revision: RevisionId,
+    },
+    Reject {
+        revision: RevisionId,
+    },
+    Revision {
+        proposed: Doc<Verified>,
+        previous: Identity<Oid>,
```
cloudhead, line 62: Do we really need to include the whole identity, or would the `oid` suffice?

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,680 @@
...
+                        }
+                        Some(Redactable::Redacted) => return Err(ApplyError::Redacted(revision)),
+                        None => return Err(ApplyError::Missing(revision)),
+                    }
+                }
+                Action::Edit { title, description } => {
+                    self.title.set(title, op.clock);
+                    self.description.set(description, op.clock);
+                }
+                Action::Publish => self.revisions.merge(Gate::closed()),
```
cloudhead, line 294: Ok, I think I understand what you're going for here, but I'm not sure it's the right approach :thinking: 

For one, I think I'd adopt the same design as patch/issue, where there is a state that can be "open" or "closed". It avoids needing this gate type, and doesn't overwrite the data, ie. it can still be displayed if necessary.

cloudhead: What you can do if you want to always merge towards a final state (least upper bound) is simply use a `Max<State>`, where the `State` enum's partial-ord is defined such that the greatest element is the "closed" state.

radicle/src/cob/identity.rs
FintanH: So I explored this and it's quite interesting. The `Identity` is used for getting the difference between the proposed document and the previous one as well as checking if a quorum is reached. Now, if I change this field to `Oid` it means we need to pass the previous `Identity` to those calls. In turn, that means `previous` wasn't actually used in any other way. So in essence we're always proposing the change against the current `Identity`. 

This seems to make sense to me, it's kind of like always comparing to the latest canonical branch. I'm trying to think if this has any downsides -- one potential might be that a review is being performed on a current `Identity` that is out of date, however, correct me if I'm wrong, this wouldn't happen because they pulled the identity proposal and so they should have _at least_ the identity that the peer proposed, right?

radicle/src/cob/identity.rs
FintanH: Ya, I suppose that could work :) I guess I was trying to be too clever by stoping any publishing happening after the initial publish, but that can just be a check on the state.

FintanH:
I'm happy with where this is at. I'm going to write some test examples for the CLI and then should be good to go :+1:

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,701 @@
...
+/// [`Identity`] then it may be published to the person's local
+/// storage using [`Proposal::publish`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Proposal {
+    /// Title of the proposal.
+    title: LWWReg<Max<String>>,
+    /// Description of the proposal.
+    description: LWWReg<Max<String>>,
+    /// Current state of the proposal.
+    pub state: LWWReg<Max<State>>,
```
cloudhead, line 135: This is already exposed as a method

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,701 @@
...
+            (State::Open, State::Open) => Ordering::Equal,
+            (State::Closed, State::Closed) => Ordering::Equal,
+            (State::Published, State::Published) => Ordering::Equal,
+            (State::Open, State::Closed) => Ordering::Less,
+            (State::Open, State::Published) => Ordering::Less,
+            (State::Closed, State::Open) => Ordering::Greater,
+            (State::Closed, State::Published) => Ordering::Less,
+            (State::Published, State::Open) => Ordering::Greater,
+            (State::Published, State::Closed) => Ordering::Greater,
+        }
```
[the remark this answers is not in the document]
cloudhead: If so, I'd probably also include a test to prevent the order being changed by mistake.

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,701 @@
...
+
+        Ok(())
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Verdict {
+    /// An accepting verdict must supply the [`Signature`] over the
+    /// new proposed [`Doc`].
+    Accept(Signature),
```
cloudhead, line 365: Are we still ok with signing the whole doc, vs. the blob id? It's still time to change that.

radicle-cli/src/commands.rs
```diff
@@ -14,6 +14,8 @@ pub mod rad_delegate;
 pub mod rad_edit;
 #[path = "commands/help.rs"]
 pub mod rad_help;
+#[path = "commands/identity.rs"]
+pub mod rad_identity;
```
cloudhead, line 18: I'd rename the file and module to use `id`, we try to match them exactly to the command name

radicle/src/cob/identity.rs
cloudhead: Yeah, what might happen though is that different users see different diffs, because their latest is different? I think having a `previous: Oid` (could be blob or commit id) might make sense to be able to throw an error in those cases.

So for example, I have an out of date id branch, I make a proposal, you receive it, and you see that it's a proposal for an old commit -- you reject it. Then I create a new up to date proposal after syncing.

radicle-cli/src/commands/identity.rs
```diff
@@ -0,0 +1,519 @@
...
+        Operation::Publish { id, rev } => {
+            let mut proposal = proposals.get_mut(&id)?;
+            let (rid, _) = publish_select(&proposal, rev, &previous, interactive)?;
+            let yes = confirm(interactive, "Are you sure you want to publish?");
+            if yes {
+                Proposal::publish(&proposal, &rid, signer.public_key(), &repo, &signer)?;
+                proposal.publish(&signer)?;
+                term::success!("Published new identity 🌱",);
+                print(&proposal, &previous, None)?;
+            }
```
cloudhead, line 344: Does this do the merge as well? ie. once all signatures have been collected, how does one commit to the identity branch?

radicle/src/cob/identity.rs
FintanH: iirc, signing the blob id made sense for trying to match the `Patch` cob, but that turned out to be tricky. The only pro I see for signing the doc is that it's stable in the face of the SHA scheme changing for Git. Otherwise, they seem equivalent.

wdyt?

radicle/src/cob/identity.rs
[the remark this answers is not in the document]
FintanH: Ya, I was ensuring the order so that it doesn't get changed by reordering the fields. Why rely on a test for this, out of interest?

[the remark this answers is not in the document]
cloudhead: My thought is that the above is better expressed by ordering the fields in the enum, as it's much more readable and less prone to error.

However, since other developers may not realize that the order matters, we should also have a test that just has like `asser!(x < y && y < z)` etc.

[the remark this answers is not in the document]
cloudhead: We do exactly this here for example, since the order of the fields matter: https://github.com/radicle-dev/heartwood/blob/master/radicle/src/cob/issue.rs#L478

radicle-cli/examples/rad-id-rebase.md
```diff
@@ -0,0 +1,430 @@
...
+
+✗ no
+```
+
+We can now update the proposal to have both keys in the delegates set:
+
+```
+$ rad id update dad0a006af01b2ee61d3839b0d331b260729b545 --rev z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi/4 --delegates z6MkedTZGJGqgQ2py2b8kGecfxdt2yRdHWF6JpaZC47fovFn --no-confirm
+ok Identity proposal 'dad0a006af01b2ee61d3839b0d331b260729b545' updated 🌱
+ok Revision 'z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi/6'
```
FintanH, line 299: @cloudhead I have this oddity that I'm trying to figure out. When this `update` is run then we get back this revision as the last `OpId`, `z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi/6`.

If [this line below](https://github.com/radicle-dev/heartwood/pull/192/files#diff-c40c5d42e0913666973cf38374515886a882e2144cf5063b0e13d671f94037e1R345) is changed to use `z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi/6` then it says the revision was missing. I was able to confirm this by printing out the existing revisions which were:
* z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi/1
* z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi/4
* z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi/5

So, what's happening to those other revisions? :thinking:

radicle/src/cob/identity.rs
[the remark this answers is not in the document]
FintanH: Cool, thanks for the explanation. I was thinking more along the lines of "The Code is Law", but happy to change it to this instead :)

radicle/src/cob/identity.rs
cloudhead: Hmm yeah, I guess signing hashes is much more standard generally, because you can verify the signature with less data and the signing is also faster. I actually think we should maybe move to signing the `git hash-object` hash of the blob (in a different pr though).

FintanH: Sounds good :)

radicle-cli/examples/rad-id-rebase.md
cloudhead: Hmm keep in mind that creating a revision might use multiple actions, in which case those gaps in the clocks are expected.

radicle/src/cob/identity.rs
FintanH: So, I've reworked this and now emit warnings when there's a mismatch. I also introduced the subcommand `rebase` that updates to the `Identity::current` in the next revision.

radicle-cli/examples/rad-id-rebase.md
FintanH: Hmmm ok, I'll have to check why the first revision is used as the one for the latest, i.e. why is `5` used instead of `6`

cloudhead: I'm guessing the wrong op-id is returned by the revision creation process, will take a look.

radicle/src/cob/identity.rs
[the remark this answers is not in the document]
cloudhead: Yeah, it's just one of those few cases where field order has a real consequence, and I can imagine someone organizing fields alphabetically breaking everything without realizing, if we don't have a test :)

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,746 @@
...
+    pub fn rejected(&self) -> Vec<PublicKey> {
+        self.verdicts()
+            .filter_map(|(key, v)| match v {
+                Verdict::Accept(_) => None,
+                Verdict::Reject => Some(*key),
+            })
+            .collect()
+    }
+
+    pub fn reaches_quorum(&self, previous: &Identity<Oid>) -> bool {
```
cloudhead, line 452: We have this convention for predicate functions to use `is_*` which I'd like to keep, so maybe `is_quorum_reached`?

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,746 @@
...
+        &mut self,
+        title: impl ToString,
+        description: impl ToString,
+        current: impl Into<Oid>,
+        proposed: Doc<Verified>,
+    ) -> (OpId, OpId) {
+        let revision = self.revision(current.into(), proposed);
+        let meta = self.edit(title, description);
+
+        (revision, meta)
```
cloudhead, line 548: I'm guessing this is the culprit of your revision problem.

Due to the `edit`, the clock is being incremented by one *after* the revision is created.
I'd consider splitting the edit outside of this function, or tying title/description to the revision.

cloudhead: I think in the Patch cob, we have a seperate edit and update function, which avoids this problem.

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,746 @@
...
+        proposed: Doc<Verified>,
+        signer: &G,
+    ) -> Result<(OpId, OpId), Error> {
+        self.transaction("Add revision", signer, |tx| {
+            let revision = tx.revision(current.into(), proposed);
+            let meta = tx.edit(title, description);
+
+            (revision, meta)
+        })
+    }
```
cloudhead, line 661: Same here, I'd split this out.

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,746 @@
...
+        description: impl ToString,
+        current: impl Into<Oid>,
+        proposed: Doc<Verified>,
+        signer: &G,
+    ) -> Result<ProposalMut<'a, 'g>, Error> {
+        let (id, patch, clock) =
+            Transaction::initial("Create proposal", &mut self.raw, signer, |tx| {
+                tx.revision(current.into(), proposed);
+                tx.edit(title, description);
+            })?;
```
cloudhead, line 720: This is the only case where I'd keep both revision & edit in the same tx/function.

radicle/src/cob/identity.rs
FintanH: Ah, right. The `meta` is on the `Proposal` and not the revision itself! :facepalm: Couldn't I just return `revision`? The `OpId` for `meta` is kind of useless, or is it?

FintanH: Ended up splitting them out, makes sense now :)

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,705 @@
...
+impl<'a> Deref for Proposals<'a> {
+    type Target = store::Store<'a, Proposal>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.raw
+    }
+}
+
+impl<'a> Proposals<'a> {
+    /// Open an patches store.
```
cloudhead, line 644: Can you update the docstrings?

radicle/src/cob/identity.rs
```diff
@@ -0,0 +1,705 @@
...
+    ) -> Result<OpId, Error> {
+        self.transaction("Accept", signer, |tx| tx.accept(revision, signature))
+    }
+
+    /// Reject a proposal revision.
+    pub fn reject<G: Signer>(&mut self, revision: RevisionId, signer: &G) -> Result<OpId, Error> {
+        self.transaction("Reject", signer, |tx| tx.reject(revision))
+    }
+
+    /// Edit patch metadata.
```
cloudhead, line 579:
```suggestion
    /// Edit proposal metadata.
```

PR #195, opened 2023-01-04
rad-remote-helper: honor RAD_PASSPHRASE

slack-coder:
Make testing more convenient by reducing its dependency on SSH Agent. It may be opted out of by setting the `RAD_PASSPHRASE` environment variable.

Also, reduce complexity by using the same function as the radicle-cli to source the signer.  It will ensure consistency across the commands.


radicle-cli/src/terminal/io.rs
```diff
@@ -176,20 +176,22 @@ pub fn abort<D: fmt::Display>(prompt: D) -> bool {
...
+    if signer.is_ok() || !interactive {
+        return signer.map(|s| s.boxed()).map_err(|e| e.into());
+    }
+
+    let passphrase = secret_input();
+    let spinner = spinner("Unsealing key...");
+    let signer = MemorySigner::load(&profile.keystore, passphrase)?;
+
+    spinner.finish();
+    Ok(signer.boxed())
```
cloudhead: This function is specifically designed for interactive use inside the cli, I don't think it makes sense to have a boolean change its behavior.. I'd just add a `passphrase_signer` function somewhere, and share that code with the remote helper.

slack-coder: > This function is specifically designed for interactive use inside the cli, I don't think it makes sense to have a boolean change its behavior.. I'd just add a `passphrase_signer` function somewhere, and share that code with the remote helper.

Where would you recommend?  Would the new dependency on 'radicle-cli' be considered a problem?  Term::io appears best but not perfect because of its use of env vars for input/configuration.

> I'd just add a passphrase_signer function somewhere, and share that code with the remote helper.

I'll rename and split the function(s).

slack-coder:
The patch was significantly changed.

Code for loading the profile's signer is moved to 'radicle::profile' so it can be shared with the radicle's git helper.

radicle/src/profile.rs
```diff
@@ -32,6 +34,14 @@ pub mod env {
     pub const RAD_SOCKET: &str = "RAD_SOCKET";
     /// Passphrase for the encrypted radicle secret key.
     pub const RAD_PASSPHRASE: &str = "RAD_PASSPHRASE";
+
+    pub fn read_passphrase() -> Option<super::Passphrase> {
+        let Ok(passphrase) = std::env::var(RAD_PASSPHRASE) else {
+            return None;
+        };
+
+        Some(super::Passphrase::from(passphrase.trim_end().to_owned()))
```
slack-coder, line 43: @cloudhead should we remove the 'trim_end' here?  Some people may have passphrases ending in spaces.

FintanH (commented):
Mucho bueno

radicle/src/profile.rs
cloudhead: :+1: 

It's coming from an env var here, but yeah we can remove the trim.

radicle/src/profile.rs
```diff
@@ -15,13 +15,15 @@ use std::path::{Path, PathBuf};
...
-use crate::crypto::ssh::agent::{Agent, AgentSigner};
+use crate::crypto::ssh::agent::Agent;
 use crate::crypto::ssh::{Keystore, Passphrase};
-use crate::crypto::PublicKey;
+use crate::crypto::{PublicKey, Signer};
 use crate::node;
 use crate::storage::git::transport;
 use crate::storage::git::Storage;
 
+use radicle_crypto::ssh::keystore;
```
cloudhead, line 25: Can you get this from `crate::crypto`?

radicle/src/profile.rs
```diff
@@ -32,6 +34,14 @@ pub mod env {
     pub const RAD_SOCKET: &str = "RAD_SOCKET";
     /// Passphrase for the encrypted radicle secret key.
     pub const RAD_PASSPHRASE: &str = "RAD_PASSPHRASE";
+
+    pub fn read_passphrase() -> Option<super::Passphrase> {
+        let Ok(passphrase) = std::env::var(RAD_PASSPHRASE) else {
+            return None;
+        };
+
```
cloudhead, line 42:
```suggestion
```

radicle-cli/src/terminal/io.rs
```diff
@@ -177,19 +177,16 @@ pub fn abort<D: fmt::Display>(prompt: D) -> bool {
 
 /// Get the signer. First we try getting it from ssh-agent, otherwise we prompt the user.
 pub fn signer(profile: &Profile) -> anyhow::Result<Box<dyn Signer>> {
-    let signer = if let Ok(passphrase) = read_passphrase_from_env_var() {
```
cloudhead, line 180: Was there a specific reason you removed the env var check here?

cloudhead:
Much better!

slack-coder:
Points addressed :+1:

FintanH (approved):

xphoniex (approved):

PR #196, opened 2023-01-05
Publish radicle-httpd binary for testing

adaszko:
On a push to master, this job is going to upload `radicle-httpd` binary to a Google Cloud Storage bucket at a path `gs://heartwood-artifacts/<THIS_REPO_COMMIT_SHA>/radicle-httpd` that can subsequently be used in e2e tests [[1]] by simply downloading and executing it (since it’s statically linked).  e2e tests will check if `radicle-httpd` at the given `COMMIT_SHA` pass playwright tests.

[1]: https://github.com/radicle-dev/radicle-interface/blob/master/.github/workflows/check-e2e.yml

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -23,46 +23,6 @@ jobs:
       - name: Run tests
         run: cargo test --all --verbose --all-features
 
-  release-x86_64-linux-binaries:
-    needs: build
-    permissions:
-      contents: 'read'
-      id-token: 'write'
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Checkout source code
-        uses: actions/checkout@v3
-      - name: Configure build cache
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          key: cargo-x86_64-linux-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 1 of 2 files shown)


adaszko:
> any reason why we're pushing `httpd` to gcr instead of github @adaszko ?

GitHub makes it hard to fetch artifacts from particular jobs.  Their retention policy is also hard to control.  It’s possible to use GitHub releases to store binaries but that would mean we (1) do a release on every push to master in the heartwood repo (2) actual releases that users might want are being drowned in cruft.  It’s just easier and cleaner to use a GCS bucket.

cloudhead:
This won't affect the PR checks though right? We don't want PRs to take longer to pass because of this.

adaszko:
> This won't affect the PR checks though right? We don't want PRs to take longer to pass because of this.

It will.  I can break it off into a separate workflow that runs only on pushes to `master` if that’s an issue.

cloudhead:
Hmm yeah it would make sense if it only built when we push to master actually. Same for the CLI.

adaszko:
> Hmm yeah it would make sense if it only built when we push to master actually

This is now corrected in this PR ✅ 

> Same for the CLI.

That’s already in place: For CLI, it happens when a git tag is pushed to this repo.

cloudhead:
Ah gotcha, thanks!

PR #204, opened 2023-01-06
httpd: Add api tests

xphoniex:


radicle-httpd/src/api/tests/scripts/init-e2e-profile
```diff
@@ -0,0 +1,30 @@
...
+git commit -m "Initial commit" --date $GIT_COMMITTER_DATE
+
+mkdir $WORKDIR/dir1
+echo "Hello World from dir1!" > dir1/README
+git add .
+git commit -m "Add another folder" --date $GIT_COMMITTER_DATE
+
+rad auth
+rad init --name "hello-world" --description "Rad repository for tests" --no-confirm
+rad issue new --title "Issue #1" --description "Change 'hello world' to 'hello everyone'"
```
xphoniex, line 30: there's no way to enforce `GIT_COMMITTER_DATE` to `rad issue new...`, explained further below in tests

radicle-httpd/src/api/v1.rs
```diff
@@ -18,3 +18,21 @@ pub fn router(ctx: Context) -> Router {
 
     Router::new().nest("/v1", routes)
 }
+
+#[cfg(test)]
+mod utils {
```
xphoniex, line 23: right place/name?

radicle-httpd/src/api/v1/delegates.rs
```diff
@@ -62,3 +62,29 @@ async fn delegates_projects_handler(
...
 }
+
+#[cfg(test)]
+mod routes {
+    use crate::api::v1::utils::{get_response, json, HEAD};
+
+    #[tokio::test]
+    async fn test_delegates_projects() {
+        let body = get_response(
+            "/v1/delegates/did:key:z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi/projects",
```
xphoniex, line 73: notice `delegates` route takes `did`, do we want to change that so we can take `<pk>` instead of `did:key:<pk>`? there are other places in api which takes plain `<pk>`, please advise where each should be used.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -467,3 +467,402 @@ fn stats(repo: &Repository, head: Oid) -> Result<Stats, Error> {
...
+    async fn test_projects_issues_root() {
+        let body = get_response(&format!(
+            "/v1/projects/rad:z4FucBZHZMCsxTyQE1dfE2YR59Qbp/issues"
+        ))
+        .await;
+
+        assert_eq!(
+            body,
+            json!([{
+                  "id":body[0]["id"],
```
xphoniex, line 846: `issues` are not stable (due to GIT_TIME), so they'll always have a different hash id, and timestamp.

radicle-httpd/src/api/v1/delegates.rs
cloudhead: We should always take DIDs here, in case we want to later introduce another key type.

radicle-httpd/src/api/v1.rs
cloudhead: Name is fine, but maybe should go one module up, since it's not related to `v1`, ie. in `api.rs`

radicle-httpd/src/api/v1.rs
```diff
@@ -18,3 +18,21 @@ pub fn router(ctx: Context) -> Router {
...
+    use hyper::{body, Client, Uri};
+    pub use serde_json::json;
+
+    const BASE_URL: &str = "http://0.0.0.0:8080/api";
+    pub const HEAD: &str = "609769c057e54e63ef6e3220bb65118fb9bf0d5a";
+
+    pub async fn get_response(url: &str) -> serde_json::Value {
```
cloudhead, line 30:
```suggestion
    pub async fn request(url: &str) -> serde_json::Value {
```

radicle-httpd/src/api/v1.rs
```diff
@@ -18,3 +18,21 @@ pub fn router(ctx: Context) -> Router {
...
+        let uri = format!("{}{}", BASE_URL, url).parse::<Uri>().unwrap();
```
cloudhead, line 31: https://docs.rs/hyper/latest/hyper/struct.Uri.html#method.builder

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -467,3 +467,393 @@ fn stats(repo: &Repository, head: Oid) -> Result<Stats, Error> {
...
+        assert_eq!(
+            body,
+            json!([{
+               "name":"hello-world",
+               "description":"Rad repository for tests",
+               "defaultBranch":"master",
+               "head":HEAD,
+               "patches":0,
+               "issues":1,
+               "id":"rad:z4FucBZHZMCsxTyQE1dfE2YR59Qbp"
```
cloudhead, line 488: Can we format all JSON the same, with a space between the `:` and value?, ie. `{"foo": 1}`

radicle-httpd/src/api/v1.rs
```diff
@@ -18,3 +18,21 @@ pub fn router(ctx: Context) -> Router {
...
+        let response = Client::new().get(uri).await.unwrap();
+        let body_bytes = body::to_bytes(response.into_body()).await.unwrap();
+        let body = String::from_utf8(body_bytes.to_vec()).unwrap();
+
+        serde_json::from_str(&body).unwrap()
```
cloudhead, line 36: We should be returning the status code as well (and checking it), and the headers just in case. Maybe create a `Response` type, or return a tuple.

radicle-httpd/Cargo.toml
```diff
@@ -22,7 +22,7 @@ chrono = { version = "0.4.22" }
 ethers-core = { version = "1.0" }
 fastrand = { version = "1.7.0" }
 flate2 = { version = "1" }
-hyper = { version = "0.14.17", default-features = false }
+hyper = { version = "0.14.17", default-features = false, features = ["client"] }
```
cloudhead, line 25: Only in dev-dependencies

radicle-httpd/src/api/v1.rs
xphoniex: I think this is the right place, because in future we might have a `v2` and it should have its own tests. Having them in `api.rs` would be confusing.

radicle-httpd/src/api/v1/delegates.rs
xphoniex: what about other routes? e.g.[ `node_handler`](https://github.com/radicle-dev/heartwood/blob/master/radicle-httpd/src/api/v1/node.rs#L20) takes `NodeId`, correct usage?

cloudhead: That's correct yeah.

radicle-httpd/src/api/v1.rs
xphoniex: moved it all to another file, `radicle-httpd/src/api/test.rs`

radicle-httpd/src/api/v1/projects.rs
xphoniex: fixed

radicle-httpd/src/api/test.rs
```diff
@@ -0,0 +1,128 @@
...
+        )
+        .unwrap();
+
+    Context {
+        profile: Arc::new(profile),
+        sessions: Default::default(),
+    }
+}
+
+pub async fn request(app: &Router, path: String) -> Response {
```
xphoniex, line 118: `request` can be re-written to take `path: impl ToString`, and then we can do:

```
let response = request(&app, "/stats").await;
```



instead of 
```
let response = request(&app, "/stats".to_string()).await;
```

but at the cost of doing another allocation inside the `request`:
```
Request::builder().uri(path) ---> Request::builder().uri(path.to_string())
```

worth it?

cloudhead: Yeah worth it!

xphoniex: https://github.com/radicle-dev/heartwood/compare/cf86b516b8ef8e58885a0c38b7c015113a3dd5cf..9114dccc40aa5a5e5f244b24184eea1d21b21cd3

radicle-httpd/src/api/test.rs
```diff
@@ -0,0 +1,133 @@
...
+use radicle_cli::commands::rad_init;
+
+use crate::api::Context;
+
+pub const HEAD: &str = "1e978d19f251cd9821d9d9a76d1bd436bf0690d5";
+pub const HEAD_1: &str = "f604ce9fd5b7cc77b7609beda45ea8760bee78f7";
+
+pub fn seed() -> Context {
+    let random_suffix = Alphanumeric.sample_string(&mut rand::thread_rng(), 16);
+    let temp_dir = env::temp_dir()
```
cloudhead, line 25: We prefer `tempfile::tempdir()`, because it gets auto-cleaned up at the end of the tests. Otherwise we're accumulating a bunch of data very quickly. However you'll have to pass it as a reference into this function for it to be dropped at the end of the test, and not the end of this function.

radicle-httpd/src/api/test.rs
```diff
@@ -0,0 +1,133 @@
...
+        .join("radicle-httpd-tests")
+        .join(random_suffix);
+    let workdir = temp_dir.join("hello-world");
+    let rad_home = temp_dir.join("radicle");
+
+    env::set_var("RAD_PASSPHRASE", "asdf");
+    env::set_var("RAD_DEBUG", "1");
```
cloudhead, line 32: Are these needed? We try to avoid env vars in tests if possible, for obvious reasons.

radicle-httpd/src/api/test.rs
```diff
@@ -0,0 +1,133 @@
...
+            set_upstream: false,
+        },
+        &profile,
+    )
+    .unwrap();
+
+    // eq. rad issue new
+    env::set_var("RAD_COMMIT_TIME", "1673001014");
+
+    let signer = radicle_cli::terminal::signer(&profile).unwrap();
```
cloudhead, line 98: I wouldn't use this here, you can simply create a `MemorySigner` from the profile keystore, with `MemorySigner::load(&profile.keystore)`

radicle-httpd/src/api/test.rs
xphoniex: Yes, they are needed, I think `rad init` needs them.

radicle-httpd/src/api/test.rs
```diff
@@ -0,0 +1,135 @@
...
+use radicle::git::raw as git2;
+use radicle::storage::WriteStorage;
+use radicle_cli::commands::rad_init;
+use radicle_crypto::ssh::keystore::MemorySigner;
+use radicle_crypto::Signer;
+
+use crate::api::Context;
+
+pub const HEAD: &str = "1e978d19f251cd9821d9d9a76d1bd436bf0690d5";
+pub const HEAD_1: &str = "f604ce9fd5b7cc77b7609beda45ea8760bee78f7";
```
cloudhead, line 23: Where do these come from?

radicle-httpd/src/api/test.rs
cloudhead: It's already a dependency though, and avoids needing `rand` which you pulled in.

radicle-httpd/src/api/test.rs
cloudhead: That's strange, can you double check?

xphoniex: Yea, I checked, if I remove them, prompt asks for password in tests, and if I remove debug, the identity won't be stable.

radicle-httpd/src/api/test.rs
xphoniex: they are hash of commits generated from our tree, it's stable and I've extracted them here to make further changes easier, e.g. if we change "fixture", like name of repo or sig or anything, we only update HEADs here and then good to go.

radicle-httpd/src/api/test.rs
xphoniex: [fixed](https://github.com/radicle-dev/heartwood/compare/0a4dd930144842337392c0d92496535c5e44d38a..a8270fefa9ef77337c08fd7f84b9d0db9428c23d), created wrapper `Response` too

radicle-httpd/src/api/test.rs
cloudhead: What prompts you?

cloudhead: I think I'm going to look into a more sustainable way of doing this, for now, can you set it to "radicle"? Same passphrase we use in the CLI tests. Because otherwise the test will fail if run in parallel with the CLI test.

xphoniex: [fixed](https://github.com/radicle-dev/heartwood/compare/3ab407822c6d0f3bc12a7b001ae24cf716eb8600..9c3fff3f60b89ece2b1281fbcb6be961adf8faf8)

Issue #206, opened 2023-01-06
Test installing package commands in CI


slack-coder:
Unexpected build errors for installing `radicle-cli` and `radicle-remote-helper` have been making it into the master branch.  Tests can be added to the CI pipeline to prevent this.

PR #207, opened 2023-01-06
Test installing commands in CI

slack-coder:
For an unknown reason, `cargo test` may succeed when compiling `radicle-cli` and `radicle-remote-helper` packages for installation fail.

Add a test to prevent these errors from reaching the master branch.

resolves: #206


slack-coder:
Not needed, `cargo install --locked` is the advised install method.

Follow up at https://github.com/radicle-dev/heartwood/pull/209

PR #212, opened 2023-01-09
cli: Update `inspect` command

cloudhead:
Fix order of iteration, and make output look more like commit log. Also show commit parent and message.

image

### DIFF
--- a/radicle-cli/src/commands/inspect.rs
+++ b/radicle-cli/src/commands/inspect.rs
@@ -149,10 +149,9 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
     } else if options.history {
         let repo = storage.repository(id)?;
         let head = Doc::<Untrusted>::head(signer.public_key(), &repo)?;
-        let history = repo.revwalk(head)?.collect::<Vec<_>>();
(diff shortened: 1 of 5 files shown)


FintanH (approved):
Nice :+1:

radicle-cli/src/commands/inspect.rs
```diff
@@ -149,10 +149,9 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
     } else if options.history {
         let repo = storage.repository(id)?;
         let head = Doc::<Untrusted>::head(signer.public_key(), &repo)?;
-        let history = repo.revwalk(head)?.collect::<Vec<_>>();
-        let revision = history.len();
+        let history = repo.revwalk(head)?.collect::<Vec<_>>().into_iter();
```
slack-coder, line 152: @cloudhead what is the reason to collect into a vector?  Are the iterators between revwalk() and Vec<_> different?

cloudhead: Actually that's no longer needed; as we collected to be able to reverse the iteration direction, but we no longer reverse, so could be removed.

PR #214, opened 2023-01-09
cli: alter use subcommands for patch

slack-coder:
Make `rad patch` accept subcommands for specifying its operation in place of the '--command' notation.

For instance `rad patch update` versus `rad patch --update`.  Also make the default operation list active patches, replacing patch creation. Patch creation is now done via `rad patch open`.

Make these changes in a way consistent with other commands like `rad delegate`.

Add '--no-confirm' to patch to support this subcommand's documentation test.

References: #148

### DIFF
--- a/radicle-cli/src/commands/patch.rs
+++ b/radicle-cli/src/commands/patch.rs
@@ -24,11 +24,13 @@ pub const HELP: Help = Help {
     usage: r#"
 Usage
 
-    rad patch [<option>...]
+    rad patch
+    rad patch open [<option>...]
+    rad patch update <id> [<option>...]
 
-Create options
+Create/Update options
 
-    -u, --update [<id>]        Update an existing patch (default: no)
+        --[no-]confirm         Don't ask for confirmation during clone
         --[no-]sync            Sync patch to seed (default: sync)
         --[no-]push            Push patch head to storage (default: true)
     -m, --message [<string>]   Provide a comment message to the patch or revision (default: prompt)
@@ -36,32 +38,45 @@ Create options
 
 Options
 
-    -l, --list                 List all patches (default: false)
         --help                 Print help
 "#,
 };
 
-#[derive(Debug)]
-pub enum Update {
-    No,
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+pub enum OptPatch {
+    #[default]
     Any,
+    None,
     Patch(PatchId),
 }
 
-impl Default for Update {
-    fn default() -> Self {
-        Self::No
-    }
+#[derive(Debug, Default, PartialEq, Eq)]
+pub enum OperationName {
+    Open,
+    Update,
+    #[default]
+    List,
 }
 
-#[derive(Default, Debug)]
+#[derive(Debug)]
+pub enum Operation {
+    Open {
+        message: Comment,
+    },
+    Update {
+        patch_id: OptPatch,
+        message: Comment,
+    },
+    List,
+}
+
+#[derive(Debug)]
 pub struct Options {
-    pub list: bool,
-    pub verbose: bool,
+    pub op: Operation,
+    pub confirm: bool,
     pub sync: bool,
     pub push: bool,
-    pub update: Update,
-    pub message: Comment,
+    pub verbose: bool,
 }
 
 impl Args for Options {
@@ -69,34 +84,23 @@ impl Args for Options {
         use lexopt::prelude::*;
 
         let mut parser = lexopt::Parser::from_args(args);
-        let mut list = false;
+        let mut confirm = true;
+        let mut op: Option<OperationName> = None;
         let mut verbose = false;
         let mut sync = true;
+        let mut id = OptPatch::default();
         let mut message = Comment::default();
         let mut push = true;
-        let mut update = Update::default();
 
         while let Some(arg) = parser.next()? {
             match arg {
-                // Operations.
-                Long("list") | Short('l') => {
-                    list = true;
-                }
-                Long("update") | Short('u') => {
-                    if let Ok(val) = parser.value() {
-                        let val = val
-                            .to_str()
-                            .ok_or_else(|| anyhow!("patch id specified is not UTF-8"))?;
-                        let id = PatchId::from_str(val)
-                            .map_err(|_| anyhow!("invalid patch id '{}'", val))?;
-
-                        update = Update::Patch(id);
-                    } else {
-                        update = Update::Any;
-                    }
-                }
-
                 // Options.
+                Long("confirm") => {
+                    confirm = true;
+                }
+                Long("no-confirm") => {
+                    confirm = false;
+                }
                 Long("message") | Short('m') => {
                     if message != Comment::Blank {
                         // We skip this code when `no-message` is specified.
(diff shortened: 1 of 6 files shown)


radicle-cli/src/commands/patch.rs
```diff
@@ -24,79 +24,93 @@ pub const HELP: Help = Help {
...
 "#,
 };
 
-#[derive(Debug)]
-pub enum Update {
-    No,
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum OptPatch {
     Any,
+    None,
```
cloudhead, line 48: Doesn't really work because then you can get an update with `None` as the patch_id, which is nonsensical. I'd remove this type completely and just have `Update { patch_id: Option<PatchId> }`

radicle-cli/src/commands/patch.rs
```diff
@@ -129,17 +143,43 @@ impl Args for Options {
                 Long("help") => {
                     return Err(Error::Help.into());
                 }
+
+                Value(val) if op.is_none() => match val.to_string_lossy().as_ref() {
+                    "list" => op = Some(OperationName::List),
+                    "open" => op = Some(OperationName::Create),
+                    "update" => op = Some(OperationName::Update),
```
cloudhead, line 150: Let's also support short options "l" "o" "u"

radicle-cli/src/commands/patch.rs
```diff
@@ -24,79 +24,93 @@ pub const HELP: Help = Help {
...
     fn default() -> Self {
-        Self::No
+        Self::List
     }
 }
 
-#[derive(Default, Debug)]
+#[derive(Debug)]
+pub enum Operation {
+    Create {
```
cloudhead, line 73: Rename to `Open` along with `OperationName`

radicle-cli/src/commands/patch/create.rs
```diff
@@ -194,14 +205,14 @@ pub fn run(
     term::patch::list_commits(&commits)?;
     term::blank();
 
-    if !term::confirm("Continue?") {
+    if !confirm("Continue?", &options) {
         anyhow::bail!("patch proposal aborted by user");
     }
 
-    let message = head_commit
+    let commit_message = head_commit
```
cloudhead, line 212: Why was this renamed?

radicle-cli/src/commands/patch.rs
```diff
@@ -129,17 +143,43 @@ impl Args for Options {
...
+                    "list" => op = Some(OperationName::List),
+                    "open" => op = Some(OperationName::Create),
+                    "update" => op = Some(OperationName::Update),
+
+                    unknown => anyhow::bail!("unknown operation '{}'", unknown),
+                },
+                Value(val) if op == Some(OperationName::Update) && id == OptPatch::Any => {
+                    let val = val
+                        .to_str()
+                        .ok_or_else(|| anyhow!("issue id specified is not UTF-8"))?;
```
cloudhead, line 157:
```suggestion
                        .ok_or_else(|| anyhow!("patch id specified is not UTF-8"))?;
```

cloudhead:
Looks good overall :+1:

radicle-cli/src/commands/patch.rs
slack-coder: I was following `rad issue`s lead on this one.  We should change there as well if we do here.

https://github.com/radicle-dev/heartwood/blob/master/radicle-cli/src/commands/issue.rs#L147

radicle-cli/src/commands/patch.rs
slack-coder: Its messy because working on the code required some surgery.  This is an update on the original code which does the same thing.

Both `update` and `create` use a single method whose operations relies on this argument.  Refactoring this meaningfully appeared outside of the issue's scope.

radicle-cli/src/commands/patch/create.rs
```diff
@@ -33,11 +35,18 @@ blank is also okay.
 -->
 "#;
 
+#[inline]
+fn confirm<D: fmt::Display>(prompt: D, options: &Options) -> bool {
+    !options.confirm || term::confirm(prompt)
+}
```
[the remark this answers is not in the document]
cloudhead: Yeah maybe you're right. Just feels weird to pass options into this function as well, but we can live with it. I think it might be worth having a prompter that takes a boolean and then knows whether to prompt or not, eg.
```rust
let prompt = Prompter::new(options.confirm);
if prompt.confirm("Are you sure?") {
...
}
```

Could be a nice re-usable component if we use this pattern a lot.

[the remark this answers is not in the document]
slack-coder: What do you think about something like this for now?

https://github.com/radicle-dev/heartwood/pull/214/commits/9d8e3e3bfddf178cf4db103a3dcf87ddd21717df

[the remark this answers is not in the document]
cloudhead: no no no, this is very arcane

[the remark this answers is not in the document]
cloudhead: I'd leave it how you have it for now

slack-coder:
@cloudhead all points addressed except for one issue

radicle-cli/src/commands/patch/create.rs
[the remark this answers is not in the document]
cloudhead: Yeah, objects with methods are more common and idiomatic if you need to return something from a function

cloudhead (approved):
LGTM; what's the issue with the failing test?

slack-coder:
> LGTM; what's the issue with the failing test?

Ive been trying to figure this out.  It fails to push the branch to rad. I cant figure how why theres a descrepancy, maybe the git version requires this on the integration test.

!! Pushing HEAD to storage...
== Command failed
fatal: The current branch flux-capacitor-power has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream rad flux-capacitor-power

cloudhead:
And it works outside of the testing environment?

slack-coder:
@cloudhead I've made some changes, please confirm the last two commits.

I'll get cracking on a proper fix for the documentation test dependency on `radicle-remote-helper`

radicle-cli/src/commands/patch/create.rs
```diff
@@ -66,9 +74,11 @@ pub fn run(
...
     let head_oid = head.target().ok_or(anyhow!("invalid HEAD ref; aborting"))?;
     let head_commit = workdir.find_commit(head_oid)?;
-    let head_branch = head
-        .shorthand()
-        .ok_or(anyhow!("cannot create patch from detached head; aborting"))?;
+    let head_branch = if head.is_branch() {
+        git::raw::Branch::wrap(head)
+    } else {
+        anyhow::bail!("cannot create patch from detached head; aborting")
+    };
```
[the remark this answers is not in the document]
slack-coder: Having the branch as a `Branch` instead of a `str` is useful for inspecting it (line 102) and is stricter/more expressive.

slack-coder:
The `rad patch` issue this references has a '--closes' argument not supported in these changes.

It will be added in an upcoming PR.

PR #217, opened 2023-01-10
New noise & socks5 APIs

dr-orlovsky:


dr-orlovsky:
ACK f24b97498560e76d7d384986a6d5f8ee0f93d544

radicle-node/src/wire/transport.rs
```diff
@@ -131,15 +135,15 @@ pub struct Transport<R, S, W, G: Negotiator> {
     /// Backing service instance.
     service: Service<R, S, W, G>,
     /// Worker pool interface.
-    worker: chan::Sender<WorkerReq<G>>,
+    worker: chan::Sender<WorkerReq>,
     /// Used to performs X25519 key exchange.
-    keypair: G,
+    node_sk: cyphernet::crypto::ed25519::PrivateKey,
```
cloudhead, line 140: I don't like this much, why can't we hold a signer like before?

radicle-node/src/client.rs
```diff
@@ -103,10 +104,14 @@ impl<G: crypto::Signer + crypto::Negotiator + Clone + 'static> Runtime<G> {
             rng,
         );
 
+        let node_id = signer.public_key().0;
+        let sig = signer.sign(node_id.as_slice());
+        let auth = Authenticator::new(node_id.into(), sig.0.into());
+
```
cloudhead, line 110: We'll need a nonce too actually!

cloudhead:
Amazing, it works!

radicle-node/src/client.rs
cloudhead: To save a round-trip, maybe the nonce could be the noise session hash :thinking:

PR #221, opened 2023-01-12
github: Workaround test dep on git remote helper

slack-coder:
Workaround a client documentation test dependency on radicle's radicle-remote-helper.

The framework has yet to support using the projects version directly.  A proper fix is non-trivial.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -21,7 +21,9 @@ jobs:
         env:
           RUSTFLAGS: -D warnings
       - name: Run tests
-        run: cargo test --all --verbose --all-features
+        run: |
+          cargo install --locked --path ./radicle-remote-helper
+          cargo test --all --verbose --all-features
 
   docs:
     name: Docs


slack-coder:
Should I create an issue for a good follow up fix for this?

Ideally we'd want one which doesnt require installing the packages version.

cloudhead:
Hmmm let's think about this, I don't think this is the right fix, because it'll fail locally.

slack-coder:
> Hmmm let's think about this, I don't think this is the right fix, because it'll fail locally.

The problem is it will hold back testing documentation tests.  I remember we had a similar problem with testing the old radicle client.

As mentioned above, maybe we create an issue for following up with a fix without holding back testing.

cloudhead:
But this isn't a fix, this will result in test failures locally. I'd suggest instead skipping the test function that would otherwise fail with `#[ignore]` until we find a way to run it.

slack-coder:
We can have the ignored tests whitelisted for testing in the integration tests via:

```
cargo test --package radicle-cli [test names] -- --ignored
```

I recommend it if a real fix isn't trivial, so we can work commands relying on `git-remote-rad` concurrently with it.

cloudhead:
So you mean run the tests in CI, but not locally?

slack-coder:
Yes, as a fallback

cloudhead:
Yeah guess we could do that

PR #222, opened 2023-01-13
httpd: Update `radicle_surf`

xphoniex:


xphoniex:
Screenshot from 2023-01-13 18-31-59

`path` is removed from the tree in the new model (red is new), any reason why? @FintanH @keepsimple1

just spoke with @rudolfs and we still seem to need it and can't just use `name`.

FintanH:
Out of curiosity, how does this work overlap with what @keepsimple1 is doing here? https://github.com/radicle-dev/heartwood/pull/149

The `path` was removed because the caller knows the context. In the implementation I've linked above, `json!` is used so that `httpd` serves the data the client expects.

xphoniex:
> 

Ah, I see. I missed the new changes in #149 at the first glance.

PR #226, opened 2023-01-17
cli: add `rad patch show`

slack-coder:
Show patch information with `rad patch show`.

References #148

### DIFF
--- a/radicle-cli/src/commands/patch.rs
+++ b/radicle-cli/src/commands/patch.rs
@@ -8,7 +8,6 @@ mod list;
 mod show;
 
 use std::ffi::OsString;
-use std::str::FromStr;
 
 use anyhow::anyhow;
 
@@ -94,14 +93,6 @@ pub struct Options {
     pub verbose: bool,
 }
 
-fn parse_patch_id(val: OsString) -> Result<OptPatch, anyhow::Error> {
-    let val = val
-        .to_str()
-        .ok_or_else(|| anyhow!("patch id specified is not UTF-8"))?;
-    let patch_id = PatchId::from_str(val).map_err(|_| anyhow!("invalid patch id '{}'", val))?;
-    Ok(OptPatch::Patch(patch_id))
-}
-
 impl Args for Options {
     fn from_args(args: Vec<OsString>) -> anyhow::Result<(Self, Vec<OsString>)> {
         use lexopt::prelude::*;
@@ -166,10 +157,10 @@ impl Args for Options {
                     unknown => anyhow::bail!("unknown operation '{}'", unknown),
                 },
                 Value(val) if op == Some(OperationName::Show) && patch_id == OptPatch::Any => {
-                    patch_id = parse_patch_id(val)?;
+                    patch_id = OptPatch::Patch(term::cob::parse_patch_id(val)?);
                 }
                 Value(val) if op == Some(OperationName::Update) && patch_id == OptPatch::Any => {
-                    patch_id = parse_patch_id(val)?;
+                    patch_id = OptPatch::Patch(term::cob::parse_patch_id(val)?);
                 }
                 _ => return Err(anyhow::anyhow!(arg.unexpected())),
             }
--- a/radicle-cli/src/commands/patch.rs
+++ b/radicle-cli/src/commands/patch.rs
@@ -4,6 +4,8 @@ mod common;
 mod create;
 #[path = "patch/list.rs"]
 mod list;
+#[path = "patch/show.rs"]
+mod show;
 
 use std::ffi::OsString;
 use std::str::FromStr;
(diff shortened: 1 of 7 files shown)


cloudhead:
Nice, this is missing the patch diff though :)

cloudhead:
I think it should show all commits + diff, basically I'd just use the output of `git log --patch` I think.

radicle-cli/src/commands/patch.rs
```diff
@@ -50,9 +52,20 @@ pub enum OptPatch {
...
 
+impl OptPatch {
+    #[inline]
+    pub fn to_option(self) -> Option<PatchId> {
+        match self {
+            OptPatch::Patch(patch_id) => Some(patch_id),
+            _ => None,
+        }
+    }
+}
```
FintanH, line 63:
```suggestion
impl From<OptPatch> for Option<PatchId> {
    #[inline]
    fn from(opt) -> Self {
        match opt {
            OptPatch::Patch(patch_id) => Some(patch_id),
            _ => None,
        }
    }
}
```

slack-coder:
> I think it should show all commits + diff, basically I'd just use the output of `git log --patch` I think.

done :+1:

radicle-cli/src/commands/patch.rs
```diff
@@ -50,9 +52,19 @@ pub enum OptPatch {
     Patch(PatchId),
 }
```
cloudhead, line 53: Maybe it's time now to get rid of this type, and just use `Option<PatchId>` (I mentioned this in the previous PR)

radicle-cli/src/commands/patch.rs
```diff
@@ -79,6 +94,14 @@ pub struct Options {
     pub verbose: bool,
 }
 
+fn parse_patch_id(val: OsString) -> Result<OptPatch, anyhow::Error> {
+    let val = val
+        .to_str()
+        .ok_or_else(|| anyhow!("patch id specified is not UTF-8"))?;
+    let patch_id = PatchId::from_str(val).map_err(|_| anyhow!("invalid patch id '{}'", val))?;
+    Ok(OptPatch::Patch(patch_id))
+}
```
cloudhead, line 103: It seems like we could move this function and the issue-id parsing one into a `terminal::parse` module or something, and combine the functionality into a single function.

radicle-cli/src/commands/patch.rs
slack-coder: This is best done with refactoring patch::create::run().  The function is used to create or update a patch, and the action depends on this value.

We should change OptPatch -> Option<PatchID>, but also split patch::create::run() into a 'create' and 'update' function.

cloudhead: Hmm yeah I think that makes sense! In that case let's leave it for the next PR :stuck_out_tongue:

radicle-cli/src/commands/patch.rs
slack-coder: Yes and likewise other ID parsing commands so the messages are consistent.

Would we be happy in putting it in `terminal` until the context is clearer?  Its still only 174 lines long.  It may be things in `terminal` are better organized by domain then function, for instance we have `terminal::patch` already.

radicle-cli/src/commands/patch.rs
slack-coder: It looks to deserve its own PR IMHO.

radicle-cli/src/commands/patch.rs
cloudhead: I think it would maybe be `terminal::cob` then, since it applies to all cobs.

radicle-cli/src/terminal/cob.rs
```diff
@@ -0,0 +1,15 @@
...
+use anyhow::anyhow;
+
+pub fn parse_patch_id(val: OsString) -> Result<patch::PatchId, anyhow::Error> {
+    let val = val
+        .to_str()
+        .ok_or_else(|| anyhow!("patch id specified is not UTF-8"))?;
+    let patch_id =
+        patch::PatchId::from_str(val).map_err(|_| anyhow!("invalid patch id '{}'", val))?;
+    Ok(patch_id)
+}
```
FintanH, line 15: I think that `PatchId` is just a `type` alias for `ObjectId`. So this could be used for any `cob`, in theory. I wonder if it would make sense to just call this `parse_identifier` and use it for `PatchId`, `IssueId`, etc.

FintanH (approved):
lgtm :)

radicle-cli/src/terminal/cob.rs
slack-coder: You're right about the type alias, however the error message is different.

cloudhead: Yeah I think we'll want to do something like `cob::parse_id(&id, "patch")` and `cob::parse_id(&id, "issue")`

cloudhead (approved):

PR #227, opened 2023-01-17
Upload macOS binaries for e2e testing

adaszko:
This is to facilitate testing on dev’s machines.

### DIFF
--- a/.github/workflows/frontend-testing.yml
+++ b/.github/workflows/frontend-testing.yml
@@ -5,7 +5,7 @@ on:
     branches: [master]
 
 jobs:
-  release-x86_64-linux-binaries:
+  release-gcloud-x86_64-linux-binaries:
     permissions:
       contents: 'read'
       id-token: 'write'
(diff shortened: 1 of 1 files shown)


.github/workflows/frontend-testing.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches: [master]
 
 jobs:
-  release-x86_64-linux-binaries:
+  upload-x86_64-linux-binaries:
```
cloudhead, line 8: Why are we calling these `upload-*`? "release" seemed more correct, since we're also building.

adaszko: I did not want those to be confused with creating GitHub releases

cloudhead: Sure, but it's creating a new build/release. I guess if you're worried about confusion with github release, you could call it `release-gcloud-*`

adaszko: Renamed accordingly

PR #228, opened 2023-01-17
radicle: simplify issue commenting

slack-coder:
Make commenting on an issue easier by offering a default option when 'reply_to' is None.  The default is set to the thread root comment `OpId(issue.author)`.

### DIFF
--- a/radicle/src/cob/issue.rs
+++ b/radicle/src/cob/issue.rs
@@ -289,12 +289,19 @@ impl<'a, 'g> IssueMut<'a, 'g> {
     }
 
     /// Comment on an issue.
+    ///
+    /// By default, comments are replies to the root of the thread.
     pub fn comment<G: Signer, S: ToString>(
         &mut self,
         body: S,
-        reply_to: CommentId,
+        reply_to: Option<CommentId>,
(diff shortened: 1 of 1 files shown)


cloudhead:
I don't get how this is an improvement..

slack-coder:
I don't get how issue comments are supposed to work.  Can we try making them more intuitive?

When I am generally commenting on this issue on Github, I reply on the issue, not a comment.  Without this change it appears I need to find a comment (the root?) and guess the author for 'OpId:root()'.  Is this the correct interpretation?

cloudhead:
I'm not sure if you're talking about the CRDT or the `IssueMut::comment` method, but all comments must have a parent (reply_to) except the root comment. If you want to post a top-level comment, you'd set `reply_to` to the root comment op-id.

cloudhead:
Ok gotcha. The `OpId` should not be constructed manually. You should be able to get the root comment with `issue.root()`. It will return the OpId, then use that OpId as `reply_to`.

cloudhead:
Closing this. If there's still a feeling that the API is not adequate, we can discuss potential changes in an issue maybe.

PR #230, opened 2023-01-18
httpd: Update axum from 0.5.x to 0.6.x

xphoniex:


radicle-httpd/src/api/v1/projects.rs
```diff
@@ -45,9 +45,11 @@ pub fn router(ctx: Context) -> Router {
                 )),
             ),
         )
+        .route("/projects/:project/tree/:sha/", get(tree_handler_root))
```
xphoniex, line 48: In 0.5, `/projects/:project/tree/:sha/*path` would also match when hitting `/projects/rad:.../tree/01.../` but not in 0.6. You either have to add a separate handler for the root version or use `route_with_tsr`

radicle-httpd/src/lib.rs
```diff
@@ -98,9 +98,9 @@ async fn git_handler(
     AxumPath((project, request)): AxumPath<(String, String)>,
     method: Method,
     headers: HeaderMap,
-    body: Bytes,
     ConnectInfo(remote): ConnectInfo<SocketAddr>,
     query: RawQuery,
+    body: Bytes,
```
xphoniex, line 103: The extractor which consumes the body should be last in new axum, should work just fine, but maybe we should add a simple test here, just to be sure?

PR #232, opened 2023-01-18
node: Get replication tests passing

### LINKED ISSUE #13: Non-blocking `git-fetch` replication
Currently, `git-fetch` runs within the service thread, which means it blocks the service from processing other requests.

We want a worker-thread architecture where one worker is assigned to each project, and `fetch` jobs are sent to these threads. The service can then be woken up when the worker is done with the work.

*Implementation details*

* Use crossbeam-channel to communicate between threads
* Use the Reactor's `Waker` to wake-up the service, which will then read from its worker channel
* Having one channel per project is not a good idea, since we could have 10K+ projects
* Worker thread pool should be started up from `client` or `main`, and be handed a waker and client handle
* An external thread-pool dependency is probably overkill for this

cloudhead:
*See commits.*

Closes #13

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -699,11 +699,12 @@ dependencies = [
 
 [[package]]
 name = "dialoguer"
-version = "0.10.2"
+version = "0.10.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


dr-orlovsky (approved):
ACK 36675e125de5c7ae7f0fbeb2ef08cb7b26b03408

cloudhead:
Merged.

PR #236, opened 2023-01-19
httpd: Change JSON schema

xphoniex:


radicle-httpd/src/api/json.rs
```diff
@@ -36,20 +37,21 @@ pub(crate) fn blob(blob: &Blob, path: &str) -> serde_json::Value {
...
+pub(crate) fn tree(tree: &Tree, path: &str, stats: &Stats) -> Value {
+    let prefix = Path::new(path);
     let entries = tree
         .entries()
         .iter()
         .map(|entry| {
             json!({
                 "path": prefix.join(entry.name()),
                 "name": entry.name(),
-                "lastCommit": serde_json::Value::Null,
```
cloudhead, line 48: I think this is supposed to be the whole commit, not just the id!

cloudhead:
Looks great otherwise.

radicle-httpd/src/api/json.rs
xphoniex: hm... looks like in interface, `lastCommit` is retrieve from `blob` end point and is not being used anywhere else, I think we should remove it.

Issue #187, opened 2022-12-28
radicle-cli: Port `rad comment`


sebastinez:
We should port `rad comment` from https://github.com/radicle-dev/radicle-cli/tree/master/comment

slack-coder:
The `rad comment` command includes a `--reply-to` argument which defines the comment being replied to.  Comment ID's (OpIDs) have yet to have a String conversion.  How should it be supported?  Or should we work around it with `term::comment_select` as `rad issue` does?

Here's what an OpID looks like:
```
pub struct OpId(Lamport, ActorId);
```

slack-coder:
We also need commenting support for `Patch` as has been done for `Issue`.

slack-coder:
This is almost complete.

How do we want comments to be show?

slack-coder:
@cloudhead this is complete

PR #237, opened 2023-01-19
cli: Support patch commenting

slack-coder:
Allow commenting on a patch as well an issue.

References: #187

### DIFF
--- a/radicle/src/cob/patch.rs
+++ b/radicle/src/cob/patch.rs
@@ -573,13 +573,13 @@ impl store::Transaction<Patch> {
         &mut self,
         revision: RevisionId,
         body: S,
-        reply_to: CommentId,
+        reply_to: Option<CommentId>,
     ) -> OpId {
         self.push(Action::Thread {
             revision,
             action: thread::Action::Comment {
                 body: body.to_string(),
-                reply_to: Some(reply_to),
+                reply_to,
             },
         })
     }
@@ -696,7 +696,7 @@ impl<'a, 'g> PatchMut<'a, 'g> {
         &mut self,
         revision: RevisionId,
         body: S,
-        reply_to: CommentId,
+        reply_to: Option<CommentId>,
         signer: &G,
     ) -> Result<CommentId, Error> {
         self.transaction("Comment", signer, |tx| tx.comment(revision, body, reply_to))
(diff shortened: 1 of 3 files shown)


slack-coder:
@cloudhead what are the expectations with patch's and comments?  Should all revisions have a root comment in their discussion?  If not, which ID should I use for comment()'s reply argument?

```
let (revision_id, revision) = patch.revisions().last().expect("patch has a revision");
let (comment_id, _) = revision.discussion.root().expect("discussion has root");
patch.comment(*revision_id, message, *comment_id, &signer)?;
```

FintanH:
Something I'd like to flag is that there are two notions of commenting with patches. One is for the discussion thread on a particular revision. The other is an inline comment for a review. 

For the former, I think it could make sense to be able to "reply to", whether it's the root or a particular comment.

For the latter, I don't think it's currently possible to reply to a comment. I believe we would need to change `comment: String` to `comments: Thread` :)

Why I bring this up in the first place though is how commenting might be confusing for one versus the other, or whether it should Just Work:tm: on both types and the application figures out how to find the correct `Thread` to reply to.

FintanH:
> > For the latter, I don't think it's currently possible to reply to a comment. I believe we would need to change `comment: String` to `comments: Thread` :)
> 
> What about the 'review' also be a comment on the discussion thread? I certainly agree with the problem. Maybe we should create an issue for follow up?

I think it makes sense for the review thread to be kind of insular as part of the review itself, as opposed to the discussion thread. Ya, an issue makes sense :ok_hand:

cloudhead:
Yup I think we're on the same page.

To answer the question: revisions don't necessarily have a root comment. I can publish a revision without a comment. So the first comment posted would start a discussion. Then, top-level comments should reply to that first comment, or to one of the subsequent comments.

```
rad comment <patch-id> # create root comment for sole revision, or reply to root comment; error if more than one revision
rad comment <patch-id> --revision <revision-index> # same as the above, but on a specific revision
rad comment <patch-id> --revision <revision-index> --reply-to <comment-id> # reply to specific comment
```
This interface will mostly be used by tooling and for testing. Users will be encouraged to use a TUI or the web client for commenting.

slack-coder:
> > > For the latter, I don't think it's currently possible to reply to a comment. I believe we would need to change `comment: String` to `comments: Thread` :)
> > 
> > 
> > What about the 'review' also be a comment on the discussion thread? I certainly agree with the problem. Maybe we should create an issue for follow up?
> 
> I think it makes sense for the review thread to be kind of insular as part of the review itself, as opposed to the discussion thread. Ya, an issue makes sense ok_hand

https://github.com/radicle-dev/heartwood/issues/244

slack-coder:
> I can publish a revision without a comment. So the first comment posted would start a discussion. Then, top-level comments should reply to that first comment, or to one of the subsequent comments.

Is it OK to have multiple 'root' comments?  Comments are added concurrently leading to multiple 'roots'.

slack-coder:
> > 
> 
> My understanding is that the root comment is the first comment made by the person who created the issue.

You're right, though heads up, we had ironed things out recently.  Conceptually Thread comments without a 'reply-to' were common, but they were also misinterpreted as 'roots'.  Therefore Thread.root() was replaced with Thread.first() etc.

radicle/src/cob/patch.rs
```diff
@@ -1042,6 +1042,38 @@ mod test {
...
+        let (revision_id, _) = patch.revisions().last().expect("patch has a revision");
+        assert!(
+            patch
+                .comment(*revision_id, "patch comment", None, &signer)
+                .is_ok(),
+            "can comment on patch"
+        );
+
+        let (_, revision) = patch.revisions().last().expect("patch has a revision");
+        let (_, comment) = revision.discussion.first().expect("comment saved");
```
cloudhead, line 1073: We generally use `unwrap` in tests, as the code is too noisy otherwise, and it's enough to point to the line that panicked.

slack-coder: Noted :+1:

PR #252, opened 2023-01-26
node: Improve logging output

cloudhead:
image


slack-coder:
Maybe another term in place of 'target' like 'source'?  Its describing the origin of the log statement and not where its destined.

Otherwise :+1:

cloudhead:
> Maybe another term in place of 'target' like 'source'? Its describing the origin of the log statement and not where its destined.
> 
> Otherwise +1

`target` is part of the `log` crate's API, it can't be changed.

https://docs.rs/log/latest/log/

slack-coder:
> > 
> 
> `target` is part of the `log` crate's API, it can't be changed.
> 
> https://docs.rs/log/latest/log/

lol nevermind then

Issue #245, opened 2023-01-23
String format for radicle::cob::op::OpId


slack-coder:
A heartwood `Comment` uses the OpId type to uniquely identify it, and it may be used in other places.  This ID type needs a string format to allow referencing it to users.  For example, replying to a comment via command line argument.

How should the be formatted and parsed?

```
/// Identifies an [`Op`] internally and within the change graph.
#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
pub struct OpId(Lamport, ActorId);
```

cloudhead:
I was thinking we could do `<actor-id> "/" <lamport>`, so for example `z6MksFqXN3Yhqk8pTJdUGLwATkRfQvwZXPqR2qMEhbS9wzpT/12`

slack-coder:
> I was thinking we could do `<actor-id> "/" <lamport>`, so for example `z6MksFqXN3Yhqk8pTJdUGLwATkRfQvwZXPqR2qMEhbS9wzpT/12`

:+1:  implemented so we have something to work with.  Lets keep this issue open until we settle on this format or decide on another.

PR #255, opened 2023-01-27
Add Op ID string formatting

slack-coder:
To help users, create an Op ID string format of the form '[cob::op::ActorId]/[cob::op::Lamport]'.

The formatting is not final and is intended to support development around interfacing with comments.

References: #245

### DIFF
--- a/radicle-crdt/src/clock.rs
+++ b/radicle-crdt/src/clock.rs
@@ -1,8 +1,11 @@
+use std::fmt;
+use std::str::FromStr;
 use std::time::SystemTime;
 use std::time::UNIX_EPOCH;
 
 use num_traits::Bounded;
 use serde::{Deserialize, Serialize};
+use thiserror::Error;
 
 use crate::ord::Max;
 use crate::Semilattice as _;
@@ -45,11 +48,47 @@ impl Lamport {
     }
 }
 
+impl fmt::Display for Lamport {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.get())
+    }
+}
+
 impl From<u64> for Lamport {
     fn from(counter: u64) -> Self {
         Self {
             counter: Max::from(counter),
         }
+    }
+}
+
+impl From<Lamport> for u64 {
+    fn from(arg: Lamport) -> Self {
+        arg.get()
+    }
+}
+
+/// Error decoding an operation from an entry.
+#[derive(Error, Debug)]
+pub enum LamportError {
+    #[error("invalid lamport clock value")]
+    Invalid,
+}
+
+impl FromStr for Lamport {
+    type Err = LamportError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::try_from(s)
+    }
+}
+
+impl TryFrom<&str> for Lamport {
+    type Error = LamportError;
+
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        let v = s.parse::<u64>().map_err(|_| LamportError::Invalid)?;
+        Ok(v.into())
     }
 }
 
--- a/radicle/src/cob/op.rs
+++ b/radicle/src/cob/op.rs
@@ -1,4 +1,7 @@
 use std::collections::BTreeMap;
+use std::fmt;
+use std::str;
+use std::str::FromStr;
 
 use nonempty::NonEmpty;
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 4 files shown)


radicle/src/cob/op.rs
```diff
@@ -34,6 +37,42 @@ impl OpId {
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::try_from(s)
+    }
+}
+
+impl TryFrom<&str> for OpId {
+    type Error = OpIdError;
+
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        let strs: Vec<_> = s.split('/').collect();
```
cloudhead, line 65: `split_once` works better for this

radicle/src/cob/op.rs
```diff
@@ -34,6 +37,42 @@ impl OpId {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}/{}", self.1, self.0)
+    }
+}
+
+/// Error decoding an operation from an entry.
+#[derive(Error, Debug)]
+pub enum OpIdError {
+    #[error("invalid Op ID")]
+    Invalid(),
```
cloudhead, line 50:
```suggestion
    Invalid,
```

radicle-crdt/src/clock.rs
```diff
@@ -53,6 +62,36 @@ impl From<u64> for Lamport {
...
+}
+
+impl TryFrom<&str> for Lamport {
+    type Error = LamportError;
+
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        let v = s.parse::<u64>().map_err(|_| LamportError::Invalid())?;
+        Ok(v.into())
+    }
+}
```
cloudhead, line 93: It's a bit redundant to offer `TryFrom` when you have `FromStr` already, we don't normally do that.

radicle-crdt/src/clock.rs
```diff
@@ -53,6 +62,36 @@ impl From<u64> for Lamport {
...
+    fn from(arg: Lamport) -> Self {
+        arg.get()
+    }
+}
+
+/// Error decoding an operation from an entry.
+#[derive(Error, Debug)]
+pub enum LamportError {
+    #[error("invalid Lamport Clock value")]
+    Invalid(),
```
FintanH, line 75: It might be useful to carry the parse error inside here and use the `#[source]` pragma

radicle-crdt/src/clock.rs
FintanH: I've definitely used that before. Gives some flexibility to do `Lamport::try_from` or `parse`.

radicle/src/cob/op.rs
```diff
@@ -34,6 +37,42 @@ impl OpId {
...
+impl fmt::Display for OpId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}/{}", self.1, self.0)
+    }
+}
+
+/// Error decoding an operation from an entry.
+#[derive(Error, Debug)]
+pub enum OpIdError {
+    #[error("invalid Op ID")]
```
FintanH, line 49: It's quite annoying when errors just say that it's invalid. Can we add the different variants of why it's gone wrong? i.e. missing '/', incorrect Lamport clock, and incorrect ActorId.

radicle-crdt/src/clock.rs
slack-coder: Indeed git::Oid was the inspiration.  Its not redundant in use, it makes try_from more complete and supports FromStr because its what people are used to.

radicle/src/cob/op.rs
slack-coder: The difference here is are the errors for developers or users?

If a user inputs a bad Op ID its because they copy+pasted it wrong.  The error would be redundant at this point.

radicle-crdt/src/clock.rs
slack-coder: How helpful would it be?  It feels redundant as Lamport Clocks don't have an underlying structure like SSH Keys or otherwise.

radicle/src/cob/op.rs
FintanH: I would say both and I'd argue that a lot of users are "developers" in the general sense so it will help them also diagnose when things go wrong.

Say they paste something they thought was an Op ID, which do you think is more useful:
1) "invalid Op ID"
2) "expected an Op ID of the form '<actor id>/<clock>', but the '<actor id>' was malformed"

I know which one I would like to see :)

radicle-crdt/src/clock.rs
FintanH: It's useful to know what went wrong when trying to parse -- this time it's really for the developers of this code base. The `#[source]` will provide the error that occurred when parsing the `u64` -- which is arguably the underlying structure of the clock.

radicle/src/cob/op.rs
slack-coder: I would prefer to see the first in logging, the later comes across as verbose.  I only see information on how the formatting should be being useful if they are handcrafted, or if they are empty.

FintanH: At the very least I would report what the string was then. In my mind, I'm comparing this to getting back `File error` and not knowing which file name had an error.

slack-coder: I've updated the error to distinguish between an empty string and a badly formatted Op ID string

FintanH: You can likely google "useful error messages" and I'd say you'd find similar advice across any articles. One I read a while back (would need to revisit) is [this post by Tikhon Jelvis](https://jelv.is/blog/Structure-your-Errors/). I'm of the opinion that more information is better when it comes to errors -- you can filter information presented to you but it's harder to _add_ information.

Feel free to proceed anyway. I'll take another look at the PR.

radicle/src/cob/op.rs
```diff
@@ -182,3 +226,16 @@ impl<G: Signer, A: Clone> Actor<G, A> {
...
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_opid_try_from_str() {
+        let str = "z6MksFqXN3Yhqk8pTJdUGLwATkRfQvwZXPqR2qMEhbS9wzpT/12";
+
+        let id = OpId::try_from(str).expect("Op ID parses string");
+        assert_eq!(str, id.to_string(), "string conversion is consistent");
+    }
```
FintanH, line 240: Would make sense to test the unhappy paths too :)

FintanH (approved):

radicle/src/cob/op.rs
slack-coder: Thanks, Ill check it out.  I find recommendations better because the signal to noise ratio is too high.

radicle/src/cob/op.rs
slack-coder: Hmmm on it

slack-coder: https://github.com/radicle-dev/heartwood/pull/255/commits/65f58353218fa571b36263dd106a96656f0c153c

PR #259, opened 2023-01-30
node: Fix some replication bugs

cloudhead:


radicle-node/src/service.rs
```diff
@@ -394,36 +394,44 @@ where
 
         match cmd {
             Command::Connect(id, addr) => self.reactor.connect(id, addr),
-            Command::Fetch(id, resp) => {
+            Command::Fetch(rid, resp) => {
                 if !self
                     .tracking
-                    .is_repo_tracked(&id)
+                    .is_repo_tracked(&rid)
```
slack-coder, line 400: If adding context is the goal, I'm unsure if the rename is much more helpful.  The documentation says its a project ID, maybe 'project_id' would be better? 'rid' sounds like 'repo id' or 'remote id'.

```
    /// Fetch the given project from the network.
    Fetch(Id, chan::Sender<FetchLookup>),
```

radicle/src/storage/git.rs
```diff
@@ -143,20 +143,26 @@ impl Storage {
         self.path.as_path()
     }
 
-    pub fn projects(&self) -> Result<Vec<Id>, Error> {
-        let mut projects = Vec::new();
+    pub fn repositories(&self) -> Result<Vec<Id>, Error> {
+        let mut repos = Vec::new();
```
slack-coder, line 145: Just curious, why the change from project -> repository?

radicle-node/src/service.rs
cloudhead: RID stands for repo id indeed, I'm slowly renaming stuff as I go.

radicle/src/storage/git.rs
cloudhead: Project is incorrect/overly specific. "Project" is a term for repositories that have a project payload. But here we're just returning all repositories under storage.

radicle/src/git.rs
```diff
@@ -507,6 +507,16 @@ pub mod url {
...
 }
 
+/// Git environment variables.
+pub mod env {
+    /// Set of environment vars to reset git's configuration to default.
+    pub const GIT_RESET: [(&str, &str); 3] = [
+        ("GIT_CONFIG", "/dev/null"),
+        ("GIT_CONFIG_GLOBAL", "/dev/null"),
+        ("GIT_CONFIG_NOSYSTEM", "1"),
+    ];
```
slack-coder, line 517: Maybe GIT_VOID_CONFIG?  Maybe a term which better fits what they are instead of what they're used for.

radicle-node/src/main.rs
```diff
@@ -14,6 +14,7 @@ use radicle_node::{logger, service};
 struct Options {
     connect: Vec<(NodeId, Address)>,
     external_addresses: Vec<Address>,
+    daemon: Option<net::SocketAddr>,
```
slack-coder, line 17: Where is the daemon socket being exposed?  I believed git communication was happening over radicle's connection via 8776 (might be wrong port number), but this suggests a return of the `git url` parameter to the node.

radicle-node/src/service.rs
slack-coder: Cool, looks like the documentation would be updated with this change

radicle-node/src/service.rs
```diff
@@ -512,53 +512,56 @@ where
...
+                        rid,
+                        initiated,
+                        remote,
+                        namespaces,
+                        result,
+                    })
+                    .is_err()
+                {
+                    log::error!(target: "service", "Error sending fetch result for {rid}..");
+                    self.fetch_reqs.remove(&rid);
```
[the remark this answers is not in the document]
cloudhead: Send errors are always due to the channel being disconnected, so not really worth it.

PR #260, opened 2023-01-30
radicle-httpd: Add issue endpoints

sebastinez:
WIP


radicle-httpd/src/api/v1/projects.rs
```diff
@@ -47,8 +48,14 @@ pub fn router(ctx: Context) -> Router {
         .route("/projects/:project/remotes/:peer", get(remote_handler))
         .route("/projects/:project/blob/:sha/*path", get(blob_handler))
         .route("/projects/:project/readme/:sha", get(readme_handler))
-        .route("/projects/:project/issues", get(issues_handler))
-        .route("/projects/:project/issues/:id", get(issue_handler))
+        .route(
+            "/projects/:project/issues",
+            post(issue_create_handler).get(issues_handler),
+        )
```
xphoniex, line 54: any reason why this is name `issue_create_handler` and not `create_issue_handler`?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -399,6 +407,89 @@ async fn issues_handler(
...
+    let storage = &ctx.profile.storage;
+    let signer = ctx.profile.signer().unwrap();
+    let repo = storage.repository(project)?;
+    let mut issues = Issues::open(ctx.profile.public_key, &repo)?;
+    issues
+        .create(issue.title, issue.description, &issue.tags, &signer)
+        .map_err(Error::from)?;
+
+    Ok::<_, Error>(())
+}
```
xphoniex, line 436: maybe best if you return a `success: true` here instead of empty response?

radicle-httpd/src/api/v1/projects.rs
sebastinez: Hey @xphoniex, no specific reason, I think I wanted to keep starting with the corresponding issue like `project...` `remotes...` etc.

radicle-httpd/src/api/v1/projects.rs
sebastinez: This was talked about with @cloudhead, since we don't require any additional information at this point, and we can check for an OK response for success.
Any reason what benefit we would have with the `success: true` body?

xphoniex: returning empty response would mean any empty response in the future / other routes can be counted as a success as well, which might not hold true, e.g. if someone else implements it and is not aware of this.

we don't lose much by returning a `success: true`, it's just standard.

sebastinez: Ok makes sense, will add it 👍  thanks!

radicle-httpd/src/api/v1/projects.rs
xphoniex: nitpick: that's what I was assuming, maybe go with `issue_creation_handler`?

sebastinez: yeah sounds good 👍

sebastinez: Done in c9834bd

radicle-httpd/src/api/v1/projects.rs
sebastinez: Done in c9834bd

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -399,6 +407,89 @@ async fn issues_handler(
...
+                }
+                thread::Action::Redact { id } => {
+                    actor.redact(id);
+                }
+            }
+        }
+    };
+
+    Ok::<_, Error>(())
+}
```
xphoniex, line 491: here as well

sebastinez: ok done 👍

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -399,6 +407,89 @@ async fn issues_handler(
...
+    let repo = storage.repository(project)?;
+    let mut issues = Issues::open(ctx.profile.public_key, &repo)?;
+    issues
+        .create(issue.title, issue.description, &issue.tags, &signer)
+        .map_err(Error::from)?;
+
+    Ok::<_, Error>(Json(json!({ "success": "true" })))
+}
+
+/// Update an issue
```
xphoniex, line 438: /// Update an issue.

xphoniex (changes requested):

xphoniex:
otherwise lgtm!

xphoniex (approved):

radicle-httpd/src/api/v1/projects.rs
sebastinez: Done in 78ae1ba

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -399,6 +407,89 @@ async fn issues_handler(
...
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub struct IssueCreate {
+    pub title: String,
+    pub description: String,
+    pub tags: Vec<Tag>,
+}
+
+/// Create a new issue
```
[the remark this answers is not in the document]
sebastinez: Done in 78ae1ba

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -399,6 +407,89 @@ async fn issues_handler(
...
+#[derive(Debug, Deserialize, Serialize)]
+pub struct IssueCreate {
+    pub title: String,
+    pub description: String,
+    pub tags: Vec<Tag>,
+}
+
+/// Create a new issue.
+/// `POST /projects/:project/issues`
+async fn issue_creation_handler(
```
cloudhead, line 419:
```suggestion
async fn issue_create_handler(
```

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -399,6 +407,89 @@ async fn issues_handler(
...
+    State(ctx): State<Context>,
+    AuthBearer(token): AuthBearer,
+    Path(project): Path<Id>,
+    Json(issue): Json<IssueCreate>,
+) -> impl IntoResponse {
+    let sessions = ctx.sessions.write().await;
```
cloudhead, line 425:
```suggestion
    let sessions = ctx.sessions.read().await;
```
Should be enough? Is the presence of that token automatically mean it's authorized? Don't remember what we landed on.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -399,6 +407,89 @@ async fn issues_handler(
...
+    sessions.get(&token).ok_or(Error::Auth("Unauthorized"))?;
+    let storage = &ctx.profile.storage;
+    let signer = ctx.profile.signer().unwrap();
```
cloudhead, line 428: I'd return an "error unauthorized" here, since it means the user key is not in ssh agent.

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -399,6 +407,89 @@ async fn issues_handler(
...
+    let repo = storage.repository(project)?;
+    let mut issues = Issues::open(ctx.profile.public_key, &repo)?;
+    issues
+        .create(issue.title, issue.description, &issue.tags, &signer)
+        .map_err(Error::from)?;
+
+    Ok::<_, Error>(Json(json!({ "success": "true" })))
```
cloudhead, line 435:
```suggestion
    Ok::<_, Error>(Json(json!({ "success": true })))
```
And we should return `201 Created`

radicle-httpd/src/api/v1/projects.rs
xphoniex: just saw this

Screenshot from 2023-02-03 11-41-58

we already have a `tree_handler_root` so maybe best if we stick to it, and use `issue_handler_create` here?

radicle-httpd/src/api/v1/projects.rs
sebastinez: Yeah that's right, will change this.

radicle-httpd/src/api/v1/projects.rs
sebastinez: I'm fine with both of them, but I think cloudheads suggestion sounds more natural in this case.

radicle-httpd/src/api/v1/projects.rs
sebastinez: done in 4cde666

radicle-httpd/src/api/v1/projects.rs
sebastinez: done in 4cde666

radicle-httpd/src/api/v1/projects.rs
sebastinez: done in 4cde666

radicle-httpd/src/api/v1/projects.rs
sebastinez: kept @cloudhead suggestions, as it sounds more natural in 4cde666

cloudhead (approved):

PR #262, opened 2023-02-02
radicle-httpd: Add `GET /api/v1/sessions/:id` route to get session information

sebastinez:
We currently don't have a way on for the web app to check the validity of a session_id, so we need a route to do that.

Also I didn't want to limit OK responses just to authorized sessions, but also return a OK response with the Unauthorized session information, which can become handy.

### DIFF
--- a/radicle-httpd/src/api/auth.rs
+++ b/radicle-httpd/src/api/auth.rs
@@ -1,27 +1,21 @@
 use radicle::crypto::PublicKey;
-use serde::{Serialize, Serializer};
-use time::OffsetDateTime;
+use serde::{Deserialize, Serialize};
+use time::{serde::timestamp, OffsetDateTime};
 
-#[derive(Clone, PartialEq, PartialOrd)]
-pub struct DateTime(pub OffsetDateTime);
-
-impl Serialize for DateTime {
-    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        serializer.serialize_str(&format!("{}", self.0))
-    }
-}
-
-#[derive(Clone, Serialize, PartialEq)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 #[serde(rename_all = "lowercase")]
 pub enum AuthState {
     Authorized,
     Unauthorized,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct Session {
     pub status: AuthState,
     pub public_key: PublicKey,
-    pub issued_at: DateTime,
-    pub expires_at: DateTime,
+    #[serde(with = "timestamp")]
+    pub issued_at: OffsetDateTime,
+    #[serde(with = "timestamp")]
+    pub expires_at: OffsetDateTime,
 }
--- a/radicle-httpd/src/api/auth.rs
+++ b/radicle-httpd/src/api/auth.rs
@@ -11,25 +11,17 @@ impl Serialize for DateTime {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Serialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
 pub enum AuthState {
-    Authorized(Session),
-    Unauthorized(Session),
+    Authorized,
+    Unauthorized,
 }
 
 #[derive(Clone)]
 pub struct Session {
-    pub status: String,
+    pub status: AuthState,
     pub public_key: PublicKey,
     pub issued_at: DateTime,
     pub expires_at: DateTime,
-}
-
-impl From<AuthState> for Session {
-    fn from(other: AuthState) -> Self {
-        match other {
-            AuthState::Authorized(s) => s,
-            AuthState::Unauthorized(s) => s,
-        }
-    }
 }
--- a/radicle-httpd/src/api/auth.rs
+++ b/radicle-httpd/src/api/auth.rs
@@ -11,6 +11,8 @@ impl Serialize for DateTime {
     }
 }
 
+#[derive(Clone, Serialize)]
+#[serde(tag = "type", rename_all = "camelCase")]
 pub enum AuthState {
     Authorized(Session),
     Unauthorized {
--- a/radicle-httpd/src/api/auth.rs
+++ b/radicle-httpd/src/api/auth.rs
@@ -11,21 +11,25 @@ impl Serialize for DateTime {
     }
 }
 
-#[derive(Clone, Serialize)]
-#[serde(tag = "type", rename_all = "camelCase")]
+#[derive(Clone)]
 pub enum AuthState {
     Authorized(Session),
-    Unauthorized {
-        public_key: PublicKey,
-        expires_at: DateTime,
-    },
+    Unauthorized(Session),
 }
 
-// We copy the implementation of siwe::Message here to derive Serialization and Debug
-#[derive(Clone, Serialize)]
-#[serde(rename_all = "camelCase")]
+#[derive(Clone)]
 pub struct Session {
+    pub status: String,
     pub public_key: PublicKey,
     pub issued_at: DateTime,
     pub expires_at: DateTime,
+}
+
+impl From<AuthState> for Session {
+    fn from(other: AuthState) -> Self {
+        match other {
+            AuthState::Authorized(s) => s,
+            AuthState::Unauthorized(s) => s,
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


xphoniex:
> ... but also return a OK response with the Unauthorized session information, which can become handy.

did you make sure this is not leaking sensitive info?

otherwise lgtm, please squash the commits

sebastinez:
> > ... but also return a OK response with the Unauthorized session information, which can become handy.
> 
> did you make sure this is not leaking sensitive info?
> 
> otherwise lgtm, please squash the commits

We just return timestamps and the session_id and the public_key of the session.

sebastinez:
Squashed commits

xphoniex (approved):

FintanH:
> > > ... but also return a OK response with the Unauthorized session information, which can become handy.
> > 
> > 
> > did you make sure this is not leaking sensitive info?
> > otherwise lgtm, please squash the commits
> 
> We just return timestamps and the session_id and the public_key of the session.

It might make sense to not derive `Serialize` if this is a worry. If sensitive info is added then it will automatically get added to the serialized output.

sebastinez:
> It might make sense to not derive `Serialize` if this is a worry. If sensitive info is added then it will automatically get added to the serialized output.

Thanks @FintanH for the catch, I reworked the PR a bit in commit 50fcc7d wrt this

radicle-httpd/src/api/auth.rs
```diff
@@ -11,19 +11,25 @@ impl Serialize for DateTime {
...
-    },
+    Unauthorized(Session),
 }
 
-// We copy the implementation of siwe::Message here to derive Serialization and Debug
-#[derive(Clone, Serialize)]
-#[serde(rename_all = "camelCase")]
+#[derive(Clone)]
 pub struct Session {
+    pub status: String,
```
FintanH, line 22: Not a blocker on my end, but I'm a bit confused why `status` is carried within the `Session`. Isn't that encoded via `AuthState`?

FintanH (approved):

radicle-httpd/src/api/auth.rs
sebastinez: Yeah, that's the reason I wanted the serializer add it to the json ouput as a `derive tag`operation.

Will see if I find a better way to handle it.

sebastinez: Ok I removed the `AuthState` enum wrapper around `Session`, which seemed to make less sense without SIWE.
Seems nicer altogether
03dae8a

radicle-httpd/src/api/v1/sessions.rs
```diff
@@ -82,13 +103,14 @@ async fn session_signin_handler(
...
             public_key: request.pk,
-            issued_at: DateTime(OffsetDateTime::now_utc()),
+            issued_at: session.issued_at.to_owned(),
             expires_at: DateTime(expiration_time),
         };
-        sessions.insert(session_id.clone(), AuthState::Authorized(session));
+        sessions.insert(session_id.clone(), session);
 
-        return Ok::<_, Error>(());
+        return Ok::<_, Error>(Json(json!({ "success": "true" })));
```
cloudhead, line 113: What's the idea here, isn't `200 OK` sufficient?

radicle-httpd/src/api/v1/sessions.rs
```diff
@@ -41,16 +44,38 @@ async fn session_create_handler(State(ctx): State<Context>) -> impl IntoResponse
...
     let mut sessions = ctx.sessions.write().await;
-    sessions.insert(session_id.clone(), auth_state);
+    sessions.insert(session_id.clone(), session.clone());
+
+    Ok::<_, Error>(Json(json!({
+        "sessionId": session_id,
+        "publicKey": session.public_key,
+        "issuedAt": session.issued_at,
+        "expiresAt": session.expires_at
+    })))
```
cloudhead, line 61: Can we actually return `201 Created` for all POSTs?

radicle-httpd/src/api/v1/sessions.rs
```diff
@@ -131,5 +153,8 @@ mod routes {
         .await;
 
         assert_eq!(response.status(), StatusCode::OK);
+
+        let response = get(&app, format!("/sessions/{}", session_info.session_id)).await;
+        assert_eq!(response.status(), StatusCode::OK);
```
cloudhead, line 158: An improvement on the test would be to GET before the PUT, check that it's unauthorized, then GET after the PUT, and check that it's authoriezed.

radicle-httpd/src/api/v1/sessions.rs
xphoniex: e.g. if you use a cli cmd, and returns nothing, you don't know whether it suceeded or not, there's always `echo $` but actual textual output is nicer.

radicle-httpd/src/api/v1/sessions.rs
```diff
@@ -82,13 +103,14 @@ async fn session_signin_handler(
...
             .unwrap();
         let session = Session {
+            status: AuthState::Authorized,
             public_key: request.pk,
-            issued_at: DateTime(OffsetDateTime::now_utc()),
+            issued_at: session.issued_at.to_owned(),
             expires_at: DateTime(expiration_time),
         };
-        sessions.insert(session_id.clone(), AuthState::Authorized(session));
+        sessions.insert(session_id.clone(), session);
```
cloudhead, line 111: I wonder, could you do a `get_mut()` and then just mutate those fields, instead of having to recreate a session object and inserting?

radicle-httpd/src/api/v1/sessions.rs
cloudhead: Yeah, guess that's useful. Let's use a boolean then, so `{ "success": true }`.

radicle-httpd/src/api/v1/sessions.rs
FintanH, line 61: I'm wondering should all `json!` be placed [here](https://github.com/radicle-dev/heartwood/blob/master/radicle-httpd/src/api/json.rs)? Would help with the duplication.

FintanH (commented):
Nice :)

radicle-httpd/src/api/v1/sessions.rs
sebastinez: Yeah that would work, done in 1150daa4e04ef0195d3e478beb3840fcf5b014fe

radicle-httpd/src/api/v1/sessions.rs
sebastinez: Ok yeah makes sense done in 1150daa4e04ef0195d3e478beb3840fcf5b014fe

radicle-httpd/src/api/v1/sessions.rs
sebastinez: sure thing, see 1150daa4e04ef0195d3e478beb3840fcf5b014fe

radicle-httpd/src/api/v1/sessions.rs
sebastinez: has been done in 6f62e5c

radicle-httpd/src/api/v1/sessions.rs
sebastinez: hmm not a bad idea, not sure it would help with the duplication, since we return similar but not equal objects in the GET and POST route.

sebastinez: Ok did make both json objects equal and extracted it to `api::json`
741252e

radicle-httpd/src/api/v1/sessions.rs
xphoniex: the[ `true` part is string](https://github.com/radicle-dev/heartwood/commit/6f62e5cb93ff288160c9e56e00127a39f49689ab#diff-83176232347eff805dc12f6bbb8553f21f50e79f430bc2d7248c5a93c930d45aR112-R114), want it changed to boolean:
```
"success": "true" -> "success": true
```

xphoniex: you could also extract it and re-use `api::json::success` in both places instead.

rudolfs:
Could we also return the User's ID along with the session params? Would like to shot it in the UI.

radicle-httpd/src/api/v1/sessions.rs
```diff
@@ -61,34 +79,26 @@ async fn session_signin_handler(
...
             return Err(Error::Auth("Session expired"));
         }
         let payload = format!("{}:{}", session_id, request.pk);
         request
             .pk
             .verify(payload.as_bytes(), &request.sig)
             .map_err(Error::from)?;
-        let expiration_time = OffsetDateTime::now_utc()
+        session.status = AuthState::Authorized;
+        session.public_key = request.pk;
```
cloudhead, line 96:
```suggestion
```
It's already equal

cloudhead:
I've taken the liberty to fix the two outstanding issues and rebase.

PR #264, opened 2023-02-02
Properly implement control channel for fetch

cloudhead:


radicle-cli/src/commands/clone.rs
```diff
@@ -133,50 +131,45 @@ pub enum CloneError<H: node::Handle> {
...
-        FetchLookup::Found { seeds, results } => {
-            // TODO: If none of them succeeds, output an error. Otherwise tell the caller
-            // how many succeeded.
-            for result in results.iter().take(seeds.len()) {
-                match &*result {
-                    Ok(_updates) => {}
-                    Err(_err) => {}
-                }
+    // Get seeds.
+    let seeds = node.seeds(id)?;
```
slack-coder, line 153: Comment probably isn't needed

cloudhead: Indeed, I will expand it to something more useful, since lots of stuff is going on in this function.

radicle-cli/tests/commands.rs
```diff
@@ -192,3 +152,23 @@ fn rad_patch() {
...
+
+    // Setup a test project.
+    let _ = alice.project("acme", "");
+
+    let alice = alice.spawn(Config::default());
+    let mut bob = bob.spawn(Config::default());
+
+    bob.connect(&alice).converge([&alice]);
+
+    test("examples/rad-clone.md", working, Some(&bob.home), []).unwrap();
```
slack-coder, line 173: This is better than before :+1:

As a follow up, how about having the full workflow covered in the documentation tests?  Having everything covered in the documentation tests makes it very clear to readers as well as testers, with everything being in one place.

Here Alice's project setup could take place in a documentation test, or part of 'rad-init.md'.

radicle-cli/tests/framework/mod.rs
```diff
@@ -162,7 +164,14 @@ impl TestFormula {
                 } else {
                     PathBuf::from(&assertion.program)
                 };
+                log::debug!(target: "test", "Running `{}` in `{}`..", program.display(), self.cwd.display());
 
+                if !program.exists() {
+                    log::error!(target: "test", "Program {} does not exist..", program.display());
```
slack-coder, line 170: Nitpick, would 'command' be better than 'program'?  'command' is more well used no?

radicle-cli/tests/commands.rs
cloudhead: Hmm true, although rad-auth won't be covered, so not sure we're gaining much? We're also starting the node in the test and waiting for it to converge..

I think if we want to test everything, we should find a way to test everything; I'm not sure we gain much from having 80% in the doc test vs 50%..

radicle-node/src/control.rs
```diff
@@ -70,208 +76,179 @@ enum DrainError {
...
+        let arg1 = arg1
+            .parse()
+            .map_err(|e| CommandError::InvalidCommandArg(arg1, Box::new(e)))?;
+        let arg2 = arg2
+            .parse()
+            .map_err(|e| CommandError::InvalidCommandArg(arg2, Box::new(e)))?;
+
+        Ok((arg1, arg2))
+    }
+}
```
slack-coder, line 250: Maybe we could generalize this by using a parser? Either as a function or a type. This could simplify the code above and generalize it.  So:

```
let cmd: Command = json::from_str(input)?;
let mut parser = parser(cmd);
...
        CommandName::Fetch => {
            let rid: Id = parser.parse()?;
            let nid: NodeId = parser.parse()?;
```

or something like:

```
        CommandName::Fetch => {
            let (rid, args) = parseArgs(cmd.args)?;
            let (nid, args) = parseArgs(args)?;
```

radicle-cli/tests/commands.rs
slack-coder: > I think if we want to test everything, we should find a way to test everything

I'm unsure about covering all the nuances but :+1:

radicle/src/node.rs
```diff
@@ -1,32 +1,79 @@
...
+impl CommandResult {
+    /// Create an "updated" response.
+    pub fn updated() -> Self {
+        Self::Okay { updated: true }
+    }
+
+    /// Create an "ok" response.
+    pub fn ok() -> Self {
+        Self::Okay { updated: false }
+    }
```
slack-coder, line 53: Are the 'updated' flags for user UI? Or something else?

If we run the command twice in a row, would it matter if the first is Ok(update: true) and the second is Ok(updated: false)?

cloudhead: It's for user feedback yeah. User needs to know if something happened, or if it was a no-op.

> If we run the command twice in a row, would it matter if the first is Ok(update: true) and the second is Ok(updated: false)?

This would be the expected behavior for certain commands, yes.

radicle-node/src/control.rs
cloudhead: We could, but I think it's not worth it until there are more cases.

PR #268, opened 2023-02-03
NoiseXK

cloudhead:


radicle-node/src/wire/protocol.rs
```diff
@@ -49,12 +47,99 @@ impl<G: Signer + EcSign> fmt::Debug for Control<G> {
...
+
+    fn artifact(&self) -> Option<Self::Artifact> {
+        println!(">>>>>>>>>> {} ARTIFACT", self.node);
+        None
+        // self.state.remote_cert().cloned()
+    }
+
+    fn is_init(&self) -> bool {
+        println!(">>>>>>>>>> {} IS INIT", self.node);
+        true
```
dr-orlovsky, line 130: I do not remember what this method should do, but it looks like this `true` is the source why the handshake doesn't progress

PR #273, opened 2023-02-06
radicle-httpd: Add DELETE route to remove sessions

sebastinez:
To allow the web client to remove a session if requested by the user, we should remove the entry represented by the `session_id`.

New added route:
`DELETE /sessions/:id`

Things to talk about:
- [ ] Should we maybe require some auth challenge for killing a session, so it wouldn't be as easy as brute force spamming a httpd with delete requests. (don't think there is a benefit, for the attacker).

### DIFF
--- a/radicle-httpd/src/api/v1/sessions.rs
+++ b/radicle-httpd/src/api/v1/sessions.rs
@@ -4,6 +4,7 @@ use axum::extract::State;
 use axum::response::IntoResponse;
 use axum::routing::{post, put};
 use axum::{Json, Router};
+use axum_auth::AuthBearer;
 use hyper::StatusCode;
 use radicle::crypto::{PublicKey, Signature};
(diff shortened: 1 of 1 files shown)


cloudhead:
I think for consistency, all updates to internal state should require being signed in.

radicle-httpd/src/api/v1/sessions.rs
```diff
@@ -103,6 +106,22 @@ async fn session_signin_handler(
     Err(Error::Auth("Session already authorized"))
 }
 
+/// Delete session.
+/// `DELETE /sessions/:id`
+async fn session_delete_handler(
+    State(ctx): State<Context>,
+    AuthBearer(token): AuthBearer,
+    Path(session_id): Path<String>,
```
sebastinez, line 114: I kept the `session_id` in the route path, since we are deleting a particular session id

PR #276, opened 2023-02-07
cli: Use localhost for rad web

rudolfs:
Since sessionStorage and localStorage use protocol + host + port as scope, we want to align this between the CLI and radicle-interface. Using localhost has other benefits as well, because browsers treat it differently when doing requests from the deployed app which is being served via https.


sebastinez (approved):
I'm okay with this. 👍

PR #279, opened 2023-02-08
rad:  Add comment `reply-to`

slack-coder:
Allow replying to patch and issue comments using `rad comment --reply-to` option.

references #187

### DIFF
--- a/radicle-cli/src/commands/comment.rs
+++ b/radicle-cli/src/commands/comment.rs
@@ -7,6 +7,7 @@ use radicle::cob;
 use radicle::cob::issue::Issues;
 use radicle::cob::patch::Patches;
 use radicle::cob::store;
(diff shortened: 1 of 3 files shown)


radicle-cli/src/commands/comment.rs
```diff
@@ -22,11 +23,12 @@ pub const HELP: Help = Help {
     usage: r#"
 Usage
 
-    rad comment <id> [-m <text>]
+    rad comment <id> [options...]
 
 Options
 
     -m, --message               Comment message
+        --reply-to <id>         Index of comment writing a reply for
```
cloudhead, line 31: Something like this maybe:
```suggestion
        --reply-to <comment>        Reply to a comment
```

cloudhead (approved):

PR #277, opened 2023-02-07
radicle-httpd: Add more issue endpoint tests

sebastinez:
- Adds a `PATCH` method to `api::test` for e.g. updating issues.
- Adds endpoint tests for creating issues, creating a top-level comment, and replying to a comment.
- Also moves around some parts for better DX

PR #280, opened 2023-02-08
Update `radicle-git` and `radicle-surf`

xphoniex:


Cargo.toml
```diff
@@ -30,8 +30,20 @@ incremental = false
 
 [patch.crates-io.radicle-git-ext]
 git = "https://github.com/radicle-dev/radicle-git"
-rev = "48022590d0439e7fb196605a573a725bbbf0e4d2"
+#rev = "48022590d0439e7fb196605a573a725bbbf0e4d2"
+branch = "diff-eof"
```
xphoniex, line 34: awaiting merge of https://github.com/radicle-dev/radicle-git/pull/109

radicle-httpd/src/api/json.rs
```diff
@@ -46,14 +47,18 @@ pub(crate) fn session(session_id: String, session: &Session) -> Value {
...
-    json!({
-        "binary": blob.is_binary(),
-        "content": blob.content(),
-        "name": name_in_path(path),
-        "path": path,
-        "lastCommit": commit(blob.commit())
-    })
+pub(crate) fn blob<T: AsRef<[u8]>>(blob: &Blob<T>, path: &str) -> Value {
+    if let Ok(content) = str::from_utf8(blob.content()) {
+        json!({
```
xphoniex, line 52: `blob` is now generic and returns a vec of bytes, thus I'm converting it into a utf8 string here. do we want to treat binary blobs differently?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -634,7 +634,8 @@ mod routes {
                                 }
                               ]
                             }
-                          ]
+                          ],
+                          "eof": "noneMissing",
```
xphoniex, line 638: effect of https://github.com/radicle-dev/radicle-git/pull/109 ?

xphoniex:
as you can see in this [job](https://github.com/radicle-dev/heartwood/actions/runs/4125966845/jobs/7127244067), hash for `rad issue` test has changed, a similar effort is made [here](https://github.com/radicle-dev/heartwood/pull/277/files) to fix this, but this hash is different. please advise?

(I checked and it's not related to `.env("LANG", "C")` which has been added in aforementioned pr)

xphoniex:
As you can see in [this](https://github.com/radicle-dev/heartwood/actions/runs/4134049594/jobs/7144701255) run, the hash for test of `rad issue` has been changed, yet we have not made any changes to `issue` internals and it was supposed to have been stable.

With Alexis, we've narrowed down the potential causes to either `git2 (0.15.0 -> 0.16.1)` or `git-commit (0.2 -> 0.3)`/`git-ref-format (0.1 -> 0.2)`.

Any ideas what's causing this @FintanH @keepsimple1 ?

FintanH:
> As you can see in [this](https://github.com/radicle-dev/heartwood/actions/runs/4134049594/jobs/7144701255) run, the hash for test of `rad issue` has been changed, yet we have not made any changes to `issue` internals and it was supposed to have been stable.
> 
> With Alexis, we've narrowed down the potential causes to either `git2 (0.15.0 -> 0.16.1)` or `git-commit (0.2 -> 0.3)`/`git-ref-format (0.1 -> 0.2)`.
> 
> Any ideas what's causing this @FintanH @keepsimple1 ?

It would help to know how you narrowed it down to those scenarios.

How are we stabilising the issue ids in the first place? Are we using the `git` process to run any commits for it? Why is it only this issue id that keeps changing and not patches or anything else?

radicle-httpd/src/api/json.rs
keepsimple1: Good question. The previous version relies on `BlobContent` type serialization provided by `radicle-surf` itself.  Now we removed `BlobContent` and return `&[u8]` directly. 

Your change looks good to me, and you can add handling for the binary case too, for example converting into Base64. That provides the max freedom and not to rely on the serialization from `radicle-surf`. 

That said, I think we can add a couple of new methods in `radicle-surf` to help: 

`blob.content_str()` that returns `&str` when the content is not binary, otherwise returns empty.  
`blob.content_base64()` that returns `String` encoded in Base64. The caller uses this method when `is_binary()` is true.

(The reason to have two methods rather than one method is that, I don't want to clone a new string in the regular case of text content).

keepsimple1:
> Any ideas what's causing this @FintanH @keepsimple1 ?

I found the root cause: [in `radicle-git`](https://github.com/radicle-dev/radicle-git/blob/4457f1242877bbd9fb32b218d85dfc088ca169a7/git-commit/src/lib.rs#L290) . 

It changes the `commit.to_string()` output, which further changes the hash of the commit object. 

I removed this line and did a test, the problem is gone.

xphoniex:
> 
> It changes the `commit.to_string()` output, which further changes the hash of the commit object.

so what's being added in the new string representation? is it necessary or can be reverted?

keepsimple1:
> so what's being added in the new string representation? is it necessary or can be reverted?

It's simply a newline ;-).  I think we can revert it.  Here is my test log :

before change:
```
    14 + Create issue
    15 + Rad-Resource: d96f425412c9f8ad5d9a9a05c9831d0728e2338d
```

after change:
```
    14 + Create issue
    15 + 
    16 + Rad-Resource: d96f425412c9f8ad5d9a9a05c9831d0728e2338d
```

keepsimple1:
More accurately, we shall fix it.  The newline was added in [this commit](https://github.com/radicle-dev/radicle-git/commit/295b0d25e5065ed7f59ae4743f9c5149ca49aa92).  But it introduces an extra newline in normal cases. I will look into a fix...

keepsimple1:
I've opened a PR for the fix. However there are two things: 

1. It's not clear to me what exact string format / newlines we would like for the output. 
2. It seems it is a bit fragile for tests to rely on the hash of the commit object, unless we settle on a well-defined format mentioned in point 1.

keepsimple1:
My PR in `radicle-git` failed its own `to_string()` tests. I looked at the output in tests, and I actually think it looks better [without the fix](https://github.com/radicle-dev/radicle-git/blob/4457f1242877bbd9fb32b218d85dfc088ca169a7/git-commit/t/src/commit.rs#L18-L31). 

In other words, I think the extra newline separates the trailers into its own section, which looks good to me.  My suggestion is that for this PR we can document this, and update the tests in `heartwood` to expect a new hash.  And we can discuss what we want to deal with this coupling going forward.

cloudhead:
Trailers are supposed to be separated by a new-line, according to this section of the git docs:

```
By default the new trailer will appear at the end of all the existing trailers. If there is no existing trailer, the new trailer will appear after the commit message part of the output, and, if there is no line with only spaces at the end of the commit message part, one blank line will be added before the new trailer.
```
https://git-scm.com/docs/git-interpret-trailers

Therefore, if the current output doesn't include a blank line, I'd consider that a bug, with the update to git-commit being a fix. In that case we should update the hashes.

The exact format should also be standardized, and it's something we'll include in the relevant RIP.

keepsimple1:
> Therefore, if the current output doesn't include a blank line, I'd consider that a bug, with the update to git-commit being a fix. In that case we should update the hashes.

👍  we should update the hashes then.

radicle-httpd/src/api/json.rs
keepsimple1: I thought about it a bit more. I still recommend to handle the serialization of both plain and binary case in this function. Something like this: 

```rust
    let base64_string: String;
    let content = match str::from_utf8(blob.content()) {
        Ok(s) => s,
        Err(_) => { base64_string = base64::encode(blob.content()); &base64_string }
    };
    json!({
        "binary": blob.is_binary(),
        "content": content,
        "name": name_in_path(path),
        "path": path,
        "lastCommit": commit(blob.commit())
    })
```

I think adding `blob.content_str()` and `blob.content_base64()` in `radicle-surf` is against our basic idea of letting our client (i.e. `httpd`) to do their serialization. And such methods' purposes are too narrow, not something coherent with the blob itself.

FintanH:
Ya, it seems like the first time I got it wrong and ended up fixing it. Apologies for the wild goose chase :sweat_smile:

radicle-httpd/src/api/json.rs
xphoniex: not sure about the benefits of using `base64`, doesn't github also return string representation of binary files when clicking raw?

xphoniex: also can `str::from_utf8_unchecked` be used here?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -423,11 +423,14 @@ async fn issue_create_handler(
...
     let mut issues = Issues::open(ctx.profile.public_key, &repo)?;
-    issues
+    let issue = issues
         .create(issue.title, issue.description, &issue.tags, &signer)
         .map_err(Error::from)?;
 
-    Ok::<_, Error>((StatusCode::CREATED, Json(json!({ "success": true }))))
+    Ok::<_, Error>((
+        StatusCode::CREATED,
+        Json(json!({ "success": true, "id": format!("{}", issue.id()) })),
```
cloudhead, line 432: `issue.id().to_string()`

radicle-httpd/src/api/json.rs
cloudhead: I wouldn't do that unless we're 100% sure that all non-binary files are UTF-8.

radicle-httpd/src/api/json.rs
```diff
@@ -46,14 +47,18 @@ pub(crate) fn session(session_id: String, session: &Session) -> Value {
...
+            "binary": blob.is_binary(),
+            "content": content,
+            "name": name_in_path(path),
+            "path": path,
+            "lastCommit": commit(blob.commit())
+        })
+    } else {
+        json!({"success": false, "message": "error parsing file"})
```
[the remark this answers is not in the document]
cloudhead: The UI actually is able to display a route with a binary blob, it just shows up as "Binary content"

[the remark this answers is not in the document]
xphoniex: we're not failing here either, only if binary is not uft-8. I don't know of any other `fn` to convert bytes to string?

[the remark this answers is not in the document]
cloudhead: No like, what we used to do is if `blob.is_binary()`, we show everything except the content. If it's not binary, we show the UTF-8 representation. If it's not binary and not utf-8, we return an error.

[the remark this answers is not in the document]
FintanH: This is what executable/binary files look like on GitHub https://github.com/radicle-dev/git-platinum/blob/master/bin/cat, for reference.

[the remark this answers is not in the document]
xphoniex: > This is what executable/binary files look like on GitHub https://github.com/radicle-dev/git-platinum/blob/master/bin/cat, for reference.

but how would we implement `?raw=true` part?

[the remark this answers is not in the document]
FintanH: `?is_binary=true`?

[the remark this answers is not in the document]
FintanH: Or what do you mean? When `?raw=true` it ended up downloading the file for me. Which I guess makes sense. If I wanted to view the file's content personally then I'd probably use some other tool for that.

radicle-httpd/src/api/v1/projects.rs
xphoniex: https://github.com/radicle-dev/heartwood/compare/0613d00fa6ae673abcb7ebf24fe5a004d97c3931..82a8189ae11c6a6612fc3bf4ba3cd92928f1494e#diff-85a73839d407aa508b9e78b3756b0e9b6e8d1087a0cafac6a8325a8cf427b65eR432

radicle-httpd/src/api/json.rs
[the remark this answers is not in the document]
xphoniex: [diff](https://github.com/radicle-dev/heartwood/compare/0613d00fa6ae673abcb7ebf24fe5a004d97c3931..82a8189ae11c6a6612fc3bf4ba3cd92928f1494e#diff-c99690711746c76aea217441371c5d7b1657d4463916a2a62de0617149f93e6bR51-R65)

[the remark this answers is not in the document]
cloudhead: raw=true cannot work through the httpd api, since the api only serves json. So I wouldn't worry about it for now.

radicle-httpd/src/api/json.rs
```diff
@@ -46,14 +47,22 @@ pub(crate) fn session(session_id: String, session: &Session) -> Value {
...
         "name": name_in_path(path),
         "path": path,
         "lastCommit": commit(blob.commit())
-    })
+    });
+
+    if !blob.is_binary() {
+        match str::from_utf8(blob.content()) {
+            Ok(content) => response["content"] = content.into(),
+            Err(err) => return json!({"success": false, "error": err.to_string()}),
```
cloudhead, line 61:
```suggestion
            Err(err) => return json!({ "error": err.to_string() }),
```
`success` should only be used for PUT/POST/PATCH, ie. if there's an operation being carried out.
Looks good otherwise!

radicle-httpd/src/api/json.rs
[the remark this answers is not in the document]
xphoniex: yeah, it should be delegated to axum

radicle-httpd/src/api/json.rs
xphoniex: [diff](https://github.com/radicle-dev/heartwood/compare/82a8189ae11c6a6612fc3bf4ba3cd92928f1494e..69bf6421accc8a349cd36fe79d5c7e5479ebf647)

cloudhead (approved):

PR #281, opened 2023-02-08
cli: Allow specifying protocol for `rad web` frontend

rudolfs:
This also adds an informative text instructing the user what clicking the link does and improves the CLI argument names.
If no arguments are passed point the user to the production web app at https://app.radicle.xyz.

### DIFF
--- a/radicle-cli/src/commands/web.rs
+++ b/radicle-cli/src/commands/web.rs
@@ -18,8 +18,8 @@ Usage
 
 Options
 
-    --host, -h             httpd host to bind to
-    --web, -w              interface host to bind to
+    --backend, -b          httpd to bind to
+    --frontend, -f         Web interface to bind to
     --verbose, -v          Verbose output
     --help                 Print help
 "#,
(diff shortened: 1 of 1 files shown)


sebastinez (commented):
LGTM, not sure though on removing the protocol from the backend param. 🤔 
Since it can't be `https` currently. But makes probably sense to have both params taking a full URL.

radicle-cli/src/commands/web.rs
```diff
@@ -84,15 +84,16 @@ pub fn sign(signer: Box<dyn Signer>, session: &SessionInfo) -> Result<Signature,
...
         .call()?
         .into_json()?;
     let profile = ctx.profile()?;
     let signer = profile.signer()?;
     let signature = sign(signer, &session)?;
+    term::info!("\nOpen the following link to authenticate:");
     term::info!(
-        "http://{}/session/{}?pk={}&sig={}",
-        options.web,
+        "  👉 {}/session/{}?pk={}&sig={}\n",
```
cloudhead, line 95: We prefer to use `term::blank()` instead of an `\n`

cloudhead (approved):

PR #284, opened 2023-02-08
Announce refs on `rad init`

cloudhead:
See commits.

### DIFF
--- a/radicle-cli/src/terminal/io.rs
+++ b/radicle-cli/src/terminal/io.rs
@@ -87,7 +87,7 @@ pub fn help(name: &str, version: &str, description: &str, usage: &str) {
 }
 
 pub fn usage(name: &str, usage: &str) {
-    eprintln!(
+    println!(
         "{} {}\n{}",
         style("==").red(),
         style(format!("Error: rad-{name}: invalid usage")).red(),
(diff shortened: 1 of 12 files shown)


radicle-cli/src/terminal/io.rs
```diff
@@ -87,16 +87,16 @@ pub fn help(name: &str, version: &str, description: &str, usage: &str) {
...
         "{} {}\n{}",
         style("==").red(),
         style(format!("Error: rad-{name}: invalid usage")).red(),
         style(prefixed(TAB, usage)).red().dim()
     );
 }
 
-pub fn eprintln(prefix: impl fmt::Display, msg: impl fmt::Display) {
-    eprintln!("{prefix} {msg}");
+pub fn println(prefix: impl fmt::Display, msg: impl fmt::Display) {
```
FintanH, line 98: These changes allow us to test the unhappy paths in the CLI examples, right?

I noticed I couldn't do that in the `rad id` tests.

radicle-cli/tests/framework/mod.rs
```diff
@@ -8,6 +8,9 @@ use snapbox::cmd::{Command, OutputAssert};
 use snapbox::{Assert, Substitutions};
 use thiserror::Error;
 
+/// Error lines in the CLI are prefixed with this string.
+const ERROR_PREFIX: &str = "==";
```
FintanH, line 12: I might need to update my `rad id` output, since I was using `=== <header> ===`. Any suggestions on what the headers should look like?

radicle-cli/src/terminal/io.rs
cloudhead: Yes exactly! It didn't really make sense for half the output to be in stdout and the other half in stderr.

radicle-cli/tests/framework/mod.rs
cloudhead: Yeah indeed I think that should change. I would keep it simple for now and just have headers with a different color to the body text, plus bold or underline. At some point maybe we'll spice things up a bit.

PR #285, opened 2023-02-09
node: Move to NoiseXK

cloudhead:
Move from NoiseNN with a custom authentication protocol, to NoiseXK.
Uses ECDH on the ed25519 curve via the `ec25519` library.


dr-orlovsky (approved):
LGTM!

PR #289, opened 2023-02-09
radicle-httpd: Add `assignees` to Issues::create

sebastinez:
We currently don't allow to pass assignees during the issue creation.

This PR allows the passing of a vector of ActorIds during issue creation by the httpd or the cli
Also modifies some tests to check that the issue creation, handles tags and assignees accordingly

### DIFF
--- a/radicle/src/cob/issue.rs
+++ b/radicle/src/cob/issue.rs
@@ -193,7 +193,11 @@ impl Deref for Issue {
 }
 
 impl store::Transaction<Issue> {
-    pub fn assign(&mut self, add: Vec<ActorId>, remove: Vec<ActorId>) -> OpId {
+    pub fn assign(
+        &mut self,
+        add: impl IntoIterator<Item = ActorId>,
+        remove: impl IntoIterator<Item = ActorId>,
+    ) -> OpId {
         let add = add.into_iter().collect::<Vec<_>>();
         let remove = remove.into_iter().collect::<Vec<_>>();
 
@@ -273,10 +277,10 @@ impl<'a, 'g> IssueMut<'a, 'g> {
     /// Assign one or more actors to an issue.
     pub fn assign<G: Signer>(
         &mut self,
-        assignees: Vec<ActorId>,
+        assignees: impl IntoIterator<Item = ActorId>,
         signer: &G,
     ) -> Result<OpId, Error> {
-        self.transaction("Assign", signer, |tx| tx.assign(assignees, vec![]))
+        self.transaction("Assign", signer, |tx| tx.assign(assignees, []))
     }
 
     /// Set the issue title.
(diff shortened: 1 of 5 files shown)


radicle/src/cob/issue.rs
```diff
@@ -493,11 +495,24 @@ mod test {
...
+                &signer,
+            )
             .unwrap();
 
-        issue.assign(vec![assignee, assignee_two], &signer).unwrap();
+        let id = issue.id;
+        let issue = issues.get(&id).unwrap().unwrap();
+        let assignees: Vec<_> = issue.assigned().cloned().collect::<Vec<_>>();
+        assert_eq!(1, assignees.len());
+        assert!(assignees.contains(&assignee));
```
xphoniex, line 512: please add an empty line between setups and asserts.

xphoniex (commented):
The code looks good. However UX-wise, is this something we want? e.g. Github doesn't allow you to assign on creation, you have to first create the issue, and then you can assign someone - if you want - and most of the time you won't.

sebastinez:
> Github doesn't allow you to assign on creation, you have to first create the issue, and then you can assign someone.

It's the opposite in the issue creation screen you can assign yourself or someone different.
Bildschirm­foto 2023-02-09 um 16 36 12
I agree that most of the time you wouldn't do it, but in some cases e.g.:
- I like to assign some issues directly to myself
- If I know who has to do something I like to assign it directly to this person.

I wouldn't restrict that on a API level, if it's something the user doesn't want we can always hide it on the web client.

radicle/src/cob/issue.rs
```diff
@@ -411,16 +411,18 @@ impl<'a> Issues<'a> {
         title: impl ToString,
         description: impl ToString,
         tags: &[Tag],
+        assignees: &Vec<ActorId>,
```
cloudhead, line 414: This should take a slice: `&[ActorId]`, we shouldn't force the caller to have a vector at hand.

radicle/src/cob/issue.rs
```diff
@@ -669,8 +701,8 @@ mod test {
         issue.comment("Re: Ha.", c2, &signer).unwrap();
         issue.comment("Re: Ha. Ha.", c2, &signer).unwrap();
         issue.comment("Re: Ha. Ha. Ha.", c2, &signer).unwrap();
-
```
cloudhead, line 672: Think we can leave this line

sebastinez: Done in 1ad7c47

radicle/src/cob/issue.rs
sebastinez: Done in 1ad7c47aaa3417c0bc23a02efd5c2be2798c2bed

radicle/src/cob/issue.rs
```diff
@@ -515,15 +536,21 @@ mod test {
...
+                "Blah blah blah.",
+                &[],
+                &[assignee],
+                &signer,
+            )
             .unwrap();
 
-        issue.assign(vec![assignee], &signer).unwrap();
-        issue.assign(vec![assignee], &signer).unwrap();
+        issue.assign([assignee], &signer).unwrap();
```
cloudhead, line 548: This is no longer testing the same thing, I'd suggest having two different assignees, one of them is added in issue creation, and the other one is added *twice* after issue creation. Then we test that each is returned once.

sebastinez: This should do the trick 6c53eeb

FintanH (approved):
Looks good from my perspective :)

sebastinez:
Ok did a rebase and had to modify the used issue ids in tests again.

cloudhead (approved):

PR #296, opened 2023-02-13
radicle-cob: simplify trailers module

FintanH:
The macro in the trailers module existed since there were multiple trailers before the refactoring, so it minimised boilerplate code.

After the refactoring of the code there was only one trailer left. Remove the macro generation code and replace it with just the necessary code for the ResourceCommitTrailer.


cloudhead (approved):
Excellent!

PR #300, opened 2023-02-13
Handle unix signals

cloudhead:


radicle-node/src/signals.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+    Ok(())
+}
+
+/// Install global signal handlers for `SIGTERM` and `SIGINT`.
+///
+/// # Safety
+///
+/// Calls `libc` functions safely.
+unsafe fn _install() -> io::Result<()> {
```
slack-coder, line 35: why the underscore?

radicle-node/src/signals.rs
```diff
@@ -0,0 +1,55 @@
...
+        }
+        *channel = Some(notify);
+
+        unsafe { _install() }?;
+    } else {
+        return Err(io::Error::new(
+            io::ErrorKind::WouldBlock,
+            "unable to install signal handler",
+        ));
+    }
```
slack-coder, line 26: What about lifting the nested if statement using the new let statements?

i.e.
```
let Ok(mut channel) = ... else {
  return Err(..)
} 
```

slack-coder:
It could be possible to have the signals clearly mapped to their actions, maybe in main?  However it may take some refactoring, therefore as long as this works, its good for me.

radicle-node/src/signals.rs
cloudhead: There's another function called `install` that calls this function :) It's a convention I've seen in a bunch of places when you have an inner and an outer function.

radicle-node/src/signals.rs
cloudhead: Tbh I don't think it really makes a difference here, the function is small..

cloudhead:
Yeah, I opted for the signal setup to be in main, while the signal handling logic is in the library code. This allows triggering the signal handler in tests by sending a message on the channel.

adaszko (approved):
Looks good 👍

FintanH (approved):
LGTM

PR #306, opened 2023-02-14
Make `ReadStorage` more usable

cloudhead:
Previously, we were not able to access a `ReadRepository` instance through `ReadStorage`. With this change, we are able to.


FintanH (approved):
Makes sense :+1:

PR #308, opened 2023-02-14
node: Fix bug when socket file is deleted

### LINKED ISSUE #307: Node fails to shutdown if radicle.sock is deleted
The radicle-node becomes a zombie when it's socket is deleted while its running, requiring a SIGKILL.  The service should shutdown even in a hostile environment with appropriate warnings.

cloudhead:
If the socket file is deleted while the node is running, we can no longer quit with Ctrl-C. This is the "fix".

Fixes #307


vincenzopalazzo (approved):
I just noted in these days that the socket was not deleted! Good!

FintanH (approved):
Makes sense :+1:

cloudhead:
Socket is deleted since #300

Issue #324, opened 2023-02-16
Use Did for CLI Arguments


FintanH:
In a few places, we expect the `PublicKey` as the CLI argument:
* https://github.com/radicle-dev/heartwood/blob/master/radicle-cli/examples/rad-delegate.md
* https://github.com/radicle-dev/heartwood/blob/master/radicle-cli/examples/rad-issue.md
* https://github.com/radicle-dev/heartwood/blob/master/radicle-cli/examples/rad-id-rebase.md
* https://github.com/radicle-dev/heartwood/blob/master/radicle-cli/examples/rad-id.md

Instead, we should make it standard to expect the `Did`, i.e. `did:key:<key>`. It would also be nice to have a helper that detects the operator has pasted in a `PublicKey` and make a suggestion to use `did:key:<key>` instead.

PR #326, opened 2023-02-16
radicle-httpd: Use `radicle::identity::did::Did` for json output instead of `PublicKey`

sebastinez:
Instead of showing just `PublicKey` we need to show the fully qualified spec for a NID (NodeId) which is prefixed by `did:key:`
While we want to maintain `Author` and `ActorId` for internal use.

### DIFF
--- a/radicle-httpd/src/api/json.rs
+++ b/radicle-httpd/src/api/json.rs
@@ -10,7 +10,7 @@ use radicle::cob::issue::{Issue, IssueId};
 use radicle::cob::patch::{Patch, PatchId};
 use radicle::cob::thread::{self, CommentId};
 use radicle::cob::{OpId, Timestamp};
-use radicle::identity::PublicKey;
+use radicle::identity::{Did, PublicKey};
 use radicle_surf::blob::Blob;
 use radicle_surf::tree::Tree;
 use radicle_surf::{Commit, Stats};
@@ -133,6 +133,7 @@ fn name_in_path(path: &str) -> &str {
 #[derive(Serialize)]
 struct Author {
     id: PublicKey,
+    did: Did,
 }
 
 #[derive(Serialize)]
@@ -158,6 +159,7 @@ impl<'a> FromIterator<(&'a CommentId, &'a thread::Comment)> for Comments {
                 id: id.to_owned(),
                 author: Author {
                     id: comment.author(),
+                    did: comment.author().into()
                 },
                 body: comment.body().to_owned(),
                 reactions: [],
--- a/radicle/src/cob/common.rs
+++ b/radicle/src/cob/common.rs
@@ -11,11 +11,12 @@ pub use radicle_crdt::clock::Physical as Timestamp;
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct Author {
     pub id: NodeId,
+    pub did: Did,
 }
 
 impl Author {
     pub fn new(id: NodeId) -> Self {
-        Self { id }
+        Self { id, did: id.into() }
     }
 
     pub fn id(&self) -> &NodeId {


radicle/src/identity/did.rs
```diff
@@ -46,6 +47,18 @@ impl From<crypto::PublicKey> for Did {
...
+    fn from(key: Author) -> Self {
+        (key.id()).into()
+    }
+}
+
+impl From<&Author> for Did {
+    fn from(key: &Author) -> Self {
+        Self(*key.id())
+    }
+}
```
FintanH, line 60: Organisational nit: I'd expect these implementations to live beside `Author`, since `Did` is more fundamental than `cob::Author`.

radicle-httpd/src/api/json.rs
```diff
@@ -91,22 +91,26 @@ pub(crate) fn tree(tree: &Tree, path: &str, stats: &Stats) -> Value {
...
+    let author: Did = issue.author().into();
+
     json!({
         "id": id.to_string(),
-        "author": issue.author(),
+        "author": author,
         "title": issue.title(),
         "state": issue.state(),
-        "assignees": issue.assigned().collect::<Vec<_>>(),
+        "assignees": issue.assigned().collect::<Dids>(),
```
FintanH, line 101: Perhaps `assigned` should return `Did`s instead.

FintanH (commented):
Looking good, but have a couple of suggestions

radicle-httpd/src/api/json.rs
```diff
@@ -130,16 +134,11 @@ fn name_in_path(path: &str) -> &str {
...
-struct Author {
-    id: PublicKey,
-}
-
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
 struct Comment {
     id: OpId,
-    author: Author,
+    author: Did,
```
cloudhead, line 141: This prevents us from for eg. including an avatar and "name" field in the author object in the future.

radicle/src/cob/issue.rs
```diff
@@ -168,11 +168,12 @@ impl Issue {
         self.tags.iter()
     }
 
-    pub fn author(&self) -> Option<Author> {
+    pub fn author(&self) -> Author {
         self.thread
             .comments()
             .next()
             .map(|(_, c)| Author::new(c.author()))
+            .expect("Issue::author: at least one thread is present")
```
cloudhead, line 176:
```suggestion
            .expect("Issue::author: at least one comment is present")
```

radicle-httpd/src/api/json.rs
sebastinez: I'm trying to change the return type of `assigned` but my rust is too low level it seems, or I have to wrap it inside a Wrapper struct which I think is a PITA.
Can we maybe collect the iterator inside `assigned` into a `Vec<Author>`?

FintanH: I did it here https://github.com/radicle-dev/heartwood/pull/328/files#diff-d8a3305359907b1f57d685f419f888f95c5738c23a3420ca4e7150e9f42cbd4eR156-R158 :)

sebastinez: nice 👍

radicle-httpd/src/api/json.rs
```diff
@@ -169,3 +168,32 @@ impl<'a> FromIterator<(&'a CommentId, &'a thread::Comment)> for Comments {
...
+    fn from_iter<I: IntoIterator<Item = &'a PublicKey>>(iter: I) -> Self {
+        let mut dids = Vec::new();
+
+        for pubkey in iter {
+            dids.push(pubkey.into());
+        }
+
+        Dids(dids)
+    }
+}
```
FintanH, line 199: No longer needed.

radicle-httpd/src/api/json.rs
FintanH: I guess `assignees` should be the same?

sebastinez:
Ok did rewrite the PR on base of @FintanH PR, which has done the heavy lifting.
Now this PR just removes the `Option` from the `issue.author()` and adds a `Did` type to the `Author` struct.

radicle-httpd/src/api/json.rs
sebastinez: Done

radicle/src/cob/issue.rs
sebastinez: Done

radicle-httpd/src/api/json.rs
sebastinez: Has been reverted

radicle-httpd/src/api/json.rs
sebastinez: Has been reverted, uses what is implemented in #324

radicle/src/identity/did.rs
sebastinez: Has been reverted, uses what is implemented in #324

cloudhead (approved):
Nice. Probably we shouldn't rely on `author.id`, and maybe it can even be removed eventually.

cloudhead:
Argh, messed something up.

cloudhead:
Ok, what happened is that I included these changes into fintan's branch, so will leave this closed and we can merge both at the same time.

PR #327, opened 2023-02-16
node: Remove redundant handshake

cloudhead:
With `NoiseXK` handshake implemented, there is no longer a need to have an `Initialize` message and an extra state in the session.


FintanH (approved):
:fire:

PR #331, opened 2023-02-16
httpd: Add `/raw` route

xphoniex:
Please do **not** squash commits.


radicle-httpd/src/raw.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+async fn file_handler(
+    Path((project, sha, path)): Path<(Id, Oid, String)>,
+    State(profile): State<Arc<Profile>>,
+) -> impl IntoResponse {
+    let storage = &profile.storage;
+    let repo = Repository::open(paths::repository(storage, &project))?;
+    let blob = repo.blob(sha, &path)?;
+
```
xphoniex, line 29: `blob` is fully allocated, a more efficient way would be to implement a stream, any thoughts @keepsimple1 ?

radicle-httpd/src/raw.rs
```diff
@@ -0,0 +1,54 @@
...
+use axum::routing::get;
+use axum::Router;
+
+use radicle::prelude::Id;
+use radicle::profile::Profile;
+use radicle::storage::git::paths;
+use radicle_surf::{Oid, Repository};
+
+use crate::api::axum_extra::Path;
+use crate::api::error::Error;
```
xphoniex, line 14: we have resorted to using some stuff from `::api::`, any thoughts on this? (since we're no longer in `/api` but a level higher, in `/raw`)

radicle-httpd/src/raw.rs
xphoniex: btw, this causes file to be downloaded, not shown in browser. good enough for me, lmk if you want it changed.

keepsimple1: I think if we could include `repo` inside `State(profile)`,  so that we don't have to (re)open the repo every time inside the handler, that might help. 

Since `blob<'a>` has its lifetime associated with `repo`, such change will allow us return a `Blob` without calling `to_owned()`, hence would be helpful for streaming.

I could be wrong but this link seems to show it might be possible to include `repo` in the profile? https://github.com/radicle-dev/heartwood/blob/2b5552546c6dc637aaa282a990083df8a4a0c0c0/radicle-httpd/src/api.rs#L49-L52

radicle-httpd/src/api.rs
```diff
@@ -1,12 +1,18 @@
+pub mod axum_extra;
+pub mod error;
+#[cfg(test)]
+pub mod test;
```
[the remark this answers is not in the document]
FintanH: I honestly prefer all `use` statements followed by all `mod` statements, with an exception for `use` statements that come from local `mod`s.

```rust
use std;

use external;

pub mod stuff;

mod internal;
pub use internal::Things;
use internal::magic;
```

That's a personal preference though :shrug:

radicle-httpd/src/raw.rs
FintanH: > I think if we could include repo inside State(profile), so that we don't have to (re)open the repo every time inside the handler, that might help.

I'm not sure that's a good idea. You would be carrying around a handle to the `Repository` the whole time, which isn't always what you want.

radicle-httpd/src/api.rs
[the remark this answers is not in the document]
cloudhead: Hmm we've been doing it the other way around so far :D

radicle-httpd/src/raw.rs
cloudhead: The file should be shown in the browser if it's non-binary, which requires the correct content-type to be set I think.

radicle-httpd/src/raw.rs
cloudhead: Maybe better to move these out to the top level then?

radicle-httpd/src/raw.rs
```diff
@@ -0,0 +1,54 @@
...
+
+async fn file_handler(
+    Path((project, sha, path)): Path<(Id, Oid, String)>,
+    State(profile): State<Arc<Profile>>,
+) -> impl IntoResponse {
+    let storage = &profile.storage;
+    let repo = Repository::open(paths::repository(storage, &project))?;
+    let blob = repo.blob(sha, &path)?;
+
+    Ok::<_, Error>(blob.content().to_owned())
```
cloudhead, line 30: If there was a method like `blob.into() -> Vec<u8>` we could return it here without doing a copy.

keepsimple1: > If there was a method like `blob.into() -> Vec<u8>` we could return it here without doing a copy.

I don't understand as `to_owned()` is basically doing this, i.e. returning `Blob<Vec<u8>>`.  I wanted to clarify: 

- `repo.blob` returns a `Blob<BlobRef<'a>>` that does not copy or allocate anything. It reuses the `git2::Blob<'a>` internally. 

- `content()` return `&[u8]` , so does not allocate or copy.

- `to_owned()` returns `Blob<Vec<u8>>` , which makes a copy.

radicle-httpd/src/raw.rs
xphoniex: @keepsimple1 not sure what you mean, but can't you just `Box` whatever you need?

xphoniex: > The file should be shown in the browser if it's non-binary, which requires the correct content-type to be set I think.

if I set `header::CONTENT_TYPE, "text; charset=utf-8"` then:
- text files would be shown as text ✓
- images as images ✓
- binary files would start a download ✓

it shouldn't work, but browsers are doing the heavy lifting to detect the mime and works, otherwise if we want the correct mime we have to add another crate. lmk.

radicle-httpd/src/raw.rs
xphoniex: issue is not copying here, issue is serving very big files. if we serve a 200MB binary, that'll use up 200MB of memory on server, 5 people connecting at the same time means a small box will run OOM. streaming won't have this issue.

radicle-httpd/src/raw.rs
xphoniex: so there's:
1. `crate::api::axum_extra::Path`
2. `crate::api::error::Error`

first one we can safely move out. second one, only if we want all errors to be jsons, is that something we'd want for `/git` and `/raw`?

xphoniex:
are tests flaky now? it's not failing for me locally, and I didn't touch code for commands either

radicle-httpd/src/raw.rs
keepsimple1: > @keepsimple1 not sure what you mean, but can't you just `Box` whatever you need?

The purpose of my suggestion is to reduce memory usage of large blobs. 

I meant that if we could return `blob.content()` from this method, we will have a `&[u8]` that is provided by the underlying `git2::blob`. We will not use extra memory. 

For `&[u8]`,  we could stream from it if `heartwood` already have such facility. Or, add a method to return a chunk (slice) via an iterator. 

> can't you just `Box` whatever you need?

I don't think returning a `Box<&[u8]>` could not really help as the issue is the lifetime is associated with the underlying git2 repo. If we want to return `Box<Vec<u8>>`, then we will need to first make a copy to produce `Vec<u8>` anyway, which will consume extra memory.

Another approach is to write large blobs into a temp file, and then streaming from the temp file.  At the end, delete the temp file.

radicle-httpd/src/raw.rs
keepsimple1: Understood, this is my concern too, and it was the reason we refactored `repo.blob()` to return a reference to underlying git2::blob by default. 

However, if `git2::blob` already consumes big memory, we will still have a problem.  Like I mentioned in another comment, another approach is to save large blobs to a temp file on disk, then do streaming from the file.  After its done, we will delete the temp file.

xphoniex:
- Do we want to put a limit on the size of the file being downloaded?
- Do want to support zip download of just the HEAD of working copy, à la Github, as well?

If we add support for zip download, then we can keep a working copy on-disk, somewhere in `/tmp` and check for it before reading blob, and instead stream bytes from the file to endpoint.

Otherwise, we can do this operation for a files beyond a particular size.

radicle-httpd/src/raw.rs
cloudhead: Oh I see hmm. No, for raw and git, I think we just want error codes for now, no error bodies.

radicle-httpd/src/raw.rs
cloudhead: Ah my bad, I thought `Blob` internally kept an owned vector, that's why I suggested that.

How about we check the blob length; if it's smaller than `1024` bytes, we return a copy via `to_owned`. If it's larger, we return an error code [413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413)

cloudhead:
See: https://github.com/radicle-dev/heartwood/pull/331#discussion_r1109529233

cloudhead:
Rebase on master, just fixed this.

radicle-httpd/src/raw.rs
xphoniex: I'd prefer if we add an API to stream from disk for blobs > `4096`

keepsimple1: `4096` (i.e. 4KB) bytes seems too low to me. How about 4MB (or 400KB) as the threshold?  For the error code, it can work as a stopgap, but we probably need to support streaming soon.

keepsimple1: I opened a PR for the basic support of blob streaming using a tmp file: https://github.com/radicle-dev/heartwood/pull/347 .

xphoniex: > 
> How about we check the blob length; if it's smaller than `4096` bytes, we return a copy via `to_owned`. If it's larger, we return an error code [413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413)

I think we don't gain much in current implementation, because the main concern is memory usage on server side, if we already have the blob in `&[u8]`, then memory has been allocated. Bandwidth is cheap, and not much of a concern. (Unless being DDoS'd)

We can revisit this issue after #347.

cloudhead (approved):

PR #333, opened 2023-02-16
Use `Did` for json output and other places

cloudhead:
Instead of showing just public keys we need to show the fully qualified spec for a NID which is prefixed by `did:key:` While we want to maintain `Author` and `ActorId` for internal use.

### DIFF
--- a/radicle-cli/examples/rad-id.md
+++ b/radicle-cli/examples/rad-id.md
@@ -18,7 +18,7 @@ ok Identity proposal '06d9efa2a9aad06bfdf25a25690e1ec7db2c3c39' created 🌱
 title: Add Bob
 description: Add Bob as a delegate
 status:  open 
-author: z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi
+author: did:key:z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi
 
 Document Diff
 
@@ -69,21 +69,21 @@ Next we have the number of `Accepted` reviews from delegates, starting
 off with none:
 
     Accepted
-    
+
     total: 0
     keys: []
 
 The same with `Rejected` reviews:
 
     Rejected
-    
+
     total: 0
     keys: []
 
 Finally, we can see whether the `Quorum` was reached:
 
     Quorum Reached
-    
+
     ✗ no
 
 Let's see what happens when we reject the change:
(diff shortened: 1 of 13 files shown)


radicle-cli/examples/rad-id.md
```diff
@@ -136,7 +136,7 @@ Our key was added to the `Rejected` set of `keys` and the `total`
 increased to `1`.
 
     Rejected
-    
+
     total: 1
     keys: [
       "z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi"
```
FintanH, line 142: Maybe this should be `"delegates": [ "did:key:<...>", ...]`

radicle-cli/src/commands/patch/list.rs
```diff
@@ -85,7 +85,7 @@ fn print(
 ) -> anyhow::Result<()> {
     let target_head = common::patch_merge_target_oid(patch.target(), storage)?;
 
-    let you = patch.author().id() == whoami;
+    let you = &**patch.author().id() == whoami;
```
FintanH, line 88: suggestion: have a method on `Did` for checking if a `PublicKey` is owned by that `Did`

radicle/src/cob/common.rs
```diff
@@ -10,19 +10,25 @@ pub use radicle_crdt::clock::Physical as Timestamp;
...
+        Self { id: id.into() }
     }
 
-    pub fn id(&self) -> &NodeId {
+    pub fn id(&self) -> &Did {
         &self.id
     }
 }
 
+impl From<NodeId> for Author {
```
FintanH, line 26: Might make more sense to use `PublicKey` since `NodeId` is just an alias.

FintanH (approved):
Generally looks good and left a few suggestions :)

radicle-cli/examples/rad-id.md
cloudhead: Was thinking the same yes.

radicle-cli/src/commands/patch/list.rs
cloudhead: Hmmmm, I think the "ownership" relation might be a bit weird and come out of nowhere. I'd rather have maybe a `as_key()` method.

FintanH: But that's essentially what DIDs boil down to in a lot of cases, key ownership (and then associating actions across uses of those keys). But `as_key` also works, just goes in the opposite direction :)

cloudhead: Yeah, it's just that right now, our `Did` type is basically `did:key`, so I find it a bit confusing to use the verb "owns" when they're kind of the same thing..

Issue #332, opened 2023-02-16
Weird `500` error on httpd


cloudhead:
Try hitting this route: https://seed.radicle.xyz/aa/a

You'll get a 500. We should always return 404 if the route isn't valid.

FintanH:
Ah, I posted on the PR for this but... why `404` and not [`400`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400)? I would think it's a 400 because the client made a bad request using an incorrect route.

PR #335, opened 2023-02-16
httpd: Change `500` into `404` when project doesn't exist

xphoniex:
Fixes https://github.com/radicle-dev/heartwood/issues/332

### DIFF
--- a/radicle-httpd/src/api.rs
+++ b/radicle-httpd/src/api.rs
@@ -1,3 +1,6 @@
+#[cfg(test)]
+pub mod test;
+
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
@@ -22,8 +25,6 @@ mod auth;
 mod axum_extra;
 mod error;
 mod json;
-#[cfg(test)]
-mod test;
 mod v1;
 
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -63,6 +64,11 @@ impl Context {
             patches,
             id,
         })
+    }
+
+    #[cfg(test)]
+    pub fn profile(&self) -> &Arc<Profile> {
+        &self.profile
     }
 }
 
--- a/radicle-httpd/src/api.rs
+++ b/radicle-httpd/src/api.rs
@@ -1,5 +1,4 @@
-#[cfg(test)]
-pub mod test;
+pub mod auth;
 
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -21,7 +20,6 @@ use radicle::identity::Id;
 use radicle::storage::{ReadRepository, ReadStorage};
 use radicle::Profile;
 
-mod auth;
 mod axum_extra;
 mod error;
 mod json;
@@ -69,6 +67,11 @@ impl Context {
     #[cfg(test)]
     pub fn profile(&self) -> &Arc<Profile> {
         &self.profile
+    }
+
+    #[cfg(test)]
+    pub fn sessions(&self) -> &Arc<RwLock<HashMap<SessionId, auth::Session>>> {
+        &self.sessions
     }
 }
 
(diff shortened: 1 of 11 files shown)


radicle-httpd/src/lib.rs
```diff
@@ -100,13 +100,16 @@ async fn git_handler(
     query: RawQuery,
     body: Bytes,
 ) -> impl IntoResponse {
+    let mut response_headers = HeaderMap::new();
     let query = query.0.unwrap_or_default();
-    let id: Id = project.strip_suffix(".git").unwrap_or(&project).parse()?;
+    let id: Id = match project.strip_suffix(".git").unwrap_or(&project).parse() {
+        Ok(id) => id,
+        Err(_) => return Ok::<_, Error>((StatusCode::NOT_FOUND, response_headers, vec![])),
```
FintanH, line 107: Is this not a user error instead of not found? :thinking: Seems like a `400` since the repository identifier can't be parsed.

cloudhead: Hmm that would work too, all 4** errors are user errors -- the main thing is to not return a 500 here.

cloudhead: The thing is though, that it's not clear that the user is requesting a git project. The way we determine that is by checking if we can parse. So I think 404 is ok here, because it could be that the user just hit the wrong api route.

xphoniex: Yeah, I think 404 is more appropriate here as well:
- the `Id` was not found
- the route was not found

cloudhead:
Do you mind adding a small test for this? It's not clear how this change fixes it :)

xphoniex:
Cherry picked https://github.com/radicle-dev/heartwood/pull/335/commits/04898f95fc34a2de5733e26ff975f62329374ee0 from the https://github.com/radicle-dev/heartwood/pull/331 and added test.

radicle-httpd/src/api.rs
```diff
@@ -1,5 +1,4 @@
-#[cfg(test)]
-pub mod test;
+pub mod auth;
```
xphoniex, line 1: notice

xphoniex (commented):
As discussed [here](https://github.com/radicle-dev/heartwood/pull/331#discussion_r1108793290), we can move `test` utility mod from `api` mod to the root of crate, as other routes such as `/git` in this PR, and in future `/raw` will be using that.

See the required changes in this [commit](https://github.com/radicle-dev/heartwood/pull/335/commits/001837ef47872735d7b3e672b251e38b1f8faccb).

radicle-httpd/src/git.rs
```diff
@@ -0,0 +1,226 @@
...
+    use axum::body::Body;
+    use axum::extract::ConnectInfo;
+    use axum::http::Request;
+    use axum::http::{Method, StatusCode};
+    use tower::ServiceExt;
+
+    use crate::test;
+
+    #[tokio::test]
+    async fn test_issue_332() {
```
cloudhead, line 209:
```suggestion
    async fn test_invalid_route() {
```

xphoniex: any benefits? I've seen this sort of usage in other projects such as `clippy`. `invalid_route` could mean many things, whereas `issue_322` gives you complete context.

radicle-httpd/src/api.rs
xphoniex: can do `cfg(test)` and `cfg(not(test))` here, does having non-public modules affect performance in any way?

radicle-httpd/src/git.rs
cloudhead: The benefit is knowing what test fails without having to look it up. `issue_322` gives you context only in context of github.

xphoniex: [fixed](https://github.com/radicle-dev/heartwood/compare/06d5e0942902471569dfcd42fbc9a48453a14b77..22fb2d2afe976fca0ff88b13a66dd38b96b44e2c)

PR #339, opened 2023-02-20
Reject existing revisions

FintanH:
If we have two operations, a = Reivision and b = Merge, then if we
have the following for any p = Patch then:

    p.apply([a, a, b]) =/= p.apply([a, b, a])

On the RHS we end up in the case that the reinserted Revision will
make the Revision redacted.

To stop this from happening, if a Revision is existing for the same
OpId then the `apply` is rejected.

### DIFF
--- a/radicle-crdt/src/gmap.rs
+++ b/radicle-crdt/src/gmap.rs
@@ -33,6 +33,10 @@ impl<K: Ord, V: Semilattice> GMap<K, V> {
             }
         }
     }
+
+    pub fn contains_key(&self, key: &K) -> bool {
+        self.inner.contains_key(key)
+    }
 }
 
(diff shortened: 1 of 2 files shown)


FintanH:
I'm still a bit unsure if this is the right thing to do. This technically means that `GMap` composed with `Redactable` is not necessarily a CRDT since it doesn't commute :thinking:

FintanH:
I guess the problem actually comes down to being able to mutate the `GMap` using `get_mut`, which means that you're not always merging.

cloudhead:
Ohh, I think this might be the cause of an issue I've been seeing. `GMap::get_mut` should indeed not be a public function. Need to think about this.

cloudhead:
Superseded by #357

PR #340, opened 2023-02-20
Make dumped data structures more usable as input to tests

adaszko:
Quickcheck generates large data structures.  Once they’re printed on stdout, a lot of redacting is necessary in order to use them as valid inputs in example-based tests.  These changes are supposed bridge that gap just a little so that printouts can be almost copy-pasted into test cases.

### DIFF
--- a/radicle-crdt/src/ord.rs
+++ b/radicle-crdt/src/ord.rs
@@ -5,9 +5,17 @@ use serde::{Deserialize, Serialize};
 
 use crate::Semilattice;
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 #[serde(transparent)]
 pub struct Max<T>(T);
+
+impl<T: core::fmt::Debug> core::fmt::Debug for Max<T> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        // Just like the #[derive(Debug)] formatter, but without newlines to conserve screen space
+        // in large data structures.
+        write!(f, r#"Max({:?})"#, self.0)
+    }
+}
 
 impl<T> Max<T> {
     pub fn get(&self) -> &T {
--- a/radicle-crypto/src/lib.rs
+++ b/radicle-crypto/src/lib.rs
@@ -314,7 +314,7 @@ impl From<PublicKey> for String {
 
 impl fmt::Debug for PublicKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "PublicKey({self})")
+        write!(f, r#"PublicKey::from_str("{self}").unwrap()"#)
     }
 }
 


radicle-crdt/src/gmap.rs
```diff
@@ -9,7 +9,7 @@ use crate::Semilattice;
 /// Conflicting elements are merged via the [`Semilattice`] instance.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct GMap<K, V> {
-    inner: BTreeMap<K, V>,
+    pub inner: BTreeMap<K, V>,
```
FintanH, line 12: I'm not sure about the `pub` for this because it breaks encapsulation.

radicle-crdt/src/lwwmap.rs
```diff
@@ -8,7 +8,7 @@ use crate::{clock, Semilattice};
 /// the "add" takes precedence over the "remove".
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct LWWMap<K, V, C = clock::Lamport> {
-    inner: GMap<K, LWWReg<Option<V>, C>>,
+    pub inner: GMap<K, LWWReg<Option<V>, C>>,
```
FintanH, line 11: Same for here.

radicle-crdt/src/ord.rs
```diff
@@ -5,9 +5,15 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 #[serde(transparent)]
-pub struct Max<T>(T);
+pub struct Max<T>(pub T);
+
+impl<T: core::fmt::Debug> core::fmt::Debug for Max<T> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(f, r#"Max({:?})"#, self.0)
+    }
+}
```
FintanH, line 16: Is this different to what gets generated? :thinking: Is it the quotation marks?

FintanH: Ah, I see. Thanks :)

FintanH: It might be worth noting that, otherwise I imagine someone will come along and wonder the same thing :smile:

radicle-crdt/src/gmap.rs
adaszko: Yeah, I was expecting a comment like that ;)
IMO encapsulation is an inadequate relic of the shared mutable state era of OOP languages.  If you have the type system protecting your from violating invariants (by the way if mut/non-mut refs), encapsulation is just a nuisance

radicle-crdt/src/ord.rs
adaszko: Sure, will do

radicle-crdt/src/gmap.rs
FintanH: > Yeah, I was expecting a comment like that ;)

Hehehe

> IMO encapsulation is an inadequate relic of the shared mutable state era of OOP languages. If you have the type system protecting your from violating invariants (by way of mut/non-mut refs), encapsulation is just a nuisance

In this case, it's not about any shared mutable state. It's about disallowing the user of the type from accessing the internal data structure due to the CRDT semantics. A `GMap` is a "Grow-only Map" emphasising the _Grow_. If the library user suddenly has access to the `BTreeMap` then they are free to use its delete functionality. Now our CRDT semantics are out the window :disappointed:

adaszko: Right, a valid point.  I’ll remove those ‘pub’ declarations 👍

FintanH (approved):

radicle-crdt/src/ord.rs
```diff
@@ -5,10 +5,18 @@ use serde::{Deserialize, Serialize};
...
 #[serde(transparent)]
 pub struct Max<T>(T);
 
+impl<T: core::fmt::Debug> core::fmt::Debug for Max<T> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        // Just like the #[derive(Debug)] formatter, but without newlines to conserve screen space
+        // in large data structures.
+        write!(f, r#"Max({:?})"#, self.0)
+    }
+}
```
cloudhead, line 18: Makes sense.

radicle-crypto/src/lib.rs
```diff
@@ -314,7 +314,7 @@ impl From<PublicKey> for String {
 
 impl fmt::Debug for PublicKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "PublicKey({self})")
+        write!(f, r#"PublicKey::from_str("{self}").unwrap()"#)
```
cloudhead, line 317: This is weird to me, not so helpful for debugging, adds a lot of noise.

adaszko:
The PR has completely lost its substance during the review process, so let’s just forget about it

PR #345, opened 2023-02-21
cli: add missing commands for rad_help

FintanH:
The assign, id, and unassign help messages were missing from the rad help command.

Add them to complete the list.


sebastinez (approved):
Nice, this will make things a bit clearer

Issue #271, opened 2023-02-06
Node control


cloudhead:
We need some kind of way of interacting with the node:
* [x] Checking the node status (`rad node status`)
* [x] Starting the node (`rad node start`)
* [x] Stopping the node (`rad node stop`)
* [x] ~~Connecting to the node's control socket (`rad node connect`)~~
* [x] Printing out our routing table (`rad node routing`)
* [x] Printing out our tracking policies (`rad node tracking`)

FintanH:
Taking some notes and questions here while I explore this.

I'm trying to figure out how the state of the node is determined. From what I can see, a `Node` is just a wrapper around a UNIX socket. There's currently no "start" functionality, so is the `Node` started once we call the `new` constructor? Would that mean to `stop` the `Node`, we would first need to check the socket path exists and then construct the `Node` to call `shutdown`? 

Also looking at `routing` and `tracking`, it seems like there's no functionality at the service level (i.e. `service::Command` has no variant for this) that exists for getting that information yet. This would require wiring up two commands for getting the entries and serving them up, also adding functions to the `Handle` trait. Would that be correct?

FintanH:
Why is `Node` defined in `radicle::node` in the first place? It seems like the dependent are `radicle-cli`, `radicle-tools`, and `radicle-remote-helper`, without any internal use. `radicle-cli` already depends on `radicle-node` I believe, remote helper could depend on `radicle-node`, and the latter is not needed afaik. So would it make sense to move this `Handle` and `Node` stuff into `radicle-node` itself?

cloudhead:
> So would it make sense to move this `Handle` and `Node` stuff into `radicle-node` itself?

This would require `radicle-cli` to depend on `radicle-node`, which doesn't really make sense imo. It should only depend on the node "interface", not the node itself.

PR #346, opened 2023-02-21
node: improve seeds information

### LINKED ISSUE #343: Improve `Seeds` command
Currently, the `Seeds` command only returns connected seeds. This is not great, because:

1. If there are no connected seeds, it doesn't give the user the option to connect to an unconnected seed
2. If a seed is connected, but already being fetched from, it's returned anyway

We should return *all* seeds to make that command as useful as possible, but include some kind of "status", which can be either of: `[connected, fetching(rid), disconnected]`.

This will allow the CLI for eg. to make a better choice as to which seed to fetch from.

FintanH:
Previously, the seeds information would only consist of the connected NodeIds.

Improve on this by adding any disconnected and fetching seeds as well. A seed can either be disconnected, connected, or fetching -- where fetching implies connected.

To be able to see which NodeIds are being fetched from, it's necessary to add this information to the Session.

Fixes #343

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -27,7 +27,7 @@ use crate::crypto;
 use crate::crypto::{Signer, Verified};
 use crate::identity::{Doc, Id};
 use crate::node;
-use crate::node::{Address, Features, FetchResult};
+use crate::node::{Address, Features, FetchResult, Seed, Seeds};
 use crate::prelude::*;
 use crate::service::message::{Announcement, AnnouncementMessage, Ping};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
@@ -108,7 +108,7 @@ pub enum Command {
     /// Connect to node with the given address.
     Connect(NodeId, Address),
     /// Lookup seeds for the given repository in the routing table.
-    Seeds(Id, chan::Sender<Vec<NodeId>>),
+    Seeds(Id, chan::Sender<Seeds>),
     /// Fetch the given repository from the network.
     Fetch(Id, NodeId, chan::Sender<FetchResult>),
     /// Track the given repository.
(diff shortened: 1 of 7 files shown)


radicle/src/node.rs
```diff
@@ -180,6 +181,63 @@ impl Command {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
+#[serde(tag = "type", content = "id")]
```
cloudhead, line 185:
```suggestion
#[serde(tag = "state", content = "id")]
```
Would be my suggestion, since it's the state of the node session, not the type of node. I'd also make sure the tags are kebab-case, which we do everywhere else for the control socket json.

radicle-node/src/service.rs
```diff
@@ -443,16 +468,17 @@ where
...
                 );
-                resp.send(connected).ok();
+                resp.send(seeds).ok();
             }
             Command::Fetch(rid, seed, resp) => {
                 // TODO: Establish connections to unconnected seeds, and retry.
                 // TODO: Fetch requests should be queued and re-checked to see if they can
                 //       be fulfilled everytime a new node connects.
                 self.fetch_reqs.insert(rid, resp);
+                self.fetching_from.insert(seed);
```
cloudhead, line 481: You can do without this field entirely by checking the session state I think. Just check that the protocol is `Protocol::Fetch`. Perhaps you could add a `Session::is_fetching` method to simplify things.

radicle-node/src/service.rs
```diff
@@ -429,11 +432,33 @@ where
...
+                    disconnected: usize,
+                    fetching: usize,
+                }
+
+                let (stats, seeds) = match self.routing.get(&rid) {
+                    Ok(seeds) => seeds.into_iter().fold(
+                        (Stats::default(), Seeds::default()),
+                        |(mut stats, mut seeds), node| {
+                            if node != self.node_id() {
+                                if self.fetching_from.contains(&node) {
```
cloudhead, line 447: See note below about checking the session for this.

radicle-node/src/test/handle.rs
```diff
@@ -21,6 +21,7 @@ impl radicle::node::Handle for Handle {
     type Error = HandleError;
     type Sessions = service::Sessions;
     type FetchResult = FetchResult;
+    type Seeds = Seeds;
```
cloudhead, line 24: Any reason this is an associated type and not fixed for all implementors? I think we're using the same concrete type everywhere?

radicle-node/src/service.rs
FintanH: Ah, thanks! That makes more sense :grin:

radicle-node/src/test/handle.rs
FintanH: Ya, I initially thought it was the right approach because `FetchResult` was also parameterised. But to be honest, this whole `Handle` thing is a bit confusing for me. I've asked some questions there https://github.com/radicle-dev/heartwood/issues/271#issuecomment-1440484944.

I can change this to just be `Vec<Seed>` in the function, but would appreciate clearing up my questions in the other issue :blush:

cloudhead (approved):
LG

PR #347, opened 2023-02-21
[Hold] radicle-httpd: basic blob streaming support

keepsimple1:
This is to implement a basic blob streaming support mentioned in [this discussion](https://github.com/radicle-dev/heartwood/pull/331#discussion_r1110054271).

As `radicle-surf` does not concern itself with IO, nor async, I think it is better to support streaming in `heartwood` for blobs. @xphoniex .

Update:  after rebasing to #331 , I've updated `file_handler` to show the use of the streaming support.  See `BlobBody` in the patch.

### DIFF
--- a/radicle-httpd/Cargo.toml
+++ b/radicle-httpd/Cargo.toml
@@ -22,12 +22,14 @@ axum-server = { version = "0.4.4", default-features = false }
 chrono = { version = "0.4.22" }
 fastrand = { version = "1.7.0" }
 flate2 = { version = "1" }
+futures-core = { version = "0.3" }
 hyper = { version = "0.14.17", default-features = false }
 lexopt = { version = "0.2.1" }
 radicle-surf = { version = "0.9.0", default-features = false, features = ["serde"] }
 nonempty = { version = "0.8.1", features = ["serialize"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["preserve_order"] }
+tempfile = { version = "3.3.0" }
 thiserror = { version = "1" }
 time = { version = "0.3.17", features = ["parsing", "serde"] }
 tokio = { version = "1.21", default-features = false, features = ["macros", "rt-multi-thread"] }
@@ -41,6 +43,7 @@ path = "../radicle"
 version = "0.2.0"
 
 [dev-dependencies]
+futures-util = { version = "0.3" }
 hyper = { version = "0.14.17", default-features = false, features = ["client"] }
 pretty_assertions = { version = "1.3.0" }
 radicle-cli = { path = "../radicle-cli" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -856,6 +856,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "04909a7a7e4633ae6c4a9ab280aeb86da1236243a77b694a49eacd659a4bd3ac"
 
 [[package]]
+name = "futures-macro"
+version = "0.3.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bdfb8ce053d86b91919aad980c220b1fb8401a9394410e1c289ed7e66b61835d"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "futures-sink"
 version = "0.3.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -874,9 +885,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "197676987abd2f9cadff84926f410af1c183608d36641465df73ae8211dc65d6"
 dependencies = [
  "futures-core",
+ "futures-macro",
  "futures-task",
  "pin-project-lite",
  "pin-utils",
+ "slab",
 ]
 
 [[package]]
@@ -1967,6 +1980,8 @@ dependencies = [
  "chrono",
  "fastrand",
  "flate2",
+ "futures-core",
+ "futures-util",
  "hyper",
  "lexopt",
  "nonempty 0.8.1",
--- a/radicle-httpd/src/raw.rs
+++ b/radicle-httpd/src/raw.rs
@@ -1,5 +1,12 @@
+use core::pin::Pin;
+use core::task::{Context, Poll};
+use futures_core::Stream;
+use radicle_surf::blob::BlobRef;
+use std::fs::File;
+use std::io::{Read, Seek, Write};
 use std::sync::Arc;
 
+use axum::body::StreamBody;
 use axum::extract::State;
 use axum::http::header;
 use axum::response::IntoResponse;
@@ -10,7 +17,7 @@ use hyper::HeaderMap;
 use radicle::prelude::Id;
 use radicle::profile::Profile;
 use radicle::storage::git::paths;
-use radicle_surf::{Oid, Repository};
+use radicle_surf::{blob::Blob, Oid, Repository};
 
 use crate::axum_extra::Path;
 use crate::error::Error;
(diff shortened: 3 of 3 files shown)


radicle-httpd/src/api/json.rs
```diff
@@ -64,6 +71,53 @@ pub(crate) fn blob<T: AsRef<[u8]>>(blob: &Blob<T>, path: &str) -> Value {
...
+#[allow(dead_code)]
+pub(crate) fn blob_stream<P: AsRef<Path>>(
+    blob: &[u8],
+    tmp_path: P,
+    chunk_sz: usize,
+) -> Result<FileStream, crate::Error> {
+    let tmp_path = tmp_path.as_ref();
+    let mut write_f = File::create(tmp_path)?;
+    write_f.write_all(blob)?;
+    FileStream::new(tmp_path, chunk_sz)
```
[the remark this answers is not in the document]
xphoniex: (this applies to other fns/impls as well)

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -1310,4 +1310,34 @@ mod routes {
...
         );
     }
+
+    #[tokio::test]
+    async fn test_blob_stream() {
+        use crate::api::json::blob_stream;
+        use futures_util::StreamExt;
+
+        let tmp_dir = tempfile::tempdir().unwrap();
+        std::fs::create_dir_all(tmp_dir.path()).unwrap();
```
xphoniex, line 1320: is this needed? doesn't `tempfile::tempdir()` already create the dir for you?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -1310,4 +1310,34 @@ mod routes {
...
+        let tmp_file = tmp_dir.path().join("test_blob_stream");
+
+        let blob = "This is a test blob"; // 19 chars.
```
xphoniex, line 1323: is comment needed?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -1310,4 +1310,34 @@ mod routes {
             )
         );
     }
+
+    #[tokio::test]
+    async fn test_blob_stream() {
+        use crate::api::json::blob_stream;
+        use futures_util::StreamExt;
```
xphoniex, line 1317: can you move these `use` statements outside?

xphoniex (commented):
also prefix the branch name with your username please:
```
blob-stream` -> `keepsimple1/blob-stream
```

radicle-httpd/src/api/json.rs
```diff
@@ -64,6 +71,53 @@ pub(crate) fn blob<T: AsRef<[u8]>>(blob: &Blob<T>, path: &str) -> Value {
     response
 }
 
+/// Creates a [`FileStream`] from `blob`, using `tmp_path` as the file.
+#[allow(dead_code)]
```
FintanH, line 75: Why is this needed? :thinking:

radicle-httpd/src/api/json.rs
```diff
@@ -64,6 +71,53 @@ pub(crate) fn blob<T: AsRef<[u8]>>(blob: &Blob<T>, path: &str) -> Value {
...
+    let mut write_f = File::create(tmp_path)?;
+    write_f.write_all(blob)?;
+    FileStream::new(tmp_path, chunk_sz)
+}
+
+/// Represents a read stream of a file.
+pub(crate) struct FileStream {
+    f: File,
+    chunk_sz: usize,
+}
```
FintanH, line 91: I would have expected that you could have made a stream of the byte slice. Why are we opting to use a `File` here?

I'm pretty sure it's something along the lines of:
```rust
use futures::stream;

pub fn blob_stream(bs: &[u8]) -> impl Stream {
    stream::iter(bs)
}
```

The caller of `blob` can decide on their own chunking by calling [`chunks`](https://docs.rs/futures-util/0.3.26/futures_util/stream/trait.StreamExt.html#method.chunks).

radicle-httpd/src/api/v1/projects.rs
FintanH: You need to call `path` to get the `&Path` out. `tempdir` gives you back a type `TmpDir`.

keepsimple1:
> also prefix the branch name with your username please:
> 
> ```
> blob-stream` -> `keepsimple1/blob-stream
> ```

will do this in future.  (AFAIK, GitHub PR doesn't allow to change the source branch, unless creating a new PR).

radicle-httpd/Cargo.toml
```diff
@@ -47,3 +48,4 @@ radicle-cli = { path = "../radicle-cli" }
 radicle-crypto = { path = "../radicle-crypto" }
 tempfile = { version = "3.3.0" }
 tower = { version = "0.4", features = ["util"] }
+futures-util = { version = "0.3" }
```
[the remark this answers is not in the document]
keepsimple1: will do.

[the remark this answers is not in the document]
keepsimple1: Btw I thought `cargo fmt` would take care of it, but I was wrong ;-)

radicle-httpd/src/api/json.rs
```diff
@@ -1,5 +1,12 @@
 //! Utilities for building JSON responses of our API.
 
+use core::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+use futures_core::Stream;
```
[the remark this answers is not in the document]
keepsimple1: will do.

radicle-httpd/src/api/json.rs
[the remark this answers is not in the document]
keepsimple1: sure.

radicle-httpd/src/api/json.rs
```diff
@@ -64,6 +71,53 @@ pub(crate) fn blob<T: AsRef<[u8]>>(blob: &Blob<T>, path: &str) -> Value {
...
+) -> Result<FileStream, crate::Error> {
+    let tmp_path = tmp_path.as_ref();
+    let mut write_f = File::create(tmp_path)?;
+    write_f.write_all(blob)?;
+    FileStream::new(tmp_path, chunk_sz)
+}
+
+/// Represents a read stream of a file.
+pub(crate) struct FileStream {
+    f: File,
```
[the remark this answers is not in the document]
keepsimple1: no reasons ;-).  My "bad" habits from C programming ;-)

radicle-httpd/src/api/v1/projects.rs
keepsimple1: It's helpful for me to read my own test code ;-)

radicle-httpd/Cargo.toml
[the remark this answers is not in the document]
xphoniex: there's [`cargo-sort`](https://crates.io/crates/cargo-sort) but I haven't used it yet.

radicle-httpd/src/api/v1/projects.rs
keepsimple1: will do.

radicle-httpd/src/api/json.rs
keepsimple1: I haven't added a caller using this method, as I'm not familiar with how `httpd` wants to stream big files yet. Only the test case uses this in this PR.

radicle-httpd/src/api/json.rs
keepsimple1: The essence of this patch is to support streaming from a file. The [motivation](https://github.com/radicle-dev/heartwood/pull/331#discussion_r1110054271) was to store a big blobs into a tmp file so that we can reduce memory consumption. Then `httpd` would stream the blob from the tmp file.

radicle-httpd/src/api/v1/projects.rs
keepsimple1: I will double check.

radicle-httpd/src/api/json.rs
xphoniex: you can re-base on https://github.com/radicle-dev/heartwood/pull/331 and show how it works in practice.

radicle-httpd/src/api/json.rs
FintanH: I see. It _seems_ a bit weird that you have to have the bytes in memory to avoid them being in memory -- but perhaps it makes sense.

Regardless, I don't think we need to roll our own here. I believe you could use a combination of [`futures::io::BufReader`](https://docs.rs/futures/0.3.26/futures/io/struct.BufReader.html), [`futures::io::AsyncBufReadExt::lines`](https://docs.rs/futures/0.3.26/futures/io/trait.AsyncBufReadExt.html#method.lines), and [`futures::StreamExt::chunks`](https://docs.rs/futures-util/0.3.26/futures_util/stream/trait.StreamExt.html#method.chunks) to achieve this. We would create the file and pass it to `BufReader::new`.

Also, because you're using a temporary directory, be _very_ careful that you hold onto the `TempDir` value until the stream is finished, otherwise it will be dropped and the directory will be removed.

radicle-httpd/src/api/v1/projects.rs
keepsimple1: You are correct. No need to create the dir (again). I added it as for some reason the dir was not created. Maybe there were some accidental mistakes I didn't notice. Now it's good 👍 . Thanks.

radicle-httpd/src/api/json.rs
keepsimple1: > Also, because you're using a temporary directory, be _very_ careful that you hold onto the `TempDir` value until the stream is finished, otherwise it will be dropped and the directory will be removed.

yeah, I think this line holds a reference to the underlying `TempDir` in the test: 

```
let tmp_file = tmp_dir.path().join("test_blob_stream");
```

radicle-httpd/src/api/json.rs
keepsimple1: I've rebased on #331 and changed the target branch to `dave/add-raw-route`.

xphoniex: you haven't shown how this will be used in practice, inside [`file_handler`](https://github.com/radicle-dev/heartwood/commit/34b15f9f16d95de3f78e32ca4610280fdf16eb6c#diff-31099b31af3e5bf7f46f6dc4f2db59fdf633d2e65a3a06f48bfedbff35a29c9cR24).

keepsimple1: I think I did it [here](https://github.com/radicle-dev/heartwood/blob/446299ef832dce74500e3d5d0871e3efcf41f71d/radicle-httpd/src/raw.rs#L42-L43), no?

xphoniex: Hm.. this is not how I was envisioning it. As Fintan mentioned:
> It seems a bit weird that you have to have the bytes in memory to avoid them being in memory -- but perhaps it makes sense.

instead of creating a tmp file, you can write it to a fixed path, e.g. `/tmp/<rid>/<hash>/*path` and inside `file_handler`, first check for existence of the file before loading the blob.

- file exists? stream from disk.
- file doesn't exist? 
  - file greater than `BLOB_STREAM_MIN_BYTES` ? save to disk & then stream
  - file is small ? return directly
  
this way, we only create the file once, and if 5 different people try to download a 200MB binary, server only has allocated 200MB once, for a short amount of time. This however, leaves server open to exhausting disk space. So we need a manager for these caches, a simple FIFO cache with a pre-determined space (e.g. 5 GB) would be ok here.

xphoniex: On a second thought, an LRU cache is more appropriate here.

cloudhead:
I don't understand the purpose of this PR.. We're reading the file from disk, then copying it back to disk, then streaming it from disk? The file is already in memory, so we should stream it from memory, or just not handle files larger than it makes sense to copy.

xphoniex:
> 

By the time we know how big the file is, we have it in-memory. See my suggestion [here](https://github.com/radicle-dev/heartwood/pull/347#discussion_r1116587477).

cloudhead:
Having it in-memory is not the only issue: bandwidth is another problem, as that's what will usually cost money. I don't think implementing any form of caching makes sense: the file is already on disk as a blob, we shouldn't have to create a copy of everything we want to serve.

FintanH:
> I can add a size method on File [here](https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/src/fs.rs#L84), cause we can't use FileContent.

There's already a [`size`](https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/src/fs.rs#L148-L151) method on `FileContent`, so you can call `content` followed by `size` to get it.

To stream straight from disk, i.e. the git repository, we would probably need to bypass `git2` I would say and access the file system directly.

cloudhead:
Yeah, I wouldn't bother bypassing git2. Simple solution is to just return an error if the file is too large, and otherwise copy it as we currently do.

radicle-httpd/src/api/json.rs
keepsimple1: In that case, maybe we can go further and read / stream directly from the object file in git tree ?

xphoniex: that seems to have its own complications, you can put this PR on hold for now.

keepsimple1:
If we want to continue to explore, there are a couple of things we could try: 

1) stream the object file to the client, and let the client to process (i.e. decompress, remove its header).  
   - Pros: read directly from the disk, zero copy, minimum bandwidth
   - Cons: extra work in the client (front-end) code.

or 

2) store `repo` reference in the context / profile, so that `file_handler` can return a reference `&[u8]`. I also mentioned it [here ](https://github.com/radicle-dev/heartwood/pull/331#discussion_r1108759193) early.
   - Pros: read into memory once via `git2`. No additional copy or write to disk in our code.
   - Cons: Need to check: I hope `git2` won't copy the same object multiple times for multiple requests?  Will storing `repo` reference in the profile complicate other parts of `httpd`?

radicle-httpd/src/api/json.rs
keepsimple1: Is there a way to set the PR state to `hold` officially?

FintanH: > Is there a way to set the PR state to hold officially?

You can turn it back into a draft

keepsimple1:
As [discussed on zulip](https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/.E2.9C.94.20follow.20up.20on.20PR.20.23347/near/345026804),  I'm closing this PR for now.  Thanks!

PR #354, opened 2023-02-24
node: improve fetch

### LINKED ISSUE #317: Fetch specific remotes, not all
Currently when we fetch, we always fetch all remotes (`Namespaces::All`). There are two issues with that:

1. We fetch our own namespace from other peers, which can overwrite our own refs
2. We fetch refs we're not interested in

We should instead construct a fetch-spec that only includes the remotes (namespaces) we're interested in.

The fetch code is in `worker.rs`, though it might be required to modify the `Namespaces` type to include a `Many` variant, for >1 remotes.

Note that there is one caveat to the above: when we're doing a full clone, we currently don't know what remotes exist, and so we're likely going to fetch all remotes. This is a related but separate problem that needs addressing.

FintanH:
The fetch implementation will potentially fetch the local operator's remote, which is unnecessary and could result in deleted data.

Fix this by threading the local operator's public key and add an ignore refspec to the fetch arguments, i.e.

    ^refs/namespaces/<local>/*

The implementation also strictly fetches all or one namespace. This is improved upon by introducing a new variant for specifying a set of namespaces.

Fixes #317

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -33,6 +33,7 @@ use crate::prelude::*;
 use crate::service::message::{Announcement, AnnouncementMessage, Ping};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
 use crate::service::reactor::FetchDirection;
+use crate::service::tracking::Scope;
 use crate::storage;
 use crate::storage::{Inventory, ReadRepository, RefUpdate, WriteStorage};
 use crate::storage::{Namespaces, ReadStorage};
@@ -116,7 +117,7 @@ pub enum Command {
     /// Fetch the given repository from the network.
     Fetch(Id, NodeId, chan::Sender<FetchResult>),
     /// Track the given repository.
-    TrackRepo(Id, tracking::Scope, chan::Sender<bool>),
+    TrackRepo(Id, Scope, chan::Sender<bool>),
     /// Untrack the given repository.
     UntrackRepo(Id, chan::Sender<bool>),
     /// Track the given node.
@@ -267,7 +268,7 @@ where
 
     /// Track a repository.
     /// Returns whether or not the tracking policy was updated.
-    pub fn track_repo(&mut self, id: &Id, scope: tracking::Scope) -> Result<bool, tracking::Error> {
+    pub fn track_repo(&mut self, id: &Id, scope: Scope) -> Result<bool, tracking::Error> {
         self.out_of_sync = self.tracking.track_repo(id, scope)?;
         self.filter.insert(id);
 
@@ -1093,13 +1094,18 @@ where
                 }
                 debug!(target: "service", "Fetch accepted for {rid} from {remote}..");
 
-                let namespaces = if let Ok(_repo) = self.storage.repository(rid) {
-                    // FIXME(finto): using remotes breaks test_gossip_during_fetch, so we use default for now
-                    Namespaces::default()
-                } else {
-                    Namespaces::default()
+                let namespaces = match self.tracking.namespaces_for(&self.storage, &rid) {
+                    Ok(ns) => ns,
+                    Err(err) => {
+                        if let Some(resp) = self.fetch_reqs.get(&rid) {
+                            resp.send(FetchResult::Failed {
+                                reason: err.to_string(),
+                            })
+                            .ok();
+                        }
+                        return Ok(());
+                    }
                 };
-
                 // Instruct the transport to handover the socket to the worker.
                 self.reactor
                     .fetch(peer, rid, FetchDirection::Initiator { namespaces });
@@ -1204,7 +1210,7 @@ where
             Namespaces::Many(pks) => {
                 for remote_id in pks.into_iter() {
                     if refs
-                        .push((*remote_id, repo.remote(&remote_id)?.refs.unverified()))
+                        .push((*remote_id, repo.remote(remote_id)?.refs.unverified()))
                         .is_err()
                     {
                         warn!(
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -116,7 +116,7 @@ pub enum Command {
     /// Fetch the given repository from the network.
     Fetch(Id, NodeId, chan::Sender<FetchResult>),
     /// Track the given repository.
-    TrackRepo(Id, chan::Sender<bool>),
+    TrackRepo(Id, tracking::Scope, chan::Sender<bool>),
     /// Untrack the given repository.
     UntrackRepo(Id, chan::Sender<bool>),
     /// Track the given node.
@@ -136,7 +136,7 @@ impl fmt::Debug for Command {
             Self::Connect(id, addr) => write!(f, "Connect({id}, {addr})"),
             Self::Seeds(id, _) => write!(f, "Seeds({id})"),
             Self::Fetch(id, node, _) => write!(f, "Fetch({id}, {node})"),
-            Self::TrackRepo(id, _) => write!(f, "TrackRepo({id})"),
+            Self::TrackRepo(id, scope, _) => write!(f, "TrackRepo({id}, {scope})"),
             Self::UntrackRepo(id, _) => write!(f, "UntrackRepo({id})"),
             Self::TrackNode(id, _, _) => write!(f, "TrackNode({id})"),
             Self::UntrackNode(id, _) => write!(f, "UntrackNode({id})"),
@@ -491,9 +491,9 @@ where
                 self.fetch_reqs.insert(rid, resp);
                 self.fetch(rid, &seed);
             }
-            Command::TrackRepo(rid, resp) => {
+            Command::TrackRepo(rid, scope, resp) => {
                 let tracked = self
-                    .track_repo(&rid, tracking::Scope::All)
+                    .track_repo(&rid, scope)
                     .expect("Service::command: error tracking repository");
                 // TODO: Try to fetch project if we weren't tracking it before.
                 resp.send(tracked).ok();
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -1072,6 +1072,7 @@ where
                         return Ok(());
                     }
                 }
+
                 // Accept the request and instruct the transport to handover the socket to the worker.
                 self.reactor.write(peer, Message::FetchOk { rid });
                 self.reactor.fetch(peer, rid, FetchDirection::Responder);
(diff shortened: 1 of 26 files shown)


FintanH:
@cloudhead I was wondering if you could help point me to how `Namespaces` thing is being used and also where the fetch is being tested? :)

cloudhead:
Nice!

Fetch is mostly tested inside the file `e2e.rs` (`test_clone`, `test_replication`). Some of the CLI tests also test fetching, by calling `rad clone` (`radicle_cli::tests::commands::rad_clone`).

`Namespaces` is mostly used here, in `service.rs`:
```rust
            (session::State::Connected { .. }, Message::Fetch { rid }) => {
                debug!(target: "service", "Fetch requested for {rid} from {remote}..");

                // Accept the request and instruct the transport to handover the socket to the worker.
                self.reactor.write(peer, Message::FetchOk { rid });
                self.reactor.fetch(peer, rid, Namespaces::default(), false); // <- HERE
            }
```

We should be thinking about when (in what scenarios) we want to fetch all namespaces vs. only a select few. Most likely we should be fetching the id branches first, getting the delegate set, then fetching the delegates. But that's probably outside the scope of this PR, as it changes the fetch logic quite a bit and is non-trivial.

radicle-node/src/worker.rs
```diff
@@ -191,15 +192,22 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
...
         }
 
         if self.atomic {
             // Enable atomic fetch. Only works with Git 2.31 and later.
             cmd.arg("--atomic");
         }
+
+        // Ignore our own remote when fetching
+        let mut fetchspecs = fetch.namespaces.as_fetchspecs();
+        fetchspecs.push(format!("^refs/namespaces/{}/*", self.local));
```
cloudhead, line 207: How about adding a `Namespaces::Not(PubKey)` variant to help here?

FintanH: It _feels_ out of place, since some of the logic using `Namespaces` cares about whom it's fetching from. For example, whether to add the `--prune` flag.

I could add another struct though. Something like:
```rust
pub struct Exclude(NonEmpty<PublicKey>);

impl Exclude {
    pub fn as_fetchspecs(&self) -> Vec<String> {
        /* code here */
    }
}
```

cloudhead: Hmm true, maybe something to think about for `git-ref-format` at some point.

radicle-node/Cargo.toml
```diff
@@ -36,6 +36,11 @@ snapbox = { version = "0.4.3", optional = true }
 tempfile = { version = "3.3.0" }
 thiserror = { version = "1" }
 
+[dependencies.git2]
+version = "0.16.1"
+default-features = false
+features = ["vendored-libgit2"]
```
cloudhead, line 42: I think you can access this via `radicle::git::raw` if I remember. Better than re-importing due to the extra working when updating versions.

radicle-node/src/service/tracking.rs
```diff
@@ -14,43 +14,73 @@ pub use store::Error;
...
     pub fn is_repo_tracked(&self, id: &Id) -> Result<bool, Error> {
-        self.repo_policy(id).map(|policy| policy == Policy::Track)
+        self.store.is_repo_tracked(id)
     }
 
     /// Check if a node is tracked.
     pub fn is_node_tracked(&self, id: &NodeId) -> Result<bool, Error> {
-        self.node_policy(id).map(|policy| policy == Policy::Track)
+        self.store.is_node_tracked(id)
     }
```
cloudhead, line 58: Hmmmmmm these functions no longer make use of the default policies in case something isn't in the db.. is that intentional?

radicle-node/src/service/tracking.rs
```diff
@@ -14,43 +14,73 @@ pub use store::Error;
...
+    ///
+    /// Otherwise, if the `id` does not exist, and the default
+    /// [`Self::scope`] is [`Scope::All`] it returns `true`.
+    ///
+    /// It returns `false` if neither of these is the case.
+    pub fn should_fetch(&self, id: &Id) -> Result<bool, Error> {
+        Ok(match self.store.repo_entry(id)? {
+            Some(entry) => entry.policy == Policy::Track,
+            None => matches!(self.scope, Scope::All),
+        })
```
cloudhead, line 47: Not sure this makes sense, the scope is related to the nodes.. Instead of this function you simply want the implementation of `is_repo_tracked` without the changes in this diff.

radicle/src/node/tracking.rs
```diff
@@ -29,9 +29,9 @@ pub type Alias = String;
 #[derive(Default, Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub enum Policy {
     /// The resource is tracked.
+    #[default]
     Track,
     /// The resource is blocked.
-    #[default]
     Block,
 }
```
cloudhead, line 36: Wut

radicle/src/storage.rs
```diff
@@ -16,8 +17,8 @@ pub use radicle_git_ext::Oid;
 use crate::collections::HashMap;
 use crate::git::ext as git_ext;
 use crate::git::{Qualified, RefError, RefString};
-use crate::identity;
 use crate::identity::doc::DocError;
+use crate::identity::{self, Did};
```
cloudhead, line 21: :grimacing:

radicle/src/storage.rs
```diff
@@ -34,13 +35,38 @@ pub enum Namespaces {
...
+    pub fn delegates<R>(repo: &R) -> Result<Self, IdentityError>
+    where
+        R: ReadRepository,
+    {
+        Ok(Self::Many(repo.delegates()?.map(PublicKey::from)))
+    }
+
+    // TODO(finto): Should we consider forcing if not self
+    // One bad case is force fetching state that is behind
+    // But how do we fetch force fetched?
```
cloudhead, line 59: Maybe create an issue for this instead, I think it merits a conversation.

radicle/src/test/storage.rs
```diff
@@ -55,28 +55,42 @@ impl ReadStorage for MockStorage {
         Ok(self.inventory.keys().cloned().collect::<Vec<_>>())
     }
 
-    fn repository(&self, _proj: Id) -> Result<Self::Repository, Error> {
-        Ok(MockRepository {})
+    fn repository(&self, rid: Id) -> Result<Self::Repository, Error> {
+        let doc = self
+            .inventory
+            .get(&rid)
+            .expect("Mockstorage::get: missing doc");
```
cloudhead, line 62:
```suggestion
            .expect("MockStorage::repository: missing doc");
```

radicle/src/test/storage.rs
```diff
@@ -55,28 +55,42 @@ impl ReadStorage for MockStorage {
...
+        let doc = self.inventory.get(&rid).unwrap();
+        Ok(MockRepository {
+            id: rid,
+            doc: doc.clone(),
+        })
     }
 
     fn create(&self, _rid: Id) -> Result<Self::RepositoryMut, Error> {
-        Ok(MockRepository {})
+        todo!()
```
cloudhead, line 82: :eyes:

FintanH: Hahaha, ya it didn't break anything so I'm wondering what it was used for?

cloudhead: Maybe not used, but the previous implementation was ok no?

radicle-node/src/service/tracking.rs
cloudhead: So, your tracking policy is a combination of the entries in the tracking db, plus your default policy. So the idea is that if there's no entry in the db, we use whatever your default, or "fallback" policy is.

Example for repos:

```
* default=track, db=None    -> track
* default=block, db=None    -> block
* default=track, db=block     -> block
* default=block, db=track     -> track 
```

That's for repositories, now for the cases where we track, we can additional check the tracking scope, to see if a specific remote should be fetched:

```
* repo=tracked, scope=all         -> fetch the remote
* repo=tracked, scope=trusted  -> check if node is tracked in db
```

radicle/src/node/tracking.rs
FintanH: This relates to the default `Scope` being `Trusted`. So if we take these changes in tandem then we're saying, "the default behaviour is to always track and replicate from trusted nodes". Seed nodes would then likely use `Scope::All` with `Policy::Track`. Blocking is an explicit action of saying I definitely want to block this node. Or if it's something like a personal server you would say `Scope::All` and `Policy::Block` and explicitly track your own devices or peers.

I would say that the default presented here is the most common case and each case, in turn, is less common than the previous. 

Does that make sense?

radicle-node/Cargo.toml
FintanH: Actually, this can be removed because it was for the `commit` function. Thanks for the tip, regardless!

radicle/src/test/storage.rs
FintanH: I changed `MockRepository` to have `id` and `doc` and I have no way of producing a `doc` here :thinking:

FintanH: I guess in the real world, the idea is that an `Identity` is created, giving you the `Id` to use here. But creating the repository itself is not inherently linked to the creating of the identity -- only through being given the `Id`.

radicle-node/src/service/tracking.rs
FintanH: Let me digest and play with this to understand a bit better :)

FintanH: Ok, I changed those functions back. I think its use in [`is_tracking`](https://github.com/radicle-dev/heartwood/blob/master/radicle-node/src/service.rs) is a bit of a lie though, in a way, but not a big deal for this change set. I just found it confusing that it would say `true` if I didn't actually track anything -- which was causing me issues when trying to figure out test failures. However, with this current set up proposed here now everything works at each commit :)

cloudhead: Yeah, I agree with that actually. Happy to try to make the API clearer by distinguishing actual tracking with a "Track" policy.

radicle-node/src/service.rs
```diff
@@ -1110,8 +1113,68 @@ where
...
+                                            "Attempted to fetch with no trusted peers",
+                                        ),
+                                    })
+                                    .ok();
+                                }
+                                return Ok(());
+                            }
+                        }
+                    },
+                };
```
cloudhead, line 1174: This block contains a few `unwraps` that we should document with `// SAFETY`, turn to expects, or convert to errors

FintanH: I extracted this big ass block out into a function that returns the `Namespaces` or a `FetchResult` (really the failure cases) and use that function instead.

radicle-node/src/service.rs
```diff
@@ -1139,6 +1152,56 @@ where
         )
     }
 
+    // XXX: does not take &self to appease the borrow checker
```
cloudhead, line 1155: Hmm, should it be on `tracking::Config` maybe? `Config::namespaces_for(rid, storage)`

FintanH: Ya, that sounds good! Nice suggestion :)

FintanH: Fixed that up and rebased with the `FetchDirection` changes :)

cloudhead (approved):
:rocket:

PR #355, opened 2023-02-24
radicle-httpd: Group issue and patch states in project endpoint

sebastinez:
To allow the web client to show the amounts of issues and patches in their different states without fetching them all, I propose that this should be done by the radicle-httpd.

There is probably a more optimized way of doing this, feel free to suggest optimizations!

The idea is to have a JSON output like the following:

```diff
[
  {
    "name": "test-heartwood",
    "description": "test-description",
    "defaultBranch": "master",
    "delegates": [
      "did:key:z6MkkfM3tPXNPrPevKr3uSiQtHPuwnNhu2yUVjgd2jXVsVz5"
    ],
    "head": "1dc55cc08ab5f72c1ad552effb115665bde03990",
    "patches": {
+     "proposed": 0,
+     "draft": 0,
+     "archived": 0
    },
    "issues": {
+    "open": 11,
+    "closed": 0
    },
    "id": "rad:z22nAstJ662ouwRwADTpuzB6R9hVq"
  }
]
```

### DIFF
--- a/radicle-httpd/src/api.rs
+++ b/radicle-httpd/src/api.rs
@@ -51,8 +51,8 @@ impl Context {
         let doc = repo.identity_of(self.profile.id())?;
         let payload = doc.project()?;
         let delegates = doc.delegates;
-        let issues = (Issues::open(self.profile.public_key, &repo)?).count()?;
-        let patches = (Patches::open(self.profile.public_key, &repo)?).count()?;
+        let issues = Issues::open(self.profile.public_key, &repo)?.counts()?;
+        let patches = Patches::open(self.profile.public_key, &repo)?.counts()?;
 
         Ok(project::Info {
             payload,
@@ -143,6 +143,7 @@ mod project {
     use nonempty::NonEmpty;
     use serde::Serialize;
 
+    use radicle::cob;
     use radicle::git::Oid;
     use radicle::identity::project::Project;
     use radicle::identity::Id;
@@ -157,8 +158,8 @@ mod project {
         pub payload: Project,
         pub delegates: NonEmpty<Did>,
         pub head: Oid,
-        pub patches: usize,
-        pub issues: usize,
+        pub patches: cob::patch::PatchCounts,
+        pub issues: cob::issue::IssueCounts,
         pub id: Id,
     }
 }
--- a/radicle-httpd/src/api/v1/delegates.rs
+++ b/radicle-httpd/src/api/v1/delegates.rs
@@ -49,9 +49,9 @@ async fn delegates_projects_handler(
             }
 
             let Ok(issues) = Issues::open(ctx.profile.public_key, &repo) else { return None };
-            let Ok(issues) = (*issues).count() else { return None };
+            let Ok(issues) = issues.counts() else { return None };
             let Ok(patches) = Patches::open(ctx.profile.public_key, &repo) else { return None };
-            let Ok(patches) = (*patches).count() else { return None };
+            let Ok(patches) = patches.counts() else { return None };
 
             Some(Info {
                 payload,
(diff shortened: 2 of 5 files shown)


radicle-httpd/src/api.rs
```diff
@@ -148,6 +174,23 @@ mod project {
     use radicle::identity::Id;
     use radicle::prelude::Did;
 
+    /// Detailed information on issue states
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct IssueState {
+        pub open: usize,
+        pub closed: usize,
+    }
```
FintanH, line 182: I'd suggest creating methods to construct these from the `Issues` and `Patches` types.

FintanH (commented):
I think the most significant optimization will be the httpd code keeping some cache for the COB states :)

The reason is that the COBs, when loaded, are read from disk every time. That's a bigger topic though :sweat_smile:

radicle-httpd/src/api.rs
```diff
@@ -51,15 +51,40 @@ impl Context {
...
+            .collect();
+        let proposed_patches: Vec<_> = patches
+            .clone()
+            .into_iter()
+            .filter(|(_, patch, _)| patch.state() == patch::State::Proposed)
+            .collect();
+        let archived_patches: Vec<_> = patches
+            .into_iter()
+            .filter(|(_, patch, _)| patch.state() == patch::State::Archived)
+            .collect();
```
FintanH, line 73: Can do the following (and the same for issues)
```suggestion
        let patch_stats = patches
            .into_iter()
            .fold(PatchState::default(), |mut state, p| {
                match p.state() {
                    patch::State::Draft => state.draft += 1,
                    patch::State::Proposed => state.proposed += 1,
                    patch::State::Archived => state.archived += 1,
                }
                state
            });
```

sebastinez: TIL about `fold` very nice!
Addressed here 3683cdd

FintanH: Nice :) The fold logic is duplicated across the 3 handlers, hence why I think there could a method to de-duplicate it and just call the method in the 3 places.

radicle-httpd/src/api.rs
sebastinez: Done here ca3c21b967e60f644321f4061d8f9ab5ce95bff7

radicle-httpd/src/api.rs
```diff
@@ -51,8 +51,29 @@ impl Context {
...
+        let issues = issues.filter_map(|s| s.ok()).fold(
+            Issues::default_state_count(),
+            |mut state, (_, i, _)| {
+                match i.state() {
+                    issue::State::Open => state.open += 1,
+                    issue::State::Closed { .. } => state.closed += 1,
+                }
+                state
+            },
+        );
```
FintanH, line 64: I actually meant moving this logic into a method. The `default` was fine :smile:

sebastinez: ok nice, great idea.
Done in 54d489b

radicle-httpd/src/api/v1/delegates.rs
```diff
@@ -48,10 +48,10 @@ async fn delegates_projects_handler(
...
             }
 
-            let Ok(issues) = Issues::open(ctx.profile.public_key, &repo) else { return None };
-            let Ok(issues) = (*issues).count() else { return None };
-            let Ok(patches) = Patches::open(ctx.profile.public_key, &repo) else { return None };
-            let Ok(patches) = (*patches).count() else { return None };
+            let Ok(issues) = Issues::open(ctx.profile.public_key, &repo) else { return None; };
+            let Ok(issues) = issues.group_by_state() else { return None };
+            let Ok(patches) = Patches::open(ctx.profile.public_key, &repo) else { return None; };
+            let Ok(patches) = patches.group_by_state() else { return None; };
```
xphoniex, line 54: why first `return None` doesn't have `;`? tbh I prefer without for all cases. which also applies to changes in [radicle-httpd/src/api/v1/projects.rs](https://github.com/radicle-dev/heartwood/blob/eebb459f4990440471783f031982e517ef04a65d/radicle-httpd/src/api/v1/projects.rs).

xphoniex (approved):
lgtm otherwise

radicle-httpd/src/api/v1/delegates.rs
sebastinez: Hey @xphoniex nice catch, yeah I think I'll remove the redundant semicolons

sebastinez: 20773cf solves this

FintanH (approved):
lgtm :)

cloudhead (approved):
LGTM

PR #356, opened 2023-02-24
httpd: Set a max allowed size of 4 MiB when serving files raw

xphoniex:


radicle-httpd/src/raw.rs
```diff
@@ -27,12 +29,16 @@ async fn file_handler(
 ) -> impl IntoResponse {
     let storage = &profile.storage;
     let repo = Repository::open(paths::repository(storage, &project))?;
-    let blob = repo.blob(sha, &path)?;
-
     let mut response_headers = HeaderMap::new();
+
+    if repo.file(sha, &path)?.content(&repo)?.size() > MAX_BLOB_SIZE {
```
xphoniex, line 34: wasn't able to find in the source whether calling `.content(..)` allocates memory for the content itself (which would defeat the purpose here).

I suppose it's just a light pointer, but need your confirmation on this @FintanH

cloudhead: I think it must load it in memory to know its size, since it's not necessarily a file. So probably best to only load it once.

radicle-httpd/src/raw.rs
```diff
@@ -27,12 +29,16 @@ async fn file_handler(
...
+        return Ok::<_, Error>((StatusCode::PAYLOAD_TOO_LARGE, response_headers, vec![]));
+    }
+
+    let blob = repo.blob(sha, &path)?;
     response_headers.insert(header::CONTENT_TYPE, "text; charset=utf-8".parse().unwrap());
```
cloudhead, line 39: It might be a good idea to cover a few of the common mime types, eg. those in https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types and fallback on `text`

radicle-httpd/src/raw.rs
xphoniex: if it's loaded in memory, can't we just get metadata of blob from disk using `fs::`?

> So probably best to only load it once.

wdym? you want it cached?

radicle-httpd/src/raw.rs
xphoniex: `text` works as expected on both chrome and firefox. browser ignores the mime and shows text, renders images, etc and on binaries starts a download.

otherwise, I'd have to add `mime_guess` crate, which I thought wouldn't be worth increasing the tree, if doesn't add any functionality.

radicle-httpd/src/raw.rs
xphoniex: to give you a corner case: image there's a 1 GB binary on server and someone tries to download that, this would lead to OOM error on server, regardless of how many times we try to download it. 

now think of a smaller size binary, it can still be used to DDoS server, just increase the number of concurrent requests.

there's really no point in allocating but then not serving based on size, because serving part is not expensive, allocating is.

xphoniex: if it ends up loading the content in order to know its size, then I suggest we get metadata of blob from disk using `fs::`.

to give you a corner case: image there's a 1 GB binary on server and someone tries to download that, this would lead to OOM error on a server (with 1GB of ram), regardless of how many times we load it.

now think of a smaller size binary, it can still be used to DDoS server, just increase the number of concurrent requests.

there's really no point in allocating but then not serving based on size, because serving part is not expensive, allocating is.

cloudhead: We can't get the metadata of the blob from disk. A blob is not necessarily an on-disk file.

radicle-httpd/src/raw.rs
cloudhead: Just because text works on a couple of browsers, doesn't mean it's the correct content-type..

There's no need for the `mime_guess` crate, you just have a map `file_extension -> mime_type`.

radicle-httpd/src/raw.rs
cloudhead: > to give you a corner case: image there's a 1 GB binary on server and someone tries to download that, this would lead to OOM error on a server (with 1GB of ram), regardless of how many times we load it.

This is best handled by just not hosting repos with 1 GB binaries in them, since you'll have the same problem at the p2p layer.

FintanH: Digging into what is called by `Blob::size`, we can see it's these two calls [blob.c](https://github.com/libgit2/libgit2/blob/80742e159c3221fd853d9a0fe735b382a7e3419d/src/libgit2/blob.c#L29-L37) & [odb.c](https://github.com/libgit2/libgit2/blob/6f2eedb7eb58f5af2e96cafeae7261c07c928d89/src/libgit2/odb.c#L183-L186). It _seems_ like it doesn't need to load the data into memory to determine the size, so this should be fine.

cloudhead: But we're doing `content().size()` above..?

FintanH: That actually turns the `Oid` in the `File` into the corresponding `git2::Blob`, which is just a pointer object.

cloudhead: `git2::Blob` loads the blob from disk though.

FintanH: I don't _think_ it does judging by the [struct internals](https://docs.rs/git2/0.16.1/src/git2/blob.rs.html#12-15). Why do you think it does?

cloudhead: The underlying `git_blob` type has a data pointer: https://github.com/libgit2/libgit2/blob/8a871d13b7f4e186b8ad943ae5a7fcf30be52e67/src/libgit2/blob.h#L17

FintanH: Hmmm it _could_ be the odb pointer instead https://github.com/libgit2/libgit2/blob/8a871d13b7f4e186b8ad943ae5a7fcf30be52e67/src/libgit2/blob.h#L21 since it's a `union`. I'm not adept enough at C to track when it actually changes to the data pointer though

cloudhead: Yeah could question... I don't know in which cases the odb pointer is used, and how that all works.. The only way it would *not* have to preload the blob in memory is if the odb is mmaped I guess..

radicle-httpd/src/raw.rs
```diff
@@ -27,12 +29,85 @@ async fn file_handler(
...
+                "aac" => "audio/aac",
+                "avi" => "video/x-msvideo",
+                "bin" => "application/octet-stream",
+                "bmp" => "image/bmp",
+                "bz" => "application/x-bzip",
+                "bz2" => "application/x-bzip2",
+                "csh" => "application/x-csh",
+                "css" => "text/css",
+                "csv" => "text/csv",
+                "doc" => "application/msword",
```
xphoniex, line 51: list taken from mozilla [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)

we can create a static map but needs [phf](https://crates.io/crates/phf), and if we're adding a crate, then I suggest `mime_guess`

xphoniex: actually, made it static [here](https://github.com/radicle-dev/heartwood/compare/a4ddfee881c5c76c07c473501a605d1d545580ad..08b132b5341b93953ab0ed91c0d2c84b02bfa38b) without importing anything

PR #357, opened 2023-02-27
cob: Use hashes for operation ids 

cloudhead:


radicle-cob/src/backend/git/change.rs
```diff
@@ -110,8 +110,15 @@ impl change::Storage for git2::Repository {
...
-        let revision = write_manifest(self, &manifest, &contents)?;
+        let (revision, blob_oids) = write_manifest(self, &manifest, &contents)?;
         let tree = self.find_tree(revision)?;
+        let contents = blob_oids
+            .into_iter()
+            .zip(contents)
+            .map(|(oid, op)| (oid.into(), op))
+            .collect::<Vec<_>>();
+        // SAFETY: We know it's not empty because it was created from a `NonEmpty`.
+        let contents = NonEmpty::from_vec(contents).unwrap();
```
FintanH, line 121:
```suggestion
        // SAFETY: We know it's not empty because it was created from a `NonEmpty`.
        let contents = NonEmpty::collect(blob_oids
            .into_iter()
            .zip(contents)
            .map(|(oid, op)| (oid.into(), op))).unwrap();
```

radicle-cob/src/history/entry.rs
```diff
@@ -11,7 +11,7 @@ use crate::pruning_fold;
 
 /// Entry contents.
 /// This is the change payload.
-pub type Contents = NonEmpty<Vec<u8>>;
+pub type Contents = NonEmpty<(Oid, Vec<u8>)>;
```
FintanH, line 14: I think it might be nice to make the `(Oid, Vec<u8>)` a `struct`. Field names would be nice for telling what it is and then also useful for using the `Entry` in the `traverse`, i.e. not using `.1` to get the inner contents.

FintanH (commented):
I can't seem to find where we test this scenario of `[revision, revision, merge] === [revision, merge, revision]`. Did you add it and if so could you point it out to me? :)

radicle-cob/src/history/entry.rs
cloudhead: Yeah true.. maybe `EncodedOp` :thinking: 
Also need to change the comment, because that's actually only used on the "read" side, since the Oid isn't actually stored on disk.

cloudhead:
> I can't seem to find where we test this scenario of `[revision, revision, merge] === [revision, merge, revision]`. Did you add it and if so could you point it out to me? :)

Ah, I added a different set of tests here: https://github.com/radicle-dev/heartwood/blob/af6ffce7de983f5dec980a2bb4a6233e82d5e62c/radicle/src/cob/thread.rs#L632-L692

I can add a more specific test in `patch` too though.

FintanH:
> Ah, I added a different set of tests here: 

Ya, I don't think that captures the mutating phenomenon since all the comments are the same.

radicle-cli/src/commands/id.rs
```diff
@@ -496,9 +504,17 @@ fn commit_select<'a>(
...
+                let (id, revision) =
+                    term::proposal::revision_commit_select(proposal, previous).unwrap();
+                (*id, revision)
+            } else {
+                let (id, revision) = proposal
+                    .revisions()
+                    .find(|(_, r)| r.is_quorum_reached(previous))
+                    .ok_or(anyhow!("No revisions with quorum found"))?;
+                (*id, revision)
+            }
```
slack-coder, line 517: maybe `match (id, interactive)`?

FintanH (approved):
I think we'd be safer to do the same thing for the identity proposal revisions as we do for the patch revisions :)

FintanH (approved):

cloudhead:
All done.

I made the revision handling change to identity, patch and thread COBs, and added some tests. I *think* we may want to move to an `LWWMap` eventually, but need to explore this.

I also created a `EntryBlob` type to hold the oid/data.

PR #360, opened 2023-02-27
cli: Don't fail clone if we have repo locally

### LINKED ISSUE #311: `rad-clone`: Don't create fork if we already have a fork
Currently we don't check, so it probably fails if we're cloning something for the second time.

### LINKED ISSUE #338: `rad clone` fails in a singe-node scenario with `failed no seeds found`
https://github.com/radicle-dev/heartwood/commit/e584bd6c1cbb5c8bb9b74a856f6fdd7c22392efe: If you push a project to your node and then try to clone it from same profile/node, you'll encounter:

```bash
== Clone failed no seeds found for rad:z5mjpEsLp9yLYyD5gez85eamUEmk
```

this problem doesn't exist in a multi-node setup.

cloudhead:
Fixes #338
Fixes #311

### DIFF
--- a/radicle-cli/src/commands/clone.rs
+++ b/radicle-cli/src/commands/clone.rs
@@ -175,27 +175,32 @@ pub fn clone<G: Signer>(
 
     // Get seeds. This consults the local routing table only.
     let mut seeds = node.seeds(id)?;
-    if !seeds.has_connections() {
-        return Err(CloneError::NotFound(id));
-    }
-    // Fetch from all seeds.
-    for seed in seeds.connected() {
-        let spinner = term::spinner(format!(
-            "Fetching {} from {}..",
-            term::format::tertiary(id),
-            term::format::tertiary(term::format::node(seed))
-        ));
+    if seeds.has_connections() {
+        // Fetch from all seeds.
+        for seed in seeds.connected() {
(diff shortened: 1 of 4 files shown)


FintanH (approved):
Makes sense :+1:

radicle-cli/src/commands/clone.rs
```diff
@@ -175,31 +175,36 @@ pub fn clone<G: Signer>(
...
+            match node.fetch(id, *seed)? {
+                FetchResult::Success { .. } => {
+                    spinner.finish();
+                }
+                FetchResult::Failed { reason } => {
+                    spinner.error(reason);
+                }
             }
         }
+        // TODO: Warn if no seeds were found, and we might be checking out stale data.
```
xphoniex, line 198: might also be helpful to assign someone to TODOs in comments as well, e.g. `TODO(dave)`

xphoniex (approved):
Looks good, tested locally and it works.

radicle-cli/src/commands/clone.rs
cloudhead: Yeah we can start doing that.

xphoniex: works well with `grep`, and even ci/bots in future.

PR #363, opened 2023-02-28
Fix a bunch of small issues

### LINKED ISSUE #314: `rad .` requires key unlock
If you tun `rad .` or `rad inspect` in a repo, and your key isn't in ssh-agent, it will return an error, or ask you to enter your passphrase.

But these commands don't sign anything and therefore shouldn't require that. Other read-only commands like `rad-self` and `rad ls` should be checked for the same behavior.

### LINKED ISSUE #315: `rad-auth` doesn't add your key to ssh-agent the first time
When running `rad auth` for the first time to initialize a new key, it should be added to ssh-agent as well, if ssh-agent is running. Otherwise it shouldn't say anything.

### LINKED ISSUE #319: Git is exposing itself
Heartwood exposes git through its own port, and should minimize exposure to the git daemon supporting it.  It should be started listen on a random high port (on localhost), however it is being exposed on git's official port.

cloudhead:
Fixes #319 
Fixes #315 
Fixes #314

### DIFF
--- a/radicle-cli/src/commands/auth.rs
+++ b/radicle-cli/src/commands/auth.rs
@@ -4,6 +4,7 @@ use std::ffi::OsString;
 use anyhow::anyhow;
 
 use radicle::crypto::ssh;
+use radicle::crypto::ssh::Passphrase;
 use radicle::{profile, Profile};
 
 use crate::terminal as term;
@@ -86,8 +87,15 @@ pub fn init(options: Options) -> anyhow::Result<()> {
         term::passphrase_confirm()
     }?;
     let spinner = term::spinner("Creating your 🌱 Ed25519 keypair...");
-    let profile = Profile::init(home, passphrase)?;
+    let profile = Profile::init(home, passphrase.clone())?;
     spinner.finish();
+
+    let spinner = term::spinner("Adding your radicle key to ssh-agent...");
+    if register(&profile, passphrase).is_ok() {
+        spinner.finish();
+    } else {
+        spinner.warn();
+    }
 
     term::success!(
         "Your Radicle ID is {}. This identifies your device.",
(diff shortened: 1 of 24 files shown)


radicle-cli/src/commands/auth.rs
```diff
@@ -65,7 +66,10 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
 }
 
 pub fn init(options: Options) -> anyhow::Result<()> {
-    term::headline("Initializing your 🌱 profile and identity");
+    term::headline(format!(
+        "Initializing your {} 🌱 identity",
+        term::format::highlight("radicle")
```
FintanH, line 71: Style question: how come we don't capitalise "radicle"?

I think it's actually a mixed bag for if we do or don't. Just noticed a "Radicle" below. I've been thinking it would be a good idea to have a sort of style guide for the CLI output to be more consistent when editing/adding commands.

radicle-cli/src/commands/inspect.rs
```diff
@@ -122,82 +112,94 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+                .spawn()?
+                .wait()?;
+        }
+        Target::Payload => {
             println!(
-                "{} {}",
-                term::format::yellow("commit"),
-                term::format::yellow(oid),
+                "{}",
+                colorizer().colorize_json_str(&serde_json::to_string_pretty(&project.payload)?)?
```
FintanH, line 142: Should we add `term::json`?

FintanH (approved):
Looks good :+1:

radicle-cli/src/commands/auth.rs
cloudhead: Yeah, good question. I've been waffling about this one. I think it looks better lowercase usually, in the terminal context. A style guide is a good idea, though we may be able to encode all of it in functions/consts, eg. for this we could have `terminal::format::RADICLE` or something, and it would be the correct color/capitalization. Same for RIDs, NIDs etc.

radicle-cli/src/commands/inspect.rs
cloudhead: Yeah probably :), `term::format::json` I guess.

radicle-cli/src/commands/auth.rs
FintanH: Ya, that'd be a great idea!

PR #366, opened 2023-02-28
radicle-httpd: sort cobs listing endpoints

sebastinez:
As talked on zulip https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/cobs.20pagination

We need the issues and patches to be returned from `GET /projects/:project/issues` and `GET /projects/:project/patches` to be sorted before doing the pagination.
This will allow us to receive the same output every time we call the endpoint with the same params.

### DIFF
--- a/radicle-httpd/src/api/v1/projects.rs
+++ b/radicle-httpd/src/api/v1/projects.rs
@@ -394,10 +394,10 @@ async fn issues_handler(
     let storage = &ctx.profile.storage;
     let repo = storage.repository(project)?;
     let issues = Issues::open(ctx.profile.public_key, &repo)?;
+    let mut issues: Vec<_> = issues.all()?.filter_map(|r| r.ok()).collect::<Vec<_>>();
+    issues.sort_by(|(_, a, _), (_, b, _)| b.timestamp().cmp(&a.timestamp()));
     let issues = issues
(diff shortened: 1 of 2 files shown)


radicle-httpd/src/api/v1/projects.rs
```diff
@@ -393,10 +393,10 @@ async fn issues_handler(
     let storage = &ctx.profile.storage;
     let repo = storage.repository(project)?;
     let issues = Issues::open(ctx.profile.public_key, &repo)?;
+    let mut issues: Vec<_> = issues.all()?.filter_map(|r| r.ok()).collect::<Vec<_>>();
```
FintanH, line 396: You should probably do the `skip` and `take` before sorting, so you minimise the elements to sort.

sebastinez: That would be nice, but the thing is that if we skip and take before sorting, we'll have elements that will be excluded from the final output.
I'm aware of the performance hit, but else we won't be able to present the issues or patches in a sorted way to the user.

FintanH: Ah yes, you're dead right! :)

cloudhead (approved):

PR #368, opened 2023-02-28
`rad fetch`

cloudhead:
`rad fetch` fetches refs from seeds, for the given project.

### DIFF
--- a/radicle-cli/examples/rad-inspect.md
+++ b/radicle-cli/examples/rad-inspect.md
@@ -68,3 +68,6 @@ date   Thu, 15 Dec 2022 17:28:04 +0000
  }
 
 ```
+
+The identity document is the metadata associated with a repository, that is
+only changeable by delegates.
--- a/radicle-cli/src/commands/clone.rs
+++ b/radicle-cli/src/commands/clone.rs
@@ -10,13 +10,14 @@ use radicle::git::raw;
 use radicle::identity::doc::{DocError, Id};
 use radicle::identity::{doc, IdentityError};
 use radicle::node;
-use radicle::node::{FetchResult, Handle as _, Node};
+use radicle::node::{Handle as _, Node};
 use radicle::prelude::*;
 use radicle::rad;
 use radicle::storage;
 use radicle::storage::git::Storage;
 
 use crate::commands::rad_checkout as checkout;
+use crate::commands::rad_fetch as fetch;
 use crate::project;
 use crate::terminal as term;
 use crate::terminal::args::{Args, Error, Help};
@@ -29,7 +30,7 @@ pub const HELP: Help = Help {
     usage: r#"
 Usage
 
-    rad clone <id> [<option>...]
+    rad clone <rid> [<option>...]
 
 Options
 
@@ -152,8 +153,10 @@ pub enum CloneError {
     Payload(#[from] doc::PayloadError),
     #[error("project error: {0}")]
     Identity(#[from] IdentityError),
-    #[error("no seeds found for {0}")]
+    #[error("repository {0} not found")]
     NotFound(Id),
+    #[error("no seeds found for {0}")]
+    NoSeeds(Id),
 }
 
 pub fn clone<G: Signer>(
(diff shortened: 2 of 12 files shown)


radicle-cli/src/commands/fetch.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    if success == 0 {
+        term::error(format!("Failed to fetch repository from {failed} seed(s)"));
+    } else {
+        term::success!("Fetched repository from {success} seed(s)");
+    }
+    Ok(())
+}
+
+pub fn fetch(rid: Id, node: &mut Node) -> Result<(usize, usize), node::Error> {
```
FintanH, line 111: I'd suggest having a type that says which seeds succeeded or failed. That way if we want to report which seeds succeeded/failed we can. Then `usize` can be calculated using `len`.

radicle-cli/src/commands/fetch.rs
```diff
@@ -0,0 +1,149 @@
...
+    --force, -f     Fetch even if the repository isn't tracked
+    --help          Print help
+
+"#,
+};
+
+#[derive(Debug)]
+pub struct Options {
+    rid: Option<Id>,
+    seed: Option<NodeId>,
```
FintanH, line 39: How about `seeds: Vec<NodeId>` instead to specify multiple?

cloudhead: Don't think it's super useful yet, was even debating having this option at all. Thing is right now you have to be connected to the seed to fetch from it.

radicle-cli/src/terminal/io.rs
```diff
@@ -19,6 +19,7 @@ use super::Error;
 use super::{style, Paint};
 
 pub const ERROR_PREFIX: Paint<&str> = Paint::red("✗");
+pub const WARNING_PREFIX: Paint<&str> = Paint::yellow("!");
```
slack-coder, line 23: :+1:

radicle-cli/examples/rad-inspect-noauth.md
```diff
@@ -0,0 +1,13 @@
...
+$ rad self
+✗ Self failed: Could not load radicle profile
+✗ To setup your radicle profile, run `rad auth`.
+
+```
+
+```
+$ rad inspect
+rad:z42hL2jL4XNk6K8oHQaSWfMgCL7ji
+```
```
slack-coder, line 13: Would this be better as a regular test?  Personally speaking, it may not even be necessary to test this at all.  Having `rad inspect` work without authentication is nice and expected behavior.

Keeping documentation tests as documentation first will help end users (and myself) when learning how to use radicle.

radicle-cli/examples/rad-inspect.md
```diff
@@ -0,0 +1,70 @@
...
+  "xyz.radicle.project": {
+    "defaultBranch": "master",
+    "description": "Radicle Heartwood Protocol & Stack",
+    "name": "heartwood"
+  }
+}
+```
+
+Finally, the `--history` flag allows you to examine the identity document's
+history:
```
slack-coder, line 44: Maybe very briefly describe what the identity document is for?

radicle-cli/src/commands/inspect.rs
```diff
@@ -122,82 +112,94 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+
+                if let Some(msg) = tip.message() {
+                    for line in msg.lines() {
+                        if line.is_empty() {
+                            println!();
+                        } else {
+                            term::indented(term::format::dim(line));
+                        }
+                    }
+                    term::blank();
```
slack-coder, line 189: This is a lot.  Would moving it into its own function help readers?

radicle-cli/src/commands/clone.rs
```diff
@@ -152,8 +153,10 @@ pub enum CloneError {
     Payload(#[from] doc::PayloadError),
     #[error("project error: {0}")]
     Project(#[from] ProjectError),
-    #[error("no seeds found for {0}")]
+    #[error("repository {0} not found")]
```
slack-coder, line 156: :+1:

radicle-cli/examples/rad-inspect-noauth.md
cloudhead: Could do yeah.

radicle-cli/src/commands/inspect.rs
cloudhead: I didn't want to do any unnecessary refactoring here, just left it as-is.

radicle-cli/examples/rad-inspect-noauth.md
cloudhead: Hmm this is more complicated actually, because I don't currently have a way of running a CLI command without a home directory, so will leave it for later.

Issue #316, opened 2023-02-15
Only fetch if we're tracking the remote


cloudhead:
Currently, when a ref announcement comes in, we fetch regardless of which peer originally announced the refs. What we should do is check our node tracking policy to see if any of the announced refs are for a peer we're tracking, and only then initiate the fetch.

This code is in the node's `service` module under `handle_announcement`:
```rust
                                if is_fresh {
                                    // TODO: Only fetch if the refs announced are for peers we're tracking.
                                    self.fetch(message.rid, announcer);
                                }
```

keepsimple1:
I'd like to take this one if no one is working on it. In the testing code, how do we insert new refs into a node (i.e. create refs announcements)?

cloudhead:
@keepsimple1 you could test this entirely at the service layer, by eg. crafting a RefAnnouncement manually, sending it to the node, and checking whether the service outputs a Fetch action. Here's an example test for announcement relays: https://github.com/radicle-dev/heartwood/blob/master/radicle-node/src/tests.rs#L513

You'd write a similar test, but assert that an `Io::Fetch` message is sent or not sent, depending on the tracking status of the peer.

Note that the relayer of the message is irrelevant wrt. tracking, what matters is the origin/author of the message, also called "announcer" in the code.

keepsimple1:
> assert that an `Io::Fetch` message is sent or not sent, depending on the tracking status of the peer.

There are two actions in `handle_announcement`:  1) do fetch  2) relay the announcement. 
If the peer is not tracked,  it is clear that we will not do fetch (action 1) .  Do we also not to relay the announcement? Reading the code, it seems to me that we can still relay the announcement because other nodes might be tracking announcer.  No?

keepsimple1:
When does a node track another node? More specifically: 

- Does the end user decide the tracking? If yes, how?
- Should the node software track other nodes automatically? If yes, in what scenarios?

cloudhead:
* The end users decides, always. But there is a "default" policy as well which is configured when you start the node.
* No, but when the tracking policy is consulted, it might decide to fetch node refs that aren't explicitly tracked, due to the default policy of the node

FintanH:
Closing via #370

PR #370, opened 2023-02-28
radicle-node: fetch refs only if tracking the node

keepsimple1:
This is to solve issue #316 

For the design considerations history, please see https://github.com/radicle-dev/heartwood/pull/370#discussion_r1133476801

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -50,6 +50,7 @@ pub use crate::service::session::Session;
 use self::gossip::Gossip;
 use self::message::InventoryAnnouncement;
 use self::reactor::Reactor;
+use self::tracking::NamespacesError;
 
 /// Target number of peers to maintain connections to.
 pub const TARGET_OUTBOUND_PEERS: usize = 8;
@@ -102,6 +103,8 @@ pub enum Error {
     Routing(#[from] routing::Error),
     #[error(transparent)]
     Tracking(#[from] tracking::Error),
+    #[error("namespaces error: {0}")]
+    Namespaces(#[from] NamespacesError),
 }
 
 /// Function used to query internal service state.
@@ -871,27 +874,29 @@ where
                 }
 
                 // TODO: Buffer/throttle fetches.
-                if self
-                    .tracking
-                    .is_repo_tracked(&message.rid)
-                    .expect("Service::handle_announcement: error accessing tracking configuration")
-                {
+                let repo_entry = self.tracking.repo_policy(&message.rid).expect(
+                    "Service::handle_announcement: error accessing repo tracking configuration",
+                );
+
+                if repo_entry.policy == tracking::Policy::Track {
                     // Refs can be relayed by peers who don't have the data in storage,
                     // therefore we only check whether we are connected to the *announcer*,
                     // which is required by the protocol to only announce refs it has.
                     if self.sessions.is_connected(announcer) {
-                        match message.is_fresh(&self.storage) {
-                            Ok(is_fresh) => {
-                                if is_fresh {
-                                    // TODO: Only fetch if the refs announced are for peers we're tracking.
-                                    self.fetch(message.rid, announcer);
-                                }
+                        match self.should_fetch_refs_announcement(message, &repo_entry.scope) {
+                            Ok(true) => self.fetch(message.rid, announcer),
+                            Ok(false) => {
+                                debug!(target: "service", "Skip fetch the refs from {announcer}")
                             }
                             Err(e) => {
-                                error!(target: "service", "Failed to check ref announcement freshness: {e}");
+                                error!(target: "service", "Failed to check refs announcement: {e}");
+                                return Err(session::Error::Misbehavior);
                             }
                         }
+                    } else {
+                        debug!(target: "service", "No sessions connected to {announcer}");
                     }
+
                     return Ok(relay);
                 } else {
                     debug!(
@@ -961,6 +966,53 @@ where
             }
         }
         Ok(false)
+    }
+
+    /// A convenient method to check if we should fetch from a `RefsAnnouncement`
+    /// with `scope`.
+    fn should_fetch_refs_announcement(
+        &self,
+        message: &RefsAnnouncement,
+        scope: &tracking::Scope,
+    ) -> Result<bool, Error> {
+        // First, check the freshness.
+        if !message.is_fresh(&self.storage)? {
+            debug!(target: "service", "All refs of {} are already in the local node", &message.rid);
+            return Ok(false);
+        }
+
+        // Second, check the scope.
+        match scope {
+            tracking::Scope::All => Ok(true),
+            tracking::Scope::Trusted => {
+                match self.tracking.namespaces_for(&self.storage, &message.rid) {
+                    Ok(Namespaces::All) => Ok(true),
+                    Ok(Namespaces::Many(nodes)) => {
+                        // Get the set of trusted nodes except self.
+                        let my_id = self.node_id();
+                        let node_set: HashSet<_> =
+                            nodes.iter().filter(|key| *key != &my_id).collect();
+
+                        // Check if there is at least one trusted ref.
+                        Ok(message
+                            .refs
+                            .iter()
+                            .any(|(pub_key, _refs)| node_set.contains(pub_key)))
+                    }
+                    Ok(Namespaces::One(key)) => {
+                        Ok(message.refs.iter().any(|(pub_key, _refs)| pub_key == &key))
+                    }
+                    Err(NamespacesError::NoTrusted { rid }) => {
+                        debug!(target: "service", "No trusted nodes to fetch {}", &rid);
+                        Ok(false)
+                    }
+                    Err(e) => {
+                        error!(target: "service", "Failed to obtain namespaces: {e}");
+                        Err(e.into())
+                    }
+                }
+            }
+        }
     }
 
     pub fn handle_message(
--- a/radicle-node/src/test/peer.rs
+++ b/radicle-node/src/test/peer.rs
@@ -5,6 +5,7 @@ use std::ops::{Deref, DerefMut};
 
 use crossbeam_channel as chan;
 use log::*;
+use radicle::storage::ReadRepository;
 
 use crate::address;
 use crate::address::Store;
@@ -250,7 +251,20 @@ where
     }
 
     pub fn refs_announcement(&self, rid: Id) -> Message {
-        let refs = BoundedVec::new();
+        let mut refs = BoundedVec::new();
+        if let Ok(repo) = self.storage().repository(rid) {
+            if let Ok(false) = repo.is_empty() {
+                if let Ok(remotes) = repo.remotes() {
+                    for (remote_id, remote) in remotes.into_iter() {
+                        if let Err(e) = refs.push((remote_id, remote.refs.unverified())) {
+                            debug!(target: "test", "Failed to push {remote_id} to refs: {e}");
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
         let ann = AnnouncementMessage::from(RefsAnnouncement {
             rid,
             refs,
--- a/radicle-node/src/tests.rs
+++ b/radicle-node/src/tests.rs
@@ -669,6 +669,98 @@ fn test_refs_announcement_relay() {
     );
 }
 
+/// Even if Alice is not tracking Bob, Alice will fetch Bob's refs for a repo she doesn't have.
+#[test]
+fn test_refs_announcement_fetch_trusted_no_inventory() {
+    logger::init(log::Level::Debug);
+
+    let tmp = tempfile::tempdir().unwrap();
+    let mut alice = Peer::config(
+        "alice",
+        [7, 7, 7, 7],
+        Storage::open(tmp.path().join("alice")).unwrap(),
+        peer::Config::default(),
+    );
+    let bob = {
+        let mut rng = fastrand::Rng::new();
+        let signer = MockSigner::new(&mut rng);
+        let storage = fixtures::storage(tmp.path().join("bob"), &signer).unwrap();
+
+        Peer::config(
+            "bob",
+            [9, 9, 9, 9],
+            storage,
+            peer::Config {
+                signer,
+                rng,
+                ..peer::Config::default()
+            },
+        )
+    };
+    let bob_inv = bob.storage().inventory().unwrap();
+    let rid = bob_inv[0];
+
+    alice.track_repo(&rid, tracking::Scope::Trusted).unwrap();
+    alice.connect_to(&bob);
+
+    // Alice receives Bob's refs.
+    alice.receive(bob.id(), bob.refs_announcement(rid));
+
+    // Alice fetches Bob's refs as this is a new repo.
+    assert_eq!(
+        alice.messages(bob.id()).next(),
+        Some(Message::Fetch { rid })
+    );
+}
+
+/// Alice and Bob both have the same repo.
+///
+/// First, Alice will not fetch from Bob's `RefsAnnouncement` as Alice does not
+/// track Bob as `Trusted`.
+///
+/// Later Alice tracks Bob, and will be able to fetch Bob's refs.
+#[test]
+fn test_refs_announcement_trusted() {
+    logger::init(log::Level::Debug);
+
+    // Create MockStorage for Alice and Bob. Both will have repo with `rid`.
+    let storage_alice = arbitrary::nonempty_storage(1);
+    let rid = *storage_alice.inventory.keys().next().unwrap();
+    let storage_bob = storage_alice.clone();
+    let mut alice = Peer::with_storage("alice", [7, 7, 7, 7], storage_alice);
+    let mut bob = Peer::with_storage("bob", [8, 8, 8, 8], storage_bob);
+
+    // Generate some refs for Bob under their own node_id.
+    let refs = arbitrary::gen::<Refs>(8);
+    let signed_refs = refs.signed(bob.signer()).unwrap();
+    let node_id = bob.id;
+    bob.storage_mut().insert_remote(rid, node_id, signed_refs);
+
+    // Alice uses Scope::Trusted, and did not track Bob yet.
+    alice.connect_to(&bob);
+    alice.track_repo(&rid, tracking::Scope::Trusted).unwrap();
+
+    // Alice receives Bob's refs
+    alice.receive(bob.id(), bob.refs_announcement(rid));
+
+    // Alice does not fetch as Alice is not tracking Bob.
+    assert!(
+        alice.messages(bob.id()).next().is_none(),
+        "Alice is not tracking bob yet."
+    );
+
+    // Alice starts to track Bob.
+    let (sender, receiver) = chan::bounded(1);
+    alice.command(Command::TrackNode(bob.id, Some("bob".to_string()), sender));
+    let policy_change = receiver.recv().map_err(runtime::HandleError::from).unwrap();
+    assert!(policy_change);
+
+    // Bob announces refs again.
+    bob.elapse(LocalDuration::from_mins(1)); // Make sure our announcement is fresh.
+    alice.receive(bob.id(), bob.refs_announcement(rid));
+    assert_matches!(alice.messages(bob.id()).next(), Some(Message::Fetch { .. }));
+}
+
 #[test]
 fn test_refs_announcement_no_subscribe() {
     let storage = arbitrary::nonempty_storage(1);
--- a/radicle/src/test/storage.rs
+++ b/radicle/src/test/storage.rs
@@ -8,6 +8,7 @@ use radicle_git_ext as git_ext;
 use crate::crypto::{Signer, Verified};
 use crate::identity::doc::{Doc, Id};
 use crate::identity::IdentityError;
+use crate::node::NodeId;
 
 pub use crate::storage::*;
 
@@ -15,6 +16,10 @@ pub use crate::storage::*;
 pub struct MockStorage {
     pub path: PathBuf,
     pub inventory: HashMap<Id, Doc<Verified>>,
+
+    /// All refs keyed by RID.
+    /// Each value is a map of refs keyed by node Id (public key).
+    pub remotes: HashMap<Id, HashMap<NodeId, refs::SignedRefs<Verified>>>,
 }
 
 impl MockStorage {
@@ -22,6 +27,7 @@ impl MockStorage {
         Self {
             path: PathBuf::default(),
             inventory: inventory.into_iter().collect(),
+            remotes: HashMap::new(),
         }
     }
 
@@ -29,7 +35,21 @@ impl MockStorage {
         Self {
             path: PathBuf::default(),
             inventory: HashMap::new(),
+            remotes: HashMap::new(),
         }
+    }
+
+    /// Add a remote `node` with `signed_refs` for the repo `rid`.
+    pub fn insert_remote(
+        &mut self,
+        rid: Id,
+        node: NodeId,
+        signed_refs: refs::SignedRefs<Verified>,
+    ) {
+        self.remotes
+            .entry(rid)
+            .or_insert(HashMap::new())
+            .insert(node, signed_refs);
     }
 }
 
@@ -64,6 +84,7 @@ impl ReadStorage for MockStorage {
         Ok(MockRepository {
             id: rid,
             doc: doc.clone(),
+            remotes: self.remotes.get(&rid).cloned().unwrap_or_default(),
         })
     }
 }
@@ -76,6 +97,7 @@ impl WriteStorage for MockStorage {
         Ok(MockRepository {
             id: rid,
             doc: doc.clone(),
+            remotes: self.remotes.get(&rid).cloned().unwrap_or_default(),
         })
     }
 
@@ -84,9 +106,11 @@ impl WriteStorage for MockStorage {
     }
 }
 
+#[derive(Clone, Debug)]
 pub struct MockRepository {
     id: Id,
     doc: Doc<Verified>,
+    remotes: HashMap<NodeId, refs::SignedRefs<Verified>>,
 }
 
 impl ReadRepository for MockRepository {
@@ -95,7 +119,7 @@ impl ReadRepository for MockRepository {
     }
 
     fn is_empty(&self) -> Result<bool, git2::Error> {
-        Ok(true)
+        Ok(self.remotes.is_empty())
     }
 
     fn head(&self) -> Result<(fmt::Qualified, Oid), IdentityError> {
@@ -114,12 +138,19 @@ impl ReadRepository for MockRepository {
         todo!()
     }
 
-    fn remote(&self, _remote: &RemoteId) -> Result<Remote<Verified>, refs::Error> {
-        todo!()
+    fn remote(&self, remote: &RemoteId) -> Result<Remote<Verified>, refs::Error> {
+        self.remotes
+            .get(remote)
+            .map(|refs| Remote::new(*remote, refs.clone()))
+            .ok_or(refs::Error::InvalidRef)
     }
 
     fn remotes(&self) -> Result<Remotes<Verified>, refs::Error> {
-        todo!()
+        Ok(self
+            .remotes
+            .iter()
+            .map(|(id, refs)| (*id, Remote::new(*id, refs.clone())))
+            .collect())
     }
 
     fn commit(&self, _oid: Oid) -> Result<git2::Commit, git_ext::Error> {


radicle-cli/tests/commands.rs
```diff
@@ -341,6 +341,12 @@ fn test_replication_via_seed() {
     alice.connect(&seed);
     bob.connect(&seed);
 
+    // Track node to enable fetch.
+    alice
+        .handle
+        .track_node(seed.id, Some("seed".to_string()))
+        .unwrap();
+
```
keepsimple1, line 349: Although this test is passing, I'm not sure why "seed" seems to be the announcer for bob's refs. Is that expected?

FintanH: It's because of the topology of the network:
```
alice -- seed -- bob
```

So when bob announces changes the seed hears them and relays them to alice. Although, I don't think having to track the seed makes much sense. It should be that alice tracks bob and hears about bob's refs so she'll replicate 🤔

cloudhead:
This is looking good.

There's one thing I forgot to mention that is important and makes the general UX a lot better: repo tracking policies include a policy `Scope`. This scope further describes the tracking policy for that repo:

```rust
/// Tracking scope of a repository tracking policy.
pub enum Scope {
    /// Track remotes of nodes that are already tracked.
    Trusted,
    /// Track remotes of repository delegates.
    DelegatesOnly,
    /// Track all remotes.
    All,
}
```

So actually what we need to do is that if the scope is `Trusted`, then we check that we're tracking the peer. If it's `DelegatesOnly`, we check that the refs come from a project delegate, and if it's `All`, we don't check at all and just fetch.

radicle-cli/tests/commands.rs
cloudhead: Yeah this is a good point. I'm guessing what's happening is that the seed is fetching from Alice, then the seed is making its own announcement of refs after it fetched, and Bob is fetching from the seed because he tracks it.

Probably we should be relaying announcements even if we don't follow the peer :thinking:.

radicle-node/src/service.rs
```diff
@@ -859,8 +869,14 @@ where
                         match message.is_fresh(&self.storage) {
                             Ok(is_fresh) => {
                                 if is_fresh {
-                                    // TODO: Only fetch if the refs announced are for peers we're tracking.
-                                    self.fetch(message.rid, announcer);
+                                    if self
+                                        .tracking
+                                        .is_node_tracked(announcer)
+                                        .expect("Service:handle_announcement: error accessing node tracking configuration") {
```
cloudhead, line 875:
```suggestion
                                        .expect("Service::handle_announcement: error accessing node tracking configuration") {
```

radicle-cli/tests/commands.rs
FintanH: Ya, I would think that gossip should definitely be relayed. Otherwise we might end up in a scenario where a limited set of resources are tracked and announced :)

keepsimple1: My question is about the last leg of the replication, where Bob has created (cloned) its own refs under their namespace. Then Alice would replicate Bob's refs via the seed.  In this operation, Alice gets the refs announcement from the seed, which is expected. However, the announcement's "announcer" is set to the seed, not Bob, which I thought was the origin. 

Here is test log from Alice's point of view, where Alice's public key is:  `z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi `,   the seed is `z6Mkux1aUQD2voWWukVb5nNUR7thrHveQG4pDQua8nVhib7Z`,  and Bob is `z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk`: 

```
z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi     service: Received message Announcement(z6Mkux1aUQD2voWWukVb5nNUR7thrHveQG4pDQua8nVhib7Z, Refs(rad:z42hL2jL4XNk6K8oHQaSWfMgCL7ji, [(PublicKey(z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk), SignedRefs { refs: Refs({RefString("refs/heads/master"): Oid(f2de534b5e81d7c6e2dcaf58c3dd91573c0a0354), RefString("refs/rad/id"): Oid(3f3596fd2ce9a1e5c32b29840c1c2d38d85cd134)}), signature: Signature(z9So7T4ji6kibR1XyPvX97k86EFGYMmcZ3HEhe3RPSx2NpdrpMsrjiCc1vg3oLEswqqDrSMi6dRWQRQKfhBhaeDx), _verified: PhantomData<radicle_crypto::Unverified> }), (PublicKey(z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi), SignedRefs { refs: Refs({RefString("refs/heads/master"): Oid(f2de534b5e81d7c6e2dcaf58c3dd91573c0a0354), RefString("refs/rad/id"): Oid(3f3596fd2ce9a1e5c32b29840c1c2d38d85cd134)}), signature: Signature(z2HaXR5GjG2UBXNjLBeCNSb6aicbGdiRaASm6ocLziDb7WfpRgQTgCruFyGSeRLoFiRc6H6VT1szb7M3b17WdRbez), _verified: PhantomData<radicle_crypto::Unverified> })])) from z6Mkux1aUQD2voWWukVb5nNUR7thrHveQG4pDQua8nVhib7Z          
``` 

later,  we can see Alice stored the refs from this announcement:

```
z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi      worker: Git:  * [new ref]         refs/namespaces/z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk/refs/heads/master -> refs/namespaces/z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk/refs/heads/master
z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi      worker: Git:  * [new ref]         refs/namespaces/z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk/refs/rad/id -> refs/namespaces/z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk/refs/rad/id
z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi      worker: Git:  * [new ref]         refs/namespaces/z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk/refs/rad/sigrefs -> refs/namespaces/z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk/refs/rad/sigrefs
z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi      worker: Fetch for rad:z42hL2jL4XNk6K8oHQaSWfMgCL7ji exited successfully
```

And, if we removed the line of `track_node(seed.id, Some("seed"))`,  the fetch will be skipped and test fails.

keepsimple1:
> So actually what we need to do is that if the scope is `Trusted`, then we check that we're tracking the peer. If it's `DelegatesOnly`, we check that the refs come from a project delegate, and if it's `All`, we don't check at all and just fetch.

Looks like we can either add this logic in this PR, or in a separate PR. I think it's easier for me to add it in a separate PR, but I'm okay either way. If there is some additional reason I'm not aware of for adding this in this PR, let me know and I will work on that.

radicle-cli/tests/commands.rs
keepsimple1: Btw, I don't think this question is blocking this PR itself.

cloudhead: Yeah, as I said, the seed is making its *own* announcement of local refs, to Bob, after it fetches from Alice. This is not Alice's announcement being relayed by the seed. That's why Bob is fetching from the seed, since he is tracking it.

But this isn't what we want actually, because we don't want Bob to have to track the seed. We want the seed to relay Alice's announcement to Bob, and Bob fetch from the seed because he tracks *Alice*.

The only way to achieve that, I think, is to actually not look at the announcer of the message, but rather at the refs themselves.

If the refs contain a remote that we are tracking, then we fetch from the announcer (the seed). Otherwise, we don't.

This should preserve most of the existing behavior, and allow us to revert some of the changes to the tests.

cloudhead:
> > So actually what we need to do is that if the scope is `Trusted`, then we check that we're tracking the peer. If it's `DelegatesOnly`, we check that the refs come from a project delegate, and if it's `All`, we don't check at all and just fetch.
> 
> Looks like we can either add this logic in this PR, or in a separate PR. I think it's easier for me to add it in a separate PR, but I'm okay either way. If there is some additional reason I'm not aware of for adding this in this PR, let me know and I will work on that.

I'm fine with adding it in a separate PR, as long as we aren't unexpectedly changing the behavior of existing tests. I *think* that if you fix the other issue mentioned above, we can get there.

radicle-cli/tests/commands.rs
keepsimple1: A `RefsAnnouncement` contains multiple refs which are not necessarily from the same remote. If we check the tracking info of the remote instead of the announcer, we are changing the granularity of the check.

Check tracking of the announcer vs. check tracking of each ref is surely different behavior. One note:

Currently `RefsAnnouncement` implementation is different from [RIP 1](https://github.com/radicle-dev/rips/blob/master/0001-heartwood.md#reference-announcements): 

in code: 

```
pub struct RefsAnnouncement {
    /// Repository identifier.
    pub rid: Id,
    /// Updated refs.
    pub refs: BoundedVec<(NodeId, SignedRefs<Unverified>), REF_REMOTE_LIMIT>,
    /// Time of announcement.
    pub timestamp: Timestamp,
}
```

in RIP 1: 

```
RefsAnnouncement = (
    NodeId,
    RepositoryId,
    Map<RefName, CommitId>,
    Signature,
)
```

i.e. in RIP 1, there is an extra field `NodeId`.  It seems that it was extracted out, and becomes `announcer` in the code. Doesn't that mean the announcer should be the original node of the refs, not just a relayer or sender?

cloudhead: Yes, we recently changed `RefsAnnouncement` (and I should update the RIP) to add more flexibility. Originally, you were only allowed to send announcements that you yourself signed, hence there was only one NodeId, since all refs were signed by that node.

Now, ref announcement messages can contain multiple remotes, each signed by the respective node. Hence the vector of (NodeId, SignedRefs) tuples.

cloudhead: The difference is that this allows us to distinguish between the announcer of the refs, and the signer of the refs.

cloudhead: > If we check the tracking info of the remote instead of the announcer, we are changing the granularity of the check.

Correct.

In the ideal case, we also only fetch the remotes that we're tracking. Eg. Seed announces refs from 5 peers, an I am only tracking 2 of those peers; I fetch only the 2 tracked peers from the announcer (if I'm connected to it).

keepsimple1: I have a related question:  in the following comments in the current code: 

https://github.com/radicle-dev/heartwood/blob/9d244b221a167c1c2bacbd00ecefc1a44ecab552/radicle-node/src/service.rs#L856-L858

What does `refs it has` mean?  Does it mean only the refs under the namespace of the announcer's public key, or all refs in the announcer's local storage (for that project)?

keepsimple1: > In the ideal case, we also only fetch the remotes that we're tracking. Eg. Seed announces refs from 5 peers, an I am only tracking 2 of those peers; I fetch only the 2 tracked peers from the announcer (if I'm connected to it).

Currently we either do fetch `All` or only one particular namespace. 
```
pub enum Namespaces {
    /// All namespaces.
    #[default]
    All,
    /// A single namespace, by public key.
    One(PublicKey),
}
```

To pick and choose the exact set of refs in an announcement, we will need to fetch refs of one peer at a time.

keepsimple1: I've made changes to the `Fetch` message. Now we will track the peer owning the refs, instead of the announcer. 

Hence this line in this test case is changed to track `bob` instead of `seed`.

FintanH: > To pick and choose the exact set of refs in an announcement, we will need to fetch refs of one peer at a time. And looks like we might need to change the Fetch message format on the wire as well to include namespaces.

That's actually what I'm working on here:
https://github.com/radicle-dev/heartwood/pull/354

cloudhead: > What does refs it has mean? Does it mean only the refs under the namespace of the announcer's public key, or all refs in the announcer's local storage (for that project)?

All refs. Put differently, `/<alice>/ref/heads/master` and `/<bob>/refs/heads/master` are distinct refs. So in that comment an exact string match of the full ref string is what is meant.

keepsimple1: I've added the logic of checking the repo tracking scope and this test no longer needs code changes.

keepsimple1:
> I'm fine with adding it in a separate PR, as long as we aren't unexpectedly changing the behavior of existing tests. I _think_ that if you fix the other issue mentioned above, we can get there.

I've updated the patch to check for the repo tracking `Scope`. For now, only when the scope is `Scope::Trusted` we will check for each node's tracking record. Hence the existing tests now pass without changes.

I have not added any special handling for `Scope::DelegatesOnly` as I'm not clear what is the proper way of verifying the delegates. I can add that in a follow-up PR if that's okay.

radicle-cli/tests/commands.rs
keepsimple1: > That's actually what I'm working on here:

Oops. I didn't know that. I have no problem if you want to land first.

radicle-node/src/service.rs
```diff
@@ -739,7 +750,7 @@ where
 
     /// Handle an announcement message.
     ///
-    /// Returns `true` if this announcement should be stored and relayed to connected peers,
+    /// Returns `true` if this announcement should be relayed to connected peers,
```
cloudhead, line 753: Announcements are stored and redelivered to connecting nodes to sync them up if they've been offline :)

radicle-node/src/service.rs
```diff
@@ -1018,14 +1046,14 @@ where
                     }
                 }
             }
-            (session::State::Connected { .. }, Message::Fetch { rid }) => {
+            (session::State::Connected { .. }, Message::Fetch { rid, namespaces }) => {
```
cloudhead, line 1049: The namespace isn't needed in the `Fetch` request. There's https://github.com/radicle-dev/heartwood/issues/379 to make that clearer.

radicle-node/src/service.rs
```diff
@@ -860,14 +871,31 @@ where
                         match message.is_fresh(&self.storage) {
                             Ok(is_fresh) => {
                                 if is_fresh {
-                                    // TODO: Only fetch if the refs announced are for peers we're tracking.
-                                    self.fetch(message.rid, announcer);
+                                    let repo_tracking = self
+                                        .tracking
+                                        .repo_entry(&message.rid)
+                                        .expect("handle_announcement: error accessing repo tracking configuration");
```
cloudhead, line 877: We're already accessing the tracking policy in the parent `if` statement, so maybe bind this result outside and it can be used here.

radicle-node/src/service.rs
```diff
@@ -860,14 +871,31 @@ where
...
+
+                                    if let Some((Scope::Trusted, _policy)) = repo_tracking {
+                                        for (pub_key, _refs) in message.refs.iter() {
+                                            if self.tracking.is_node_tracked(pub_key)
+                                            .expect("handle_announcement: error accessing node tracking configuration") {
+                                                self.fetch(message.rid, Namespaces::One(*pub_key), announcer);
```
cloudhead, line 883: Maybe we should rebase this on @FintanH's patch, which allows for setting multiple pks in a Namespaces object.

radicle-node/src/service.rs
```diff
@@ -860,14 +871,31 @@ where
...
                             Ok(is_fresh) => {
                                 if is_fresh {
-                                    // TODO: Only fetch if the refs announced are for peers we're tracking.
-                                    self.fetch(message.rid, announcer);
+                                    let repo_tracking = self
+                                        .tracking
+                                        .repo_entry(&message.rid)
+                                        .expect("handle_announcement: error accessing repo tracking configuration");
+
+                                    if let Some((Scope::Trusted, _policy)) = repo_tracking {
```
cloudhead, line 879: `policy` should be `Track`

radicle-node/src/service.rs
```diff
@@ -860,14 +871,31 @@ where
...
+                                        for (pub_key, _refs) in message.refs.iter() {
+                                            if self.tracking.is_node_tracked(pub_key)
+                                            .expect("handle_announcement: error accessing node tracking configuration") {
+                                                self.fetch(message.rid, Namespaces::One(*pub_key), announcer);
+                                            } else {
+                                                debug!("handle_announcement: skip fetch as we're not tracking {pub_key}");
+                                            }
+                                        }
+                                    } else {
+                                        self.fetch(message.rid, Namespaces::All, announcer)
```
cloudhead, line 889: Only fetch if `policy == Track`

radicle-node/src/service.rs
keepsimple1: Sure, sounds good. Can I rebase after @FintanH's patch got merged in?

radicle-node/src/service.rs
keepsimple1: yeah, the previous access is via `is_repo_tracked()` that hides the repo_entry internally. Let me see what we can do.

keepsimple1: I've refactored this part of code to bind the store result first, and then use it in both places.

radicle-node/src/service.rs
keepsimple1: Seems to me that this is the case of `Receive` in `FetchDirection` , which would have `namespaces` info. No?

cloudhead: Yes, but what I'm trying to say is that the uploader doesn't need to know about the namespaces. The `Fetch` message is sent from the fetcher to the uploader, to initiate a fetch; and the uploader doesn't need to know what namespaces are going to be fetched, since that is negotiated by git. Hence, the `Fetch` message doesn't need to contain the namespaces.

FintanH: Maybe it makes more sense looking at the changes :) https://github.com/radicle-dev/heartwood/pull/395

cloudhead:
Ok, probably time to revisit this, @FintanH 's PRs are merged.

keepsimple1:
> Ok, probably time to revisit this, @FintanH 's PRs are merged.

I've rebased to the latest, and updated the diff accordingly.

radicle-node/src/service.rs
```diff
@@ -266,6 +272,16 @@ where
...
 
+    /// Track a node.
+    /// Returns whether or not the tracking policy was added or updated.
+    pub fn track_node(
+        &mut self,
+        id: &NodeId,
+        alias: Option<&str>,
+    ) -> Result<bool, tracking::Error> {
+        self.tracking.track_node(id, alias)
+    }
```
cloudhead, line 283: What's this function for? It doesn't seem to do anything but borrow all of `&self` instead of just borrowing `self.tracking`

radicle-node/src/service.rs
```diff
@@ -868,15 +891,16 @@ where
                     .insert(message.rid, *relayer, message.timestamp)
                 {
                     if updated {
-                        info!(target: "service", "Routing table updated for {} with seed {relayer}", message.rid);
+                        info!(target: "service", "handle_announcement: Routing table updated for {} with seed {relayer}", message.rid);
```
cloudhead, line 894:
```suggestion
                        info!(target: "service", "Routing table updated for {} with seed {relayer}", message.rid);
```

radicle-node/src/service.rs
```diff
@@ -868,15 +891,16 @@ where
...
                 }
                 // TODO: Buffer/throttle fetches.
-                if self
+                let repo_policy = self
                     .tracking
-                    .is_repo_tracked(&message.rid)
-                    .expect("Service::handle_announcement: error accessing tracking configuration")
-                {
+                    .repo_policy(&message.rid)
+                    .expect("handle_announcement: error accessing repo tracking configuration");
```
cloudhead, line 901:
```suggestion
                    .expect("Service::handle_announcement: error accessing repo tracking configuration");
```

radicle-node/src/service.rs
```diff
@@ -891,14 +915,46 @@ where
...
+                                        _ => self.fetch(message.rid, Namespaces::All, announcer),
+                                    }
                                 }
                             }
                             Err(e) => {
                                 error!(target: "service", "Failed to check ref announcement freshness: {e}");
                             }
                         }
+                    } else {
+                        debug!("handle_announcement: no sessions connected to {announcer}");
```
cloudhead, line 957:
```suggestion
                        debug!(target: "service", "no sessions connected to {announcer} for fetching {}", message.rid);
```

radicle-node/src/service.rs
```diff
@@ -1103,18 +1160,23 @@ where
...
+                    match self.tracking.namespaces_for(&self.storage, &rid) {
+                        Ok(ns) => ns,
+                        Err(err) => {
+                            if let Some(resp) = self.fetch_reqs.get(&rid) {
+                                resp.send(FetchResult::Failed {
+                                    reason: err.to_string(),
+                                })
+                                .ok();
+                            }
+                            return Ok(());
```
cloudhead, line 1175: What is the reason for not always using `tracking.namespaces_for`?

radicle-node/src/service.rs
```diff
@@ -199,6 +200,10 @@ pub struct Service<R, A, S, G> {
     last_announce: LocalTime,
     /// Time when the service was initialized.
     start_time: LocalTime,
+
+    /// Storing the namespaces for fetches. Timestamp seems to be sufficient
+    /// to identify fetches at this point.
+    fetch_namespaces: HashMap<Timestamp, Namespaces>,
```
cloudhead, line 206: Fetches are tracked in the peer session (`session.rs`) already. If you need to add namespace information you can add it there, but I'm not sure why you would need to..

radicle-node/src/service.rs
```diff
@@ -891,14 +915,46 @@ where
...
+                                                    Some(d) => d.is_delegate(pub_key),
+                                                    None => false,
+                                                };
+                                                let is_node_tracked = self
+                                                    .tracking.is_node_tracked(pub_key)
+                                                    .expect("handle_announcement: error accessing node tracking configuration");
+                                                if is_node_tracked || is_delegate {
+                                                    Some(*pub_key)
+                                                } else {
+                                                    debug!("handle_announcement: skip fetch as we're not tracking {pub_key}");
```
cloudhead, line 934: For `expect` error messages, we use the standard `Type::method:` prefix that rust stdlib typically uses, but for logs we include the target only. I've provided examples in other comments.

radicle-node/src/service.rs
```diff
@@ -891,14 +915,46 @@ where
...
+                                                    None
+                                                }
+                                            }).collect();
```
cloudhead, line 937:
```suggestion
                                            }).collect();
                                            
```

radicle-node/src/service.rs
```diff
@@ -891,14 +915,46 @@ where
...
+                                            if let Some(non_empty) = NonEmpty::from_vec(trusted) {
+                                                self.fetch(
+                                                    message.rid,
+                                                    Namespaces::Many(non_empty),
+                                                    announcer,
+                                                );
+                                            } else {
+                                                debug!("handle_announcement: no trusted nodes to fetch");
```
cloudhead, line 945: Logging is only useful if you're going to include some context :)

```suggestion
                                                debug!("handle_announcement: no trusted nodes to fetch {rid} from;");
```

cloudhead: Here we also should also probably respond to any `fetch_reqs`

radicle-node/src/service.rs
```diff
@@ -891,14 +915,46 @@ where
                         match message.is_fresh(&self.storage) {
                             Ok(is_fresh) => {
                                 if is_fresh {
-                                    // TODO: Only fetch if the refs announced are for peers we're tracking.
-                                    self.fetch(message.rid, announcer);
+                                    match repo_policy.scope {
+                                        tracking::Scope::Trusted => {
+                                            // Check every node and only fetch if we're tracking the node.
```
cloudhead, line 920:
```suggestion
                                            // Check every ref and only fetch if we're tracking the node.
```

radicle-node/src/service.rs
```diff
@@ -490,7 +506,7 @@ where
                 // TODO: Fetch requests should be queued and re-checked to see if they can
                 //       be fulfilled everytime a new node connects.
                 self.fetch_reqs.insert(rid, resp);
-                self.fetch(rid, &seed);
+                self.fetch(rid, Namespaces::All, &seed);
```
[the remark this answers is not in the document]
cloudhead: Have the same question/feedback

radicle-node/src/service.rs
```diff
@@ -891,14 +915,46 @@ where
                         match message.is_fresh(&self.storage) {
                             Ok(is_fresh) => {
                                 if is_fresh {
-                                    // TODO: Only fetch if the refs announced are for peers we're tracking.
-                                    self.fetch(message.rid, announcer);
+                                    match repo_policy.scope {
+                                        tracking::Scope::Trusted => {
+                                            // Check every node and only fetch if we're tracking the node.
+                                            let doc = self.get(message.rid)?;
```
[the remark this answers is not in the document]
cloudhead: Yeah that makes sense. Once we have multi-step fetch, we'd only fetch the delegates I guess.

This makes me wonder though: should ref announcements trigger full clones? Currently they do, but it's a mechanism designed to notify peers of ref changes, not of inventory. I think if we want to preserve the current behavior, then we should indeed fetch here -- probably a good idea to confirm current behavior with a test though.

But thinking about this more, we may want to only do updates when receiving ref announcements, to simplify the code.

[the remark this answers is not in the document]
cloudhead: Not exactly: I'm making the distinction between a full clone, and a fetch that updates refs on an existing repo. Ref announcements are meant to trigger the latter, though it's "ok" to use it to trigger full clones as well. But we might be able to simplify our code if we just ignore ref announcements for repos we don't have yet, and let other parts of the code handle that, eg. inventory announcements or track-repo commands. Just a thought.

With regards to fetching refs that changed, I believe git does a good enough job with that, that we don't need to go crazy granular on the fetchspec, as long as we don't fetch namespaces we aren't tracking.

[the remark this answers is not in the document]
FintanH: > Not exactly: I'm making the distinction between a full clone, and a fetch that updates refs on an existing repo. Ref announcements are meant to trigger the latter, though it's "ok" to use it to trigger full clones as well. But we might be able to simplify our code if we just ignore ref announcements for repos we don't have yet, and let other parts of the code handle that, eg. inventory announcements or track-repo commands. Just a thought.

Mmmmm I see what you mean. I guess my question for this is: when do inventory announcements happen? Like what I track a repository to clone it and my client sees a refs announcement -- it can clone at that stage since it heard about a node that has what I want. Otherwise, it has to wait for another inventory announcement.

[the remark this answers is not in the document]
cloudhead: Yeah, I guess the way to think about it is: how do I go from knowing about a repo, to having it locally? Assuming the default tracking policy, I would send `Command::TrackRepo`, right? So when that command is received from the service, it should check the routing table and fetch it from some node.

Now it's not always that simple, for two reasons:
1. What happens if there are no seeds for that RID in the routing table?
2. What happens if you restart your node after calling TrackRepo?

For (1), what will cause us to have a seed in the routing table is an inventory-ann, so waiting for that would solve the problem. The other way is via a ref-announcement actually. We do update the routing table on ref-ann. But the point here is that the fetched would be triggered by the routing table update, not the ref-announcement. It's a subtle difference I guess, but it keeps the logic simpler.

For (2) it seems like we need some kind of check on node startup (?) that checks whether we have our tracked repos locally, or whether they should be fetched still. So eg. loop through tracked repos, for each that isn't in storage, try to fetch from seed.

radicle-node/src/service.rs
keepsimple1: My intention was to be consistent with the existing [`track_repo` method](https://github.com/radicle-dev/heartwood/blob/f8aa1daedb391069847c66e14b2da6cc2e5669ae/radicle-node/src/service.rs#L271). 

I will change it to get a `&mut tracking` only. I'm okay with either way.

radicle-node/src/service.rs
keepsimple1: I can make the change. Just to clarify, the module name `service` appears in the log line already.

radicle-node/src/service.rs
keepsimple1: The problem with always using `namespaces_for` is that: we are not specifying namespaces based on the namespaces found in the `refs_announcement` message.  We do a [db look up](https://github.com/radicle-dev/heartwood/blob/f8aa1daedb391069847c66e14b2da6cc2e5669ae/radicle-node/src/service/tracking.rs#L118) in local db for trusted nodes.  We are blindly sending all trusted node namespaces found in the db to fetch, as a response of `refs_announcement`. 

For a high level, what I am doing are:

1) we will find the namespaces of [trusted nodes](https://github.com/radicle-dev/heartwood/blob/f055c9d4b38c0472dbc6843dbeeb9f7af8b9b5f0/radicle-node/src/service.rs#L923) from `refs_announcement`. 

2) we will [record the namespaces with its fetch id (timestamp)](https://github.com/radicle-dev/heartwood/blob/f055c9d4b38c0472dbc6843dbeeb9f7af8b9b5f0/radicle-node/src/service.rs#L584-L587) .  This is to simplify the message format, i.e. not to encode a list of namespaces into the message itself.

3) when the `FetchOk` comes back, we will use the id(timestamp) to [look up the namespaces and fetch](https://github.com/radicle-dev/heartwood/blob/f055c9d4b38c0472dbc6843dbeeb9f7af8b9b5f0/radicle-node/src/service.rs#L1163) exactly them.

radicle-node/src/service/message.rs
```diff
@@ -333,10 +333,20 @@ pub enum Message {
...
+    },
 
     /// Accept a fetch request.
-    FetchOk { rid: Id },
+    FetchOk {
+        /// Repository Id
+        rid: Id,
+        /// The unique Id for the original fetch request.
+        timestamp: Timestamp,
+    },
```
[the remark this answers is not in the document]
keepsimple1: It is used to retrieve fetch related info, i.e. namespaces.  Please see the design points I listed in this comment: https://github.com/radicle-dev/heartwood/pull/370#discussion_r1133476801.

Sorry I didn't know such change should be discussed earlier. Hopefully we can use this opportunity to discuss them.

radicle-node/src/service.rs
keepsimple1: Did you mean the [`Protocol::Fetch`](https://github.com/radicle-dev/heartwood/blob/f055c9d4b38c0472dbc6843dbeeb9f7af8b9b5f0/radicle-node/src/service/session.rs#L30)?  This one is used to identify the state of the session protocol, not the fetch messages themselves.

When a peer receives `FetchOk` message, the session protocol state is actually [in `Gossip`](https://github.com/radicle-dev/heartwood/blob/f055c9d4b38c0472dbc6843dbeeb9f7af8b9b5f0/radicle-node/src/service.rs#L1147-L1151). It is clearer to carry the id (timestamp) in `FetchOk` message, instead of the protocol state.

radicle-node/src/service.rs
keepsimple1: Got it. Thanks!

radicle-node/src/service.rs
keepsimple1: > Here we also should also probably respond to any `fetch_reqs`

I missed that. I'm wondering why do we need to respond to `fetch_reqs` here in handling `refs_announcement`?

radicle-node/src/service.rs
```diff
@@ -564,6 +580,13 @@ where
             session::FetchResult::Ready(fetch) => {
                 debug!(target: "service", "Fetch initiated for {rid} with {seed}..");
 
+                if !matches!(namespaces, Namespaces::All) {
+                    // Only need to track the fetch if the namespaces is not `All`.
+                    if let Message::Fetch { rid: _, timestamp } = &fetch {
+                        self.fetch_namespaces.insert(*timestamp, namespaces);
+                    }
+                }
```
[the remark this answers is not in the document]
keepsimple1: It was not for the scenario you linked, if I understand your link correctly. But here is the high level logic: https://github.com/radicle-dev/heartwood/pull/370#discussion_r1133476801

radicle-node/src/service.rs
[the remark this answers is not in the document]
keepsimple1: I think [this response](https://github.com/radicle-dev/heartwood/pull/370#discussion_r1133476801) also should answer this one.

radicle-node/src/service.rs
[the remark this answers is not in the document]
keepsimple1: > If the `Doc` is `None` that means we've never fetched this repository before. So I think we need to line up a `fetch` with `Namespaces::All`.

I could add that, but I wanted to understand, are we supposed to receive `refs_announcement` before we have the repo doc?  Maybe I missed, but it is not clear to me if that's a valid path from reading the above. i.e. whether we should report an error, or simply handle it and fetch all?

radicle-node/src/service.rs
```diff
@@ -891,14 +915,46 @@ where
...
                             Ok(is_fresh) => {
                                 if is_fresh {
-                                    // TODO: Only fetch if the refs announced are for peers we're tracking.
-                                    self.fetch(message.rid, announcer);
+                                    match repo_policy.scope {
+                                        tracking::Scope::Trusted => {
+                                            // Check every node and only fetch if we're tracking the node.
+                                            let doc = self.get(message.rid)?;
+
+                                            let trusted: Vec<_> = message.refs.iter().filter_map(|(pub_key, _refs)| {
```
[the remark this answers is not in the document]
keepsimple1: This is addressed in https://github.com/radicle-dev/heartwood/pull/370#discussion_r1133476801

radicle-node/src/test/peer.rs
```diff
@@ -237,7 +238,17 @@ where
     }
 
     pub fn refs_announcement(&self, rid: Id) -> Message {
-        let refs = BoundedVec::new();
+        let mut refs = BoundedVec::new();
+        if let Ok(repo) = self.storage().repository(rid) {
+            if !repo.is_empty().unwrap() {
```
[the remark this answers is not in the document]
keepsimple1: It also surprised me that `is_empty()` returns a `Result`.

radicle-node/src/test/peer.rs
```diff
@@ -237,7 +238,17 @@ where
...
+                if let Ok(remotes) = repo.remotes() {
+                    for (remote_id, remote) in remotes.into_iter() {
+                        refs.push((remote_id, remote.refs.unverified())).unwrap();
```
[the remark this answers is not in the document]
keepsimple1: It was because `refs` is `BoundedVec`. We can get rid of this `unwrap()` if we use a regular `Vec`.

radicle-node/src/service.rs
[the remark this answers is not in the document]
cloudhead: We *may* receive it, but as mentioned above, I'm not 100% sure it's the right thing to do. However, since that appears to be the previous behavior, we should probably preserve it, and do a full fetch to not break existing tests. At the very least we should choose one or the other way deliberately.

radicle-node/src/service.rs
cloudhead: I *think* that makes sense, but even so, I don't see the need to change the `Fetch/FetchOk` messages. There is only one fetch happening at a time per session, so simply storing the namespaces in the session should suffice. As you mentioned, until `FetchOk` is received, we have `Protocol::Gossip { requested: Option<Id> }`, you could simply change that to `requested: Option<(Id, Namespaces)>`.

cloudhead: re: `namespaces_for`: what you're saying is that we may want to only fetch a subset of namespaces, since the announcing node may not have some of our trusted namespaces?

I think it might be ok to let Git handle that, ie. the refs that aren't found on the other side will just be ignored.. This would simplify the service code at almost no cost? cc @FintanH

radicle-node/src/service.rs
cloudhead: Right, see https://github.com/radicle-dev/heartwood/pull/370#discussion_r1133728229

All we need is to modify `Protocol::Gossip` slightly.

radicle-node/src/test/peer.rs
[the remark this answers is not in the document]
FintanH: But the `BoundedVec` is used for a reason, I believe to limit the client from trying to fetch too much information. I think in those scenarios the current approach is to drop the surplus and continue with what you have, but correct me if I'm wrong @cloudhead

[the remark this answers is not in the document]
cloudhead: Yes that's right! If `push` returns an error, we shouldn't continue pushing :)

radicle-node/src/service.rs
FintanH: > re: `namespaces_for`: what you're saying is that we may want to only fetch a subset of namespaces, since the announcing node may not have some of our trusted namespaces?
> 
> I think it might be ok to let Git handle that, ie. the refs that aren't found on the other side will just be ignored.. This would simplify the service code at almost no cost? cc @FintanH

Ya, this is correct :point_up: We just say, "Hey give me the things I'm interested in", if these don't exist then we don't get them and that's fine.

keepsimple1: > This would simplify the service code at almost no cost?

The trusted node list from local db could be much larger than the trusted nodes in `refs_announcements`. For example, there might be only one trusted ref in the announcement. In practice, it would depend on how radicle-node scales.

Again, like I said earlier, it is doable to blindly fetch using trusted nodes list from the local db. I think it's less ideal, and also defeats the purpose of `refs_announcement` because now the `refs_announcement` will behave like this: 

1. Announcer sends `refs_announcement` with a list of refs.
2. Receiver sends back `Fetch<rid>` .
3. Announcer sends  back `FetchOk<rid>`.
4. Receiver sends git fetch with all trusted namespaces from local db.
5. Announcer sends back namespaces it has for trusted nodes. 

It seems to me that the 1st step could just be `inventory_announcement` as well. 

For now, I will check the list of refs in `refs_announcement` anyway, make sure there is at least one trusted ref.

cloudhead: Yeah, I think you have a point in that it's maybe not ideal for large trusted set. Note that as you said, we should anyway check that at least one ref is interesting before requesting a fetch, and also we're able to check that the `oid` of the ref is one we do not already have (currently we don't look at ref oids), which is not something we can do with inventory announcements.

I'd be supportive in storing the matching namespace (basically an intersection between trusted and announced) as part of the session state, to be able to only request what the other peer has.

keepsimple1: > I'd be supportive in storing the matching namespace (basically an intersection between trusted and announced) as part of the session state, to be able to only request what the other peer has.

I've looked into that. Seems to me it has a problem: when we queue pending fetches, we will not be able to obtain this namespace info for a protocol state update when it's time to send out the next pending fetch. Unless we store the info with each fetch, which is back to the original approach.

radicle-node/src/test/peer.rs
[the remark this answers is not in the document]
keepsimple1: Fixed both `unwrap`s. Thanks.

radicle-node/src/test/peer.rs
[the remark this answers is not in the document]
keepsimple1: Fixed now.

radicle-node/src/service.rs
keepsimple1: I've refactored this patch :

1. Removed `Timestamp` in the `Fetch` message.

2. Re-use `namespaces_for()` to get the namespaces for a fetch. Then use them to check against the refs in `refs_announcement`.

3. Fixed `namespace_for()` to support `scope::Trusted` (merged from https://github.com/radicle-dev/heartwood/pull/420) . I understand @FintanH provided comments over there, but I wanted to ask for another review in this context. My point is that we should fetch from the initial remote / seed when the repo didn't exist locally yet.

radicle-node/src/service.rs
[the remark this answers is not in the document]
keepsimple1: Removed now.

radicle-node/src/service.rs
```diff
@@ -868,15 +891,16 @@ where
                     .insert(message.rid, *relayer, message.timestamp)
                 {
                     if updated {
-                        info!(target: "service", "Routing table updated for {} with seed {relayer}", message.rid);
+                        info!(target: "service", "handle_announcement: Routing table updated for {} with seed {relayer}", message.rid);
                     }
                 }
                 // TODO: Buffer/throttle fetches.
-                if self
+                let repo_policy = self
```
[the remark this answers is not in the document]
keepsimple1: Suggestion accepted ;-)

radicle-node/src/service.rs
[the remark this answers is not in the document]
keepsimple1: In the latest update, I chose the following behavior: if repo didn't exist and in `scope::Trusted`, we will treat the remote of the `ref_announcement` a trusted peer. Then fetch from them, only for the namespace of this remote.

If that does not make sense, I'm okay to always fetch `Namespaces::All`.

radicle-node/src/service.rs
```diff
@@ -889,16 +895,61 @@ where
...
+                                                // check if there is at least one trusted ref.
+                                                let node_set: HashSet<_> = nodes.iter().collect();
+                                                message
+                                                    .refs
+                                                    .iter()
+                                                    .filter(|(pub_key, _refs)| {
+                                                        node_set.contains(pub_key)
+                                                    })
+                                                    .count()
+                                                    > 0
```
FintanH, line 918:
```suggestion
                                                message
                                                    .refs
                                                    .iter()
                                                    .any(|(pub_key, _refs)| {
                                                        node_set.contains(pub_key)
                                                    })
```

radicle-node/src/service.rs
```diff
@@ -889,16 +895,61 @@ where
...
+                                            }
+                                            Ok(Namespaces::One(key)) => {
+                                                message
+                                                    .refs
+                                                    .iter()
+                                                    .filter(|(pub_key, _refs)| pub_key == &key)
+                                                    .count()
+                                                    > 0
```
FintanH, line 926: Same here

radicle-node/src/service.rs
```diff
@@ -889,16 +895,61 @@ where
...
                             }
+                            Ok(false) => {
+                                debug!(target: "service", "All refs from {announcer} are already in local node for {}", &message.rid)
+                            }
                             Err(e) => {
                                 error!(target: "service", "Failed to check ref announcement freshness: {e}");
                             }
                         }
+                    } else {
+                        debug!(target: "service", "no sessions connected to {announcer}");
```
FintanH, line 952:
```suggestion
                        debug!(target: "service", "No sessions connected to {announcer}");
```

radicle-node/src/service/tracking.rs
```diff
@@ -121,18 +126,23 @@ impl Config {
...
-                        NonEmpty::from_vec(trusted).map(Namespaces::Many)
                     } else {
-                        Some(Namespaces::All)
+                        // We don't have this repo yet, we should treat the `remote`
+                        // as our seed and include its pub_key(NodeId) in `trusted`.
+                        //
+                        // We are not returning `Namespaces::All` as the `remote`
+                        // might have refs from all kinds of nodes and we might not
+                        // trust all the other nodes.
+                        trusted.push(*remote);
```
FintanH, line 142: One problem with this is that we're not guaranteed `remote` is a delegate. So in the case of them not being, we don't have the set of delegates.

This is fine for a stop-gap solution, but I would say add a FIXME mentioning that and we can sort this out in the multistep fetch code.

radicle-node/src/test/peer.rs
```diff
@@ -237,7 +238,20 @@ where
...
 
     pub fn refs_announcement(&self, rid: Id) -> Message {
-        let refs = BoundedVec::new();
+        let mut refs = BoundedVec::new();
+        if let Ok(repo) = self.storage().repository(rid) {
+            if let Ok(false) = repo.is_empty() {
+                if let Ok(remotes) = repo.remotes() {
+                    for (remote_id, remote) in remotes.into_iter() {
+                        if let Err(e) = refs.push((remote_id, remote.refs.unverified())) {
+                            debug!(target: "test", "failed to push {remote_id} to refs: {e}");
```
FintanH, line 247:
```suggestion
                            debug!(target: "test", "Failed to push {remote_id} to refs: {e}");
```

radicle-node/src/test/peer.rs
```diff
@@ -237,7 +238,20 @@ where
     }
 
     pub fn refs_announcement(&self, rid: Id) -> Message {
-        let refs = BoundedVec::new();
+        let mut refs = BoundedVec::new();
+        if let Ok(repo) = self.storage().repository(rid) {
+            if let Ok(false) = repo.is_empty() {
```
cloudhead, line 243: Is this needed? I'd expect `remotes()` to just not return anything if it's empty.

radicle-node/src/service.rs
```diff
@@ -889,16 +895,61 @@ where
                     // which is required by the protocol to only announce refs it has.
                     if self.sessions.is_connected(announcer) {
                         match message.is_fresh(&self.storage) {
-                            Ok(is_fresh) => {
-                                if is_fresh {
-                                    // TODO: Only fetch if the refs announced are for peers we're tracking.
-                                    self.fetch(message.rid, announcer);
+                            Ok(true) => {
```
cloudhead, line 898: Can we perhaps reduce the level of nesting in this code? It's pretty hard to read. I'd suggest eg. doing
```rust
if !self.sessions.is_connected(announcer) {
   debug!(...);
   return Ok(relay);
}
```
... and that sort of thing where it's easy.

radicle-node/src/test/peer.rs
keepsimple1: yeah it's needed, as some implementation of `remotes()` panics:  https://github.com/radicle-dev/heartwood/blob/de8e92d48cb4810ca7323f4468504d7968dd1a02/radicle/src/test/storage.rs#L116-L117

radicle-node/src/service.rs
keepsimple1: I've extracted the big inner blob into its a convenient method.

radicle-node/src/service.rs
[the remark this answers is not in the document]
cloudhead: I think that works if it isn't breaking existing tests. We'll have a chance to re-evaluate this with fintan's multi-step fetch that is in progress.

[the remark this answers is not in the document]
keepsimple1: > I think that works if it isn't breaking existing tests.

Are you saying, for example, you don't want to have changes in tests like this one? : https://github.com/radicle-dev/heartwood/blob/68f7c6e83342687c54c28104be958cad8570dfb1/radicle-cli/tests/commands.rs#L499-L506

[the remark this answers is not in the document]
keepsimple1: Meantime, I will assume that we don't want any changes of the existing tests, and I will reverse some changes of `namespaces_for` function, i.e. for `scope::Trusted`, if the repo didn't exist yet, we will fetch using `Namespaces::All`. I will keep the changes minimum so we can move this PR forward.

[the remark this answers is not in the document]
keepsimple1: I've updated the diff to keep the existing behavior with regarding `namespaces_for`, and hence reduced the overall changes. Please take a look.

radicle-node/src/service/tracking.rs
keepsimple1: This change is removed.

radicle-node/src/tests.rs
```diff
@@ -600,6 +600,61 @@ fn test_refs_announcement_relay() {
...
+        "eve",
+        [8, 8, 8, 8],
+        Storage::open(tmp.path().join("eve")).unwrap(),
+        peer::Config::default(),
+    );
+
+    let bob = {
+        let mut rng = fastrand::Rng::new();
+        let signer = MockSigner::new(&mut rng);
+        let storage = fixtures::storage(tmp.path().join("bob"), &signer).unwrap();
```
cloudhead, line 624: There shouldn't be a need to create real storages for any of the peers, unless you need these peers to actually have the inventory?

You can simply generate an rid with `arbitrary::gen::<Id>(1)` and use that for bob's `ref_announcement` call. Alice doesn't need to know that bob actually doesn't have the inventory.

This would simplify the test a lot as you can just use `Peer::new`

radicle-node/src/tests.rs
```diff
@@ -600,6 +600,61 @@ fn test_refs_announcement_relay() {
...
+    let bob_inv = bob.inventory();
+
+    alice
+        .track_repo(&bob_inv[0], tracking::Scope::Trusted)
+        .unwrap();
+
+    alice.connect_to(&bob);
+    alice.connect_to(&eve);
+
+    alice.receive(eve.id(), Message::Subscribe(Subscribe::all()));
```
cloudhead, line 646: Was this required to make the test pass?

radicle-node/src/tests.rs
```diff
@@ -600,6 +600,61 @@ fn test_refs_announcement_relay() {
...
+
+    // Alice receives refs_announcement from Bob.
+    alice.receive(bob.id(), bob.refs_announcement(bob_inv[0]));
+
+    assert_matches!(
+        alice.messages(bob.id()).next(),
+        Some(Message::Fetch { rid }) if rid == bob_inv[0],
+        "A fetch is sent to Bob as this is a new repo for Alice"
+    );
```
cloudhead, line 655: I'm not sure what we're testing here.. Would this test fail without the changes you made? Afaict it would pass because we were always fetching previously. If so, we should additionally check that Alice doesn't fetch based on Eve's announcement no?

radicle-node/src/tests/e2e.rs
```diff
@@ -379,3 +383,78 @@ fn test_connection_crossing() {
...
+        .unwrap();
+
+    // Alice will not fetch Bob's refs as Alice does not track Bob.
+    assert!(alice
+        .storage
+        .repository(rid)
+        .unwrap()
+        .remote(&bob.id)
+        .is_err());
+}
```
cloudhead, line 460: I think this test is testing some of the stuff I was suggesting for the other test. I would suggest that we just have the other test, as the e2e tests are very slow and only really useful if we want to test the networking protocols. A service logic test should be enough since we only modified the service.

radicle-node/src/tests.rs
keepsimple1: This test only covers the basic positive case, which indeed was covered by existing other tests as well. I will remove this test.

radicle-node/src/tests/e2e.rs
keepsimple1: I don't have accurate measurements but this e2e test runs pretty fast on my not-so-fast laptop.  The reason I am keeping this test is that it is much more easier to trigger the path I wanted:  Bob announces its refs after everyone already has the repo. (I was struggling to write a no-super-complicated test case to cover such path in the service `tests.rs`) .

cloudhead: Hmm I don't see why that would be complicated. You would just call `alice.receive(bob_refs)` and `eve.receive(bob_refs)`.

The service tests are preferred for logic tests as:
* There are no false negatives (e2e tests can fail for a number of reasons)
* They run in 1/100 of the time
* It's much clearer to see what is actually being tested vs. what is just a side-effect of setting up the test

radicle-node/src/service.rs
```diff
@@ -975,6 +983,48 @@ where
...
     }
 
+    /// A convenient method to check if we should fetch from a `RefsAnnouncement`
+    /// with `scope`.
+    fn should_fetch_refs_in_announcement(
+        &self,
+        message: &RefsAnnouncement,
+        scope: &tracking::Scope,
+    ) -> Result<bool, session::Error> {
+        let should_fetch = match scope {
```
FintanH, line 993: nit: I personally wouldn't bother with the `let` and just have the match arms result in an `Ok` or `Err`. That saves you from doing `return Err` below.

FintanH (commented):
Should there be a test where a peer is tracking Alice and hears a refs announcement and then ends up having fetched the repository? Perhaps I'm wrong, but I don't _think_ the existing test does that.

radicle-node/src/tests/e2e.rs
keepsimple1: Suppose `alice` is the first node that creates the repo, how will `bob` create its own refs? One way I know is `clone`, which is done easily in `e2e` test: `NodeHandle.rad("clone", ..)`.  Is there a different way to do `clone` in the service test?

keepsimple1:
> Should there be a test where a peer is tracking Alice and hears a refs announcement and then ends up having fetched the repository? Perhaps I'm wrong, but I don't _think_ the existing test does that.

This is one of the tests : https://github.com/radicle-dev/heartwood/blob/73221e43d781f17aa8ef1830f10b56c14a3d93a5/radicle-cli/tests/commands.rs#L471 

The reason is that now the default scope is `scope::Trusted`, hence some existing tests would go through that path.

radicle-node/src/service.rs
keepsimple1: In this particular case, I found `should_fetch` seems clearer to me. Again, this is only my personal preference. Either would work.

radicle-node/src/tests.rs
keepsimple1: The path I wanted to test is:  

`alice` already has `repo A` in its storage.  Then `alice` gets `bob`'s `refs_ann`, and `alice` should not fetch because `bob` is not trusted yet. 

I was not able to create such case in this test, unless I follow the same idea I used in `e2e.rs`.  Are you saying I should be able to mimic such case without creating actual storages and repos?

radicle-node/src/tests.rs
```diff
@@ -669,6 +669,96 @@ fn test_refs_announcement_relay() {
...
+    let bob_inv = bob.storage().inventory().unwrap();
+    let rid = bob_inv[0];
+
+    let mut alice = {
+        let mut rng = fastrand::Rng::new();
+        let signer = MockSigner::new(&mut rng);
+        let storage = fixtures::storage(tmp.path().join("alice"), &signer).unwrap();
+
+        // Create a repo on Alice using the same `rid`.
+        storage.create(rid).unwrap();
```
keepsimple1, line 745: This is my poor attempt to create the repo with the given `rid` on `alice`, before getting `refs_ann` from `bob`.  It is still not fulling working even the test passed, as there is `identity branch is missing` in the log. 

I believe there probably is a much simpler way but I didn't find it yet. My brain doesn't seem to work well with these tests. :(

radicle-node/src/tests/e2e.rs
cloudhead: What I'm suggesting in the other comment is that Bob doesn't need to actually have the refs. We can forge a message from Bob, with whatever refs we want, and have Bob sign it, without Bob actually having the inventory. This is usually done with `Peer::refs_announcement(rid)`. If for some reason that function doesn't do what you want it to do, I'd like to understand better, and we'd want to manually create the refs you need for the test.

radicle-node/src/tests.rs
cloudhead: There is a `fixtures::project` that should create a new project properly. But as I mentioned elsewhere, it may not be needed.

radicle-node/src/service.rs
keepsimple1: One specific reason is that, this `match` handles a subset of errors differently, so I cannot use `?` to early return from calling `namespaces_for()`, otherwise it will better & simpler to return from each match arm.

radicle-node/src/tests/e2e.rs
keepsimple1: Sorry about all the confusions. Like I mentioned in zulip,  my intention is to test this path:  

Both peer A and peer B has repo R and their own refs.  Peer A does not fetch B's refs because A does not track B yet.  Later, peer A would track peer B and able to fetch B's refs.

So far, I can only came up with the test I've added in `e2e.rs`, which is pretty straightforward after today's updates, and it covers the scenarios above.  I agree that this test is not super fast, although it seems to be ok on my laptop.

I tried to write a different & easier / faster test in `tests.rs`, and so far not able to.  (I avoided to make the test super complicated as that will defeat the original purpose.)  But again I would be wrong playing with the test infra.  I will update later either here or separately if I learn to write a better test for this case.

radicle-node/src/service.rs
```diff
@@ -975,6 +985,48 @@ where
...
+                    Err(e) => {
+                        error!(target: "service", "failed to obtain namespaces: {e}");
+                        return Err(e).map_err(session::Error::Namespaces);
+                    }
+                }
+            }
+        };
+
+        Ok(should_fetch)
+    }
```
FintanH, line 1028: I'm not sure what you mean [here](https://github.com/radicle-dev/heartwood/pull/370/files#r1143965771) but this is what I'm talking about:
```suggestion
        match scope {
            tracking::Scope::All => Ok(true),
            tracking::Scope::Trusted => {
                match self.tracking.namespaces_for(&self.storage, &message.rid) {
                    Ok(Namespaces::All) => Ok(true),
                    Ok(Namespaces::Many(nodes)) => {
                        // Get the set of trusted nodes except self.
                        let my_id = self.node_id();
                        let node_set: HashSet<_> =
                            nodes.iter().filter(|key| *key != &my_id).collect();

                        // Check if there is at least one trusted ref.
                        Ok(message
                            .refs
                            .iter()
                            .any(|(pub_key, _refs)| node_set.contains(pub_key)))
                    }
                    Ok(Namespaces::One(key)) => {
                        Ok(message.refs.iter().any(|(pub_key, _refs)| pub_key == &key))
                    }
                    Err(NamespacesError::NoTrusted { rid }) => {
                        debug!(target: "service", "no trusted nodes to fetch {}", &rid);
                        Ok(false)
                    }
                    Err(e) => {
                        error!(target: "service", "failed to obtain namespaces: {e}");
                        Err(e.into())
                    }
                }
            }
        }
```

FintanH (approved):
This looks good to me

radicle-node/src/service.rs
cloudhead: This is simpler, I agree.

keepsimple1: Updated.

radicle-node/src/tests.rs
keepsimple1: I've updated diff, and explained at [here ](https://github.com/radicle-dev/heartwood/pull/370#discussion_r1143973195).  For `fixtures::project` particularly, it creates a new project with a new Id, not a given Id. Hence I was not using it.

radicle/src/test/storage.rs
```diff
@@ -11,24 +12,40 @@ use crate::identity::IdentityError;
 
 pub use crate::storage::*;
 
+use self::refs::SignedRefs;
+
 #[derive(Clone, Debug)]
 pub struct MockStorage {
     pub path: PathBuf,
     pub inventory: HashMap<Id, Doc<Verified>>,
+    pub remotes: HashMap<Id, Vec<Remote>>,
```
keepsimple1, line 21: I think using `HashMap<PublicKey, SignedRefs>` is probably better than `Vec<Remote>` here. But I'm running out of time today to improve on that. I'm happy to refactor it in a new PR if needed.

keepsimple1:
I finally was able to construct a relatively simple test in `tests.rs` and removed the more involved test in `tests/e2e.rs`.  

The only thing is that I had to add a new field `MockStorage.remotes` so that we can add mocked refs into the storage's repository. Then `RefsAnnouncement` will be generated with "real" refs, which helped the test.

FintanH (commented):
Nice! This looks like it solves the discussion from what I can tell :)

radicle-node/src/tests.rs
```diff
@@ -669,6 +669,101 @@ fn test_refs_announcement_relay() {
...
+    );
+
+    // Alice starts to track Bob.
+    let (sender, receiver) = chan::bounded(1);
+    alice.command(Command::TrackNode(bob.id, Some("bob".to_string()), sender));
+    let policy_change = receiver.recv().map_err(runtime::HandleError::from).unwrap();
+    assert!(policy_change);
+
+    // Bob announces refs again.
+    bob.elapse(LocalDuration::from_mins(1));
```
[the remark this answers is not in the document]
keepsimple1: To generate a fresh refs_ann, otherwise it will be ignored here: https://github.com/radicle-dev/heartwood/blob/da9d63b284bf9186eb7145641c7504646a0125a1/radicle-node/src/service.rs#L797-L800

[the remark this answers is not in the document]
FintanH: Thanks! That’s good to know 😊

keepsimple1:
Rebased to the latest master, and addressed review comments, and refactored `is_fresh` check into the `should_fetch_refs_announcement` .

radicle/src/test/storage.rs
keepsimple1: I've refactored this using `HashMap<>` instead, which I think is better, including the correctness: not to worry about the uniqueness of node Id of the remotes.

cloudhead (commented):
LGTM!

PR #371, opened 2023-03-01
cob: Don't require a local fork to exist

cloudhead:
Before this change, the identity document would be loaded for the COB signer. This meant that a fork would be needed for that signer.

After this change, it's no longer necessary, since the identity doc head is computed from available remotes.

While making those changes, it was also apparent that some of the identity-related functions had bad names and error types, this was fixed as well.


radicle/src/storage/git.rs
```diff
@@ -105,21 +74,21 @@ impl ReadStorage for Storage {
...
+    fn contains(&self, rid: &Id) -> Result<bool, IdentityError> {
         if paths::repository(&self, rid).exists() {
             let _ = self.repository(*rid)?.head()?;
             return Ok(true);
         }
         Ok(false)
     }
 
-    fn get(&self, remote: &RemoteId, proj: Id) -> Result<Option<Doc<Verified>>, ProjectError> {
+    fn get(&self, remote: &RemoteId, proj: Id) -> Result<Option<Doc<Verified>>, IdentityError> {
```
FintanH, line 85:
```suggestion
    fn get(&self, remote: &RemoteId, rid: Id) -> Result<Option<Doc<Verified>>, IdentityError> {
```

FintanH (approved):
Looks good :+1: Few places I noticed we could use the `rid` naming instead.

cloudhead:
@FintanH There are so many places still, I think we'll have to do a PR just for that eventually, and same for NodeId

PR #372, opened 2023-03-01
`rad node`

FintanH:
See the commits for more detail on the changes.

@cloudhead the `start` and `stop` commands are still a bit fuzzy for me. I'm not sure how they're expected to work :thinking:

### DIFF
--- a/radicle-cli/examples/rad-node.md
+++ b/radicle-cli/examples/rad-node.md
@@ -1,0 +1,88 @@
+The radicle node is our daemon friend that is running as a background
+process. It allows us to interact with the network as well as storing
+some key data that we may be interested in.
+
+<details>
+<summary>🚧 `rad node start` is under construction 🚧</summary>
+If the node is not running we can start it by using the `rad node
+start` command:
+
+<!-- ``` -->
+<!-- $ rad node start -->
+<!-- ``` -->
+</details>
+
+We can confirm the status of the node at any time by using the `rad
+node status` command (or just `rad node` for short):
+
+```
+$ rad node status
+✓ The node is running
+```
+
+The node also allows us to connect with other nodes in the
+network. For example, let's connect to a known peer using their Node
+ID and their network address:
+
+```
+$ rad node connect z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk 0.0.0.0:3679
+✓ Connecting to z6Mkt67…v4N1tRk@0.0.0.0:3679...
+```
+
+The node also allows us to query data that it has access too such as
+the tracking relationships and the routing table. Before we explore
+those commands we'll first track a peer so that we have something to
+see.
+
+```
+$ rad track z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk --alias Bob
+Establishing 🌱 tracking relationship for heartwood
+
+✓ Tracking relationship with Bob (z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk) established
+! Warning: fetch after track is not yet supported
+```
+
+Now, when we use the `rad node tracking` command we will see
+information for repositories that we track -- in this case a
+repository that was already created:
+
+```
+$ rad node tracking
+RID                               Scope Policy
+---                               ----- ------
+rad:z42hL2jL4XNk6K8oHQaSWfMgCL7ji all   track
+```
+
+This is the same as using the `--repos` flag, but if we wish to see
+which nodes we are specifically tracking, then we use the `--nodes`
+flag:
+
+```
+$ rad node tracking --nodes
+DID                                                      Alias Policy
+---                                                      ----- ------
+did:key:z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk Bob   track
+```
+
+To see the routing table we can use the `rad node routing` command and
+see what Repository IDs match up with the interests of which Node
+IDs. In this case, it is just our own Node ID for the project we
+created.
+
+```
+$ rad node routing
+RID                               NID
+---                               ---
+rad:z42hL2jL4XNk6K8oHQaSWfMgCL7ji z6MknSL…StBU8Vi
+```
+
+<details>
+<summary>🚧 `rad node stop` is under construction 🚧</summary>
+Finally, if we want to stop the daemon process from running we can
+issue the `rad node stop` command:
+
+<!-- ``` -->
+<!-- $ rad node stop -->
+<!-- Stopping the node... -->
+<!-- ``` -->
+</details>
--- a/radicle-cli/src/commands/node.rs
+++ b/radicle-cli/src/commands/node.rs
@@ -1,0 +1,161 @@
+use std::ffi::OsString;
+
+use anyhow::anyhow;
+use radicle::node::{Address, Node, NodeId};
+
+use crate::terminal as term;
+use crate::terminal::args::{Args, Error, Help};
+
+#[path = "node/control.rs"]
+mod control;
+#[path = "node/routing.rs"]
+mod routing;
+#[path = "node/tracking.rs"]
+mod tracking;
+
+pub const HELP: Help = Help {
+    name: "node",
+    description: "Control and query the Radicle Node",
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad node status
+    rad node (start|stop)
+    rad node connect <nid> <addr>
+    rad node routing
+    rad node tracking [--repos|--nodes]
+
+Options
+
+    --help          Print help
+    --repos         Show the tracked repositories table
+    --nodes         Show the tracked nodes table
+"#,
+};
+
+pub struct Options {
+    op: Operation,
+}
+
+pub enum Operation {
+    Connect { nid: NodeId, addr: Address },
+    Routing,
+    Start,
+    Status,
+    Stop,
+    Tracking { mode: TrackingMode },
+}
+
+#[derive(Default)]
+pub enum TrackingMode {
+    #[default]
+    Repos,
+    Nodes,
+}
+
+#[derive(Default)]
+pub enum OperationName {
+    Connect,
+    Routing,
+    Start,
+    #[default]
+    Status,
+    Stop,
+    Tracking,
+}
+
+impl Args for Options {
+    fn from_args(args: Vec<OsString>) -> anyhow::Result<(Self, Vec<OsString>)> {
+        use lexopt::prelude::*;
+
+        let mut parser = lexopt::Parser::from_args(args);
+        let mut op: Option<OperationName> = None;
+        let mut tracking_mode = TrackingMode::default();
+        let mut nid: Option<NodeId> = None;
+        let mut addr: Option<Address> = None;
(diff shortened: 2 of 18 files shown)


cloudhead:
Yeah I think they need more thinking. For now simply being able to query the node state will be super useful.

radicle-cli/examples/rad-node.md
```diff
@@ -0,0 +1,99 @@
...
+
+<details>
+<summary>🚧 `rad node start` is under construction 🚧</summary>
+If the node is not running we can start it by using the `rad node
+start` command:
+
+<!-- ``` -->
+<!-- $ rad node start -->
+<!-- ``` -->
+</details>
```
FintanH, line 13: Leaving these here but hiding them behind the collapsible element of markdown

radicle-cli/src/commands/node.rs
```diff
@@ -0,0 +1,165 @@
...
+pub const HELP: Help = Help {
+    name: "node",
+    description: "Control and query the Radicle Node",
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad node status
+    rad node (start|stop)
+    rad node connect <NID> <ADDR>
```
FintanH, line 27: Do we capitalise these or have them lowercase?

radicle-cli/src/commands/node.rs
```diff
@@ -0,0 +1,165 @@
...
+    rad node routing
+    rad node tracking [--repos|--nodes|--all]
```
FintanH, line 29: Note to self: document these flags

radicle-cli/src/commands/node/status.rs
```diff
@@ -0,0 +1 @@
+
```
FintanH, line 1: Note to self: remove

radicle-node/src/service.rs
```diff
@@ -282,8 +284,8 @@ where
         self.filter = Filter::new(
             self.tracking
                 .repo_entries()?
-                .filter(|(_, _, policy)| *policy == tracking::Policy::Track)
-                .map(|(e, _, _)| e),
+                .filter(|t| t.policy == tracking::Policy::Track)
+                .map(|t| t.id),
```
FintanH, line 288: Note to self: `filter_map`

radicle-node/src/service.rs
```diff
@@ -364,8 +366,8 @@ where
...
```
FintanH, line 370: Same

radicle/src/node/tracking.rs
```diff
@@ -0,0 +1,160 @@
+use std::fmt;
+use std::str::FromStr;
+
+use sqlite as sql;
```
FintanH, line 4: Need to use `sql` feature

radicle/src/node/tracking.rs
```diff
@@ -0,0 +1,160 @@
...
+    Default,
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    serde::Serialize,
+    serde::Deserialize,
```
FintanH, line 38: Import to ease cloudhead's mental state

radicle/src/node/tracking.rs
```diff
@@ -0,0 +1,160 @@
...
+        match self {
+            Scope::Trusted => f.write_str("trusted"),
+            Scope::DelegatesOnly => f.write_str("delegates"),
+            Scope::All => f.write_str("all"),
+        }
+    }
+}
+
+impl FromStr for Scope {
+    type Err = ();
```
FintanH, line 120: Use error type for next PR

radicle-cli/src/commands/node.rs
cloudhead: Lowercase :)

radicle-cli/examples/rad-node.md
```diff
@@ -0,0 +1,99 @@
...
+✓ Tracking relationship with Bob (z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk) established
+! Warning: fetch after track is not yet supported
+```
+
+Now, when we use the `rad node tracking` command we will see
+information for repositories that we track -- in this case a
+repository that was already created, and other nodes that we track:
+
+```
+$ rad node tracking
```
cloudhead, line 51: I think it's a little weird to output two tables; especially once they start being a bit bigger it won't be super useful. I'd suggest we make `--repos` or `--nodes` required, or have it default to repos if no flag is passed.

cloudhead: The `rad node *` commands are meant to be a little more on the "plumbing" side btw, so it's ok to be more rigid here.

radicle-cli/examples/rad-node.md
```diff
@@ -0,0 +1,99 @@
...
+node status` command (or just `rad node` for short):
+
+```
+$ rad node status
+✓ The node is running
+```
+
+The node also allows us to connect with other nodes in the
+network. For example, let's connect to a known peer using their Node
+ID and their network Address:
```
cloudhead, line 25:
```suggestion
ID and their network address:
```

radicle-cli/examples/rad-node.md
```diff
@@ -0,0 +1,99 @@
...
+<!-- $ rad node start -->
+<!-- ``` -->
+</details>
+
+We can confirm the status of the node at any time by using the `rad
+node status` command (or just `rad node` for short):
+
+```
+$ rad node status
+✓ The node is running
```
cloudhead, line 20: Would be cool to show the listen address here if it's easy; as well as the node socket file. Otherwise maybe in another PR.

radicle-cli/src/commands/node/control.rs
```diff
@@ -0,0 +1,51 @@
...
+        term::warning(&format!("Error occurred while shutting down node: {err}"));
+    } else {
+        spinner.finish();
+        term::success!("Successfully shut down the node")
+    }
+    Ok(())
+}
+
+pub fn connect(node: &mut Node, nid: NodeId, addr: Address) -> anyhow::Result<()> {
+    let spinner = term::spinner("Connecting to node...");
```
cloudhead, line 23: I'd either just say `Connecting...` or be a bit more specific and say `Connecting to {nid}@{addr}...`

radicle-cli/examples/rad-node.md
FintanH: Ok, that makes sense :+1:

radicle-cli/examples/rad-node.md
FintanH: I'd say let's do that in a follow-up since status is just using `is_running`, but maybe we'd want something more like a `status` method in `Handle` which has more verbose info.

radicle-cli/src/commands/node/control.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub fn start() -> anyhow::Result<()> {
+    todo!()
+}
+
+pub fn stop(node: Node) -> anyhow::Result<()> {
+    let spinner = term::spinner("Stopping the node...");
+    if let Err(err) = node.shutdown() {
+        spinner.finish();
+        term::warning(&format!("Error occurred while shutting down node: {err}"));
```
cloudhead, line 14: You can use `spinner.error()` or `spinner.warning()` here, instead of the `term::warning`

radicle-cli/src/commands/node/control.rs
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+        spinner.finish();
+        term::success!("Successfully shut down the node")
```
cloudhead, line 17:
```suggestion
```
I think it's redundant because we'll see the checkmark next to `Stopping the node...` already :)

radicle-cli/src/commands/node/control.rs
```diff
@@ -0,0 +1,54 @@
...
+            &err.into(),
+        )
+    } else {
+        spinner.finish();
+        term::success!(
+            "Successfully connected to {}@{}",
+            term::format::node(&nid),
+            term::format::secondary(addr)
+        )
+    }
```
cloudhead, line 44: Same comments as above, the spinner should be used to display success/failure messages instead of a separate line, because when you call `finish()` it will add a checkmark to the spinner already.

radicle-cli/src/terminal/args.rs
```diff
@@ -95,3 +96,10 @@ pub fn rid(val: &OsString) -> anyhow::Result<Id> {
     let val = val.to_string_lossy();
     Id::from_str(&val).map_err(|_| anyhow!("invalid Repository ID '{}'", val))
 }
+
+pub fn addr(val: &OsString) -> anyhow::Result<Address> {
+    let val = val.to_string_lossy();
+    Address::from_str(&val)
+        .map(Address::from)
```
cloudhead, line 103: :eyes:

FintanH: :facepalm:

radicle-cli/src/commands/node/control.rs
FintanH: Ah thanks! I missed these methods :)

radicle/src/node.rs
```diff
@@ -534,8 +583,13 @@ impl Handle for Node {
...
 
-    fn routing(&self) -> Result<chan::Receiver<(Id, NodeId)>, Error> {
-        todo!();
+    fn routing(&self) -> Result<Self::Routing, Error> {
+        let mut routes = Vec::new();
+        for result in self.call::<&str, _>(CommandName::Routing, [])? {
+            let route = result?;
+            routes.push(route);
+        }
+        Ok(routes)
```
cloudhead, line 592: Should we attempt to return an iterator here? Maybe add a TODO otherwise?

cloudhead: Same question as the functions above.

radicle/src/node.rs
```diff
@@ -449,8 +466,22 @@ impl Handle for Node {
...
+                cmd: CommandName::Connect,
+            })??;
+        match result {
+            CommandResult::Okay { .. } => {
+                log::info!(target: "node", "Successfully connected to {} {}", nid, addr)
+            }
+            CommandResult::Error { reason } => {
+                log::warn!(target: "node", "Failed to connect to {} {}: {}", nid, addr, reason)
+            }
+        }
```
cloudhead, line 483: I'd remove these log statements because actually the connection will happen asynchronously, and could take some time to establish. Right now unfortunately the command is not designed to wait for the connection to be established, though that's something we should implement at some point.

radicle-cli/src/commands/node/control.rs
FintanH: So when I have this code:
```rust
        spinner.message(format!(
            "Successfully connected to {}@{}",
            term::format::node(&nid),
            term::format::secondary(addr)
        ));
        spinner.finish()
 ```
 It only displays:
 ```
 ✓ Successfully connected to z6Mkt67…v4N1tRk@0.0.0.0:3679
 ```
 So it's not showing the:
 ```
 ✓ Connecting to z6Mkt67…v4N1tRk@0.0.0.0:3679...
 ```
 Is this a bug?
 
 The `Spinner` API has me a bit confused. What's the difference between `failed` and `error`? And why does `warn` not take any progress message?

cloudhead (approved):
Excellent!

PR #385, opened 2023-03-06
Sign COBs, simplify use

cloudhead:
*See commits*


cloudhead:
@FintanH I found it odd that `cob::remove` doesn't need a signer, but it actually makes sense, since you're just removing refs.. However is it correct that we need to sign refs after removal?

FintanH:
> @FintanH I found it odd that `cob::remove` doesn't need a signer, but it actually makes sense, since you're just removing refs.. However is it correct that we need to sign refs after removal?

Aha, I was looking at your other PR but I knew I'd seen this message somewhere but couldn't think of where :sweat_smile: 

Yes and yes, that makes sense to me :+1: I would say we sign refs after removal since we want the state to say that those refs no longer exist on our side.

PR #386, opened 2023-03-06
`rad init` fixes

cloudhead:
*See commits*

### DIFF
--- a/radicle-cli/src/commands/init.rs
+++ b/radicle-cli/src/commands/init.rs
@@ -35,6 +35,7 @@ Options
     --setup-signing      Setup the radicle key as a signing key for this repository
     --announce           Announce the new project to the network
     --no-confirm         Don't ask for confirmation during setup
+    --verbose, -v        Verbose mode
     --help               Print help
 "#,
 };
@@ -49,6 +50,7 @@ pub struct Options {
     pub setup_signing: bool,
     pub set_upstream: bool,
     pub announce: bool,
+    pub verbose: bool,
     pub track: bool,
 }
 
@@ -67,6 +69,7 @@ impl Args for Options {
         let mut setup_signing = false;
         let mut announce = false;
         let mut track = true;
+        let mut verbose = false;
 
         while let Some(arg) = parser.next()? {
             match arg {
@@ -117,6 +120,9 @@ impl Args for Options {
                 Long("no-track") => {
                     track = false;
                 }
+                Long("verbose") | Short('v') => {
+                    verbose = true;
+                }
                 Long("help") => {
                     return Err(Error::Help.into());
                 }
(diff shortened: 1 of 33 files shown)


radicle-node/src/service.rs
```diff
@@ -359,9 +359,18 @@ where
         for (id, addr) in addrs {
             self.connect(id, addr);
         }
-        // Ensure that our inventory is recorded in our routing table.
-        for id in self.storage.inventory()? {
-            self.routing.insert(id, self.node_id(), time.as_millis())?;
+        // Ensure that our inventory is recorded in our routing table, and we are tracking
+        // all of it. It can happen that inventory is not properly tracked if for eg. the
+        // user creates a new repository while the node is stopped.
```
slack-coder, line 364: 'tracked if for eg.' -> 'tracked eg.'?

radicle-cli/src/commands/init.rs
```diff
@@ -35,6 +35,7 @@ Options
     --setup-signing      Setup the radicle key as a signing key for this repository
     --announce           Announce the new project to the network
     --no-confirm         Don't ask for confirmation during setup
+    --verbose, -v        Verbose mode
     --help               Print help
```
slack-coder, line 39: For consistency, the shortened version should precede the longer one.  Or the other commands should be updated.

From `rad review`:
```
Options
    -r, --revision <number>   Revision number to review, defaults to the latest
        --[no-]sync           Sync review to seed (default: sync)
    -m, --message [<string>]  Provide a comment with the review (default: prompt)
        --no-message          Don't provide a comment with the review
        --help                Print help
```

radicle-node/src/service.rs
cloudhead: Hmm that reads differently, it's only true *if* the user create a new repo while the node is stopped..

slack-coder: Ok true, its more about 'eg.' being a synonym (?) but not direct replacement for 'example'.  So 'tracked if eg.' or 'tracked if for example'.

FintanH:
I have some feedback on commit messaging. Since the description says "See commits", I was going through them and was looking at the one titled: "Remove need to pass public key in cob store". It would have been really helpful for me to have a description of _why_ we could remove the public key. I was looking through the commit trying to figure out the reason until I spotted that the `remove` taking the signer could have been the reason -- could you confirm that? :)

I think it's useful to get into the habit of being more descriptive for larger commits, and would be happy to be called out on mine too if they're lacking in that department.

FintanH (approved):
Looks good :+1:

cloudhead:
Yeah, will make more of an effort if it's useful to someone :+1:

PR #388, opened 2023-03-06
Some fixes

cloudhead:


radicle-node/src/wire/protocol.rs
```diff
@@ -568,21 +568,37 @@ where
...
+                            io::ErrorKind::ConnectionReset,
+                        )));
+
+                        if let Some(id) = peer.id() {
+                            self.service.disconnected(*id, &reason);
+                        }
+                        peer.disconnected(reason);
+                    }
+                    None => {
+                        log::warn!(target: "wire", "Peer with fd {fd} is unknown");
```
slack-coder, line 599: Hmm is fd the file descriptor?  Has it been useful with diagnosing a node?  I haven't seen the file descriptor used in logs in other services.

radicle-cli/src/commands/patch/create.rs
```diff
@@ -34,19 +34,19 @@ pub fn handle_patch_message(
         .message()
         .ok_or(anyhow!("commit summary is not valid UTF-8; aborting"))?;
     let message = message.get(&format!("{commit_message}{PATCH_MSG}"));
+    let message = message.replace(PATCH_MSG.trim(), ""); // Delete help message.
```
slack-coder, line 37: Nice, though we should eventually improve `message.get()` to handle this

slack-coder:
otherwise :+1:

radicle-node/src/wire/protocol.rs
cloudhead: Yeah. We use it everywhere in the wire protocol logs, as it's the unique identifier for a given peer.

radicle-cli/src/commands/patch/create.rs
cloudhead: Agreed.

radicle-node/src/wire/protocol.rs
```diff
@@ -568,21 +568,37 @@ where
                 // TODO: This should be a fatal error, there's nothing we can do here.
                 log::error!(target: "wire", "Can't poll connections: {}", err);
             }
-            reactor::Error::ListenerPollError(id, err) => {
+            reactor::Error::ListenerPollError(id, _) => {
                 // TODO: This should be a fatal error, there's nothing we can do here.
-                log::error!(target: "wire", "Received error: listener {} disconnected: {}", id, err);
+                log::error!(target: "wire", "Received error: listener {} disconnected", id);
```
FintanH, line 573: Were the errors too much noise or not useful?

cloudhead: They actually weren't errors, was a bad variable name, they are the event state of the file descriptor, which is kind of useless.

PR #395, opened 2023-03-07
node: use enum for fetch direction

### LINKED ISSUE #379: Better differentiate `fetch` directions
The [`fetch`](https://github.com/radicle-dev/heartwood/blob/master/radicle-node/src/service/reactor.rs#L124-L140) function in the `reactor`, confusingly, accepts a `Namespaces` when `initiated` is set to `false`, meaning that it's the `upload-pack` side of a fetch -- which means that it will provide the references that the `receive-pack` asked for.

We should differentiate the two directions, probably with something along the lines of:
```rust
pub enum FetchDirection {
    Receive(Namespaces),
    Upload,
}
```

FintanH:
A fetch is a client-server interaction and so there are two sides to this interaction. These sides are generally referred to as receive -- the client side -- and upload -- the server side.

The `Fetch` type indicated whether it was doing a receive or upload by using a field `initiated: bool`. However, this can be confusing when you pair it with the `namespaces: Namespaces` field, since only the receive side should be sending what refspecs it wants and the upload side responds with those matching refspecs.

A better way to represent this is using an enum for which direction the fetch is being considered, where the receive side contains the Namespaces.

The resulting code then makes decisions based on what variant of the enum was passed.

Fixes #379

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -32,6 +32,7 @@ use crate::node::{Address, Features, FetchResult, Seed, Seeds};
 use crate::prelude::*;
 use crate::service::message::{Announcement, AnnouncementMessage, Ping};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
+use crate::service::reactor::FetchDirection;
 use crate::storage;
 use crate::storage::{Inventory, ReadRepository, RefUpdate, WriteStorage};
 use crate::storage::{Namespaces, ReadStorage};
@@ -524,7 +525,7 @@ where
                 resp.send(untracked).ok();
             }
             Command::AnnounceRefs(id) => {
-                if let Err(err) = self.announce_refs(id, Namespaces::One(self.node_id())) {
+                if let Err(err) = self.announce_refs(id, &Namespaces::One(self.node_id())) {
                     error!("Error announcing refs: {}", err);
                 }
             }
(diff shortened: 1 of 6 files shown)


radicle-node/src/service.rs
```diff
@@ -542,9 +543,8 @@ where
     pub fn fetched(&mut self, fetch: Fetch, result: Result<Vec<RefUpdate>, FetchError>) {
         let remote = fetch.remote;
         let rid = fetch.rid;
-        let initiated = fetch.initiated;
 
-        if initiated {
+        if let Some(namespaces) = fetch.namespaces() {
```
cloudhead, line 547: I feel like this is a bit obscure, I would rather something like `if let FetchDirection::Initiated { namespaces } = fetch {`, otherwise it's not clear what's significant about the namespaces being optional.

radicle-node/src/service/reactor.rs
```diff
@@ -31,12 +31,42 @@ pub enum Io {
...
+    /// Indicates whether the fetch request was initiated or is a response.
+    pub direction: FetchDirection,
     /// Remote peer we are interacting with.
     pub remote: NodeId,
-    /// Indicates whether the fetch request was initiated by us.
-    pub initiated: bool,
+}
+
+impl Fetch {
+    pub fn initiated(&self) -> bool {
```
cloudhead, line 41:
```suggestion
    pub fn is_initiated(&self) -> bool {
```

radicle-node/src/service/reactor.rs
```diff
@@ -31,12 +31,42 @@ pub enum Io {
...
+        /// Namespaces to fetch.
+        namespaces: Namespaces,
+    },
+    /// Server is responding to a fetch request by uploading the
+    /// specified refspecs.
+    Upload,
+}
+
+impl FetchDirection {
+    pub fn initiated(&self) -> bool {
```
cloudhead, line 67:
```suggestion
    pub fn is_initiated(&self) -> bool {
```

radicle-node/src/service/reactor.rs
```diff
@@ -31,12 +31,42 @@ pub enum Io {
...
+    /// Client is fetching in order to receive the specified refspecs
+    /// in [`Namespaces`].
+    Receive {
+        /// Namespaces to fetch.
+        namespaces: Namespaces,
+    },
+    /// Server is responding to a fetch request by uploading the
+    /// specified refspecs.
+    Upload,
+}
```
cloudhead, line 64: I think we can do a bit better here in terms of the variant names; especially I think `Receive` doesn't communicate enough (it might sound like you are receiving the fetch request).

Some proposals:
`Initiator / Responder` (this terminology is from Noise and I quite like it)
`Download / Upload`
`Fetch / Upload` (git-influenced)

radicle-node/src/service.rs
FintanH: Ah, that's a fair point. I actually had that originally but then opted for introducing this method. Just as another suggestion, what do you think of renaming the method:
```rust
pub fn receiving(&self) -> Option<&Namespaces> {
/* etc. */
}
```

Otherwise, happy to change it to the `if let` :+1:

radicle-node/src/service/reactor.rs
FintanH: Right :sweat_smile:

radicle-node/src/service.rs
cloudhead: Yeah, see my other comments about "receiving"; but I think something like `if let Some(ns) = fetch.initiated() {` could work.

radicle-node/src/service/reactor.rs
cloudhead: On this topic, I'd also try to have the `is_` functions and the one returning the namespaces match the naming of the variants better.

cloudhead: Ah gotcha.

cloudhead:
Can this be merged before the other fetch PRs? Or the other way around?

radicle-node/src/service/reactor.rs
```diff
@@ -31,12 +31,42 @@ pub enum Io {
...
+        /// Namespaces to fetch.
+        namespaces: Namespaces,
+    },
+    /// Server is responding to a fetch request by uploading the
+    /// specified `refspecs` sent by the client.
+    Responder,
+}
+
+impl FetchDirection {
+    pub fn is_initiated(&self) -> bool {
```
cloudhead, line 67: Nit: maybe `is_initiator` when it's on `FetchDirection`?

cloudhead (approved):
Looks good!

radicle-node/src/service/reactor.rs
FintanH: And why not both then? :thinking:

cloudhead: Yeah actually both would also work for me :D 

It's just that `Fetch::is_initiated` is easier to swallow because it's about the whole fetch object, which isn't an enum. But for the enum, it's nice to have it match the variant.

FintanH: Ya, I think I'll switch to `is_initiator` for both then :)

FintanH:
Good to merge :+1:

PR #398, opened 2023-03-07
Node replication fixes

cloudhead:
*See commits*


FintanH (approved):
Nice!

I actually got caught out by this session transition when messing around with the fetch direction code. I thought it was only for the initiator side, but then realised it was for both.

cloudhead:
Hehe yea, took me a few hours to debug what was going on, p2p protocols are fascinating!

PR #399, opened 2023-03-07
httpd: Run tests off another remote than seed

xphoniex:
As you can see in test cases, this doesn't change other stuff much. Maybe I should manually remove the one hosted by seed on-disk?

### DIFF
--- a/radicle-httpd/Cargo.toml
+++ b/radicle-httpd/Cargo.toml
@@ -47,6 +47,6 @@ path = "../radicle-term"
 hyper = { version = "0.14.17", default-features = false, features = ["client"] }
 pretty_assertions = { version = "1.3.0" }
 radicle-cli = { path = "../radicle-cli" }
-radicle-crypto = { path = "../radicle-crypto" }
+radicle-crypto = { path = "../radicle-crypto", features = ["test"] }
 tempfile = { version = "3.3.0" }
 tower = { version = "0.4", features = ["util"] }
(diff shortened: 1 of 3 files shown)


xphoniex:
I went with I originally suggested, `rad auth` with seed, and then used `MockSigner` for any operation afterwards, which seems to have worked.

However these tests are failing:
```
    api::v1::projects::routes::test_projects_create_patches
    api::v1::projects::routes::test_projects_issues_comment
    api::v1::projects::routes::test_projects_issues_create
    api::v1::projects::routes::test_projects_issues_reply
```

Since they are trying to create a `cob` for a fork on seed, and seed doesn't have its own fork. Calling `create` works and returns `success: true` but trying to view the item fails, because the uri is `/projects/:project/issues/:id`.

Either, we'd have to add routes `/projects/:project/remote/:remote/issues/:id`

Or, for such tests, we prepare a separate environment where everything is in seed instead and there's no `MockSigner`.

cloudhead:
Ok, perfect. So now we have to fix those routes :)

xphoniex:
> How about having two ways to setup tests: `tests::setup::contributor` (sets up our own fork with a signer), and `tests::setup::seed`, sets up a read-only seed with many forks.

Yup, https://github.com/radicle-dev/heartwood/pull/399/commits/2f440a224427c9daa1656eb9f10111fa5bc525c7. what do you think?

cloudhead (approved):

Issue #342, opened 2023-02-20
Queue worker fetch requests when necessary


cloudhead:
Currently if a node is being fetched from, and we try to fetch from it again, we log an error. We should ideally handle that more gracefully by queuing the request and retrying, or by finding another node.

keepsimple1:
Is there anyone working on this? If not, I will take a look.

keepsimple1:
> We should ideally handle that more gracefully by queuing the request and retrying, or by finding another node.

Can we use [`Reactor.outbox`](https://github.com/radicle-dev/heartwood/blob/05d2487a47afba3bf2ab3aabd13d74ddf2f8bcd8/radicle-node/src/service/reactor.rs#L47-L50) to queue the request? Is there a gotcha for that?

cloudhead:
Yeah, that's what I'm thinking. It wasn't made for this use-case, but it may just work!

cloudhead:
Closed by #401

PR #401, opened 2023-03-07
radicle-node: Queue fetch requests when necessary

keepsimple1:
This is to solve issue #342 .

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -568,12 +568,14 @@ where
                 if other == rid {
                     debug!(target: "service", "Ignoring redundant attempt to fetch {rid} from {from}");
                 } else {
-                    // TODO: If we can't fetch, it's because we're already fetching from
-                    // this peer. So we need to queue the request, or find another peer.
-                    error!(
+                    // If we can't fetch, it's because we're already fetching from
+                    // this peer. So we need to queue the request.
+                    // TODO: consider to find another peer.
+                    debug!(
                         target: "service",
-                        "Dropping fetch for {rid} from {from}: another fetch is ongoing"
+                        "Queueing fetch for {rid} from {from}: another fetch is ongoing"
                     );
+                    session.queue_fetch(rid);
                 }
             }
             session::FetchResult::NotConnected => {
@@ -651,6 +653,7 @@ where
                 // to the gossip protocol, otherwise the messages will
                 // be queued.
                 self.sync_and_announce();
+                self.process_fetch_queue(&remote);
             }
             FetchDirection::Responder => self.switch_to_gossip(remote),
         }
@@ -1273,6 +1276,16 @@ where
             }
             Err(e) => {
                 error!(target: "service", "Failed to sync inventory: {e}");
+            }
+        }
+    }
+
+    /// Execute the next pending fetch with `remote`, if any.
+    fn process_fetch_queue(&mut self, remote: &NodeId) {
+        if let Some(session) = self.sessions.get_mut(remote) {
+            if let Some(rid) = session.dequeue_fetch() {
+                debug!(target: "service", "Dequeued a pending fetch {rid} with {remote}");
+                self.fetch(rid, remote);
             }
         }
     }
--- a/radicle-node/src/service/session.rs
+++ b/radicle-node/src/service/session.rs
@@ -1,3 +1,4 @@
+use std::collections::VecDeque;
 use std::fmt;
 
 use radicle::storage::Namespaces;
@@ -154,6 +155,9 @@ pub struct Session {
     /// Last time a message was received from the peer.
     pub last_active: LocalTime,
 
+    /// Fetches queued due to another ongoing fetch.
+    pending_fetches: VecDeque<Id>,
+
     /// Connection attempts. For persistent peers, Tracks
     /// how many times we've attempted to connect. We reset this to zero
     /// upon successful connection.
@@ -192,6 +196,7 @@ impl Session {
             persistent,
             last_active: LocalTime::default(),
             attempts: 1,
+            pending_fetches: VecDeque::new(),
             rng,
         }
     }
@@ -209,6 +214,7 @@ impl Session {
             persistent,
             last_active: LocalTime::default(),
             attempts: 0,
+            pending_fetches: VecDeque::new(),
             rng,
         }
     }
@@ -362,5 +368,13 @@ impl Session {
             reactor.write(self, Message::Ping(msg));
         }
         Ok(())
+    }
+
+    pub(crate) fn queue_fetch(&mut self, rid: Id) {
+        self.pending_fetches.push_back(rid);
+    }
+
+    pub(crate) fn dequeue_fetch(&mut self) -> Option<Id> {
+        self.pending_fetches.pop_front()
     }
 }
--- a/radicle-node/src/tests.rs
+++ b/radicle-node/src/tests.rs
@@ -1262,3 +1262,89 @@ fn prop_inventory_exchange_dense() {
         .gen(qcheck::Gen::new(8))
         .quickcheck(property as fn(MockStorage, MockStorage, MockStorage));
 }
+
+#[test]
+fn test_queued_fetch() {
+    let storage = arbitrary::nonempty_storage(3);
+    let mut repo_keys = storage.inventory.keys();
+    let rid = *repo_keys.next().unwrap();
+    let rid2 = *repo_keys.next().unwrap();
+    let rid3 = *repo_keys.next().unwrap();
+    let mut alice = Peer::with_storage("alice", [7, 7, 7, 7], storage);
+    let bob = Peer::new("bob", [8, 8, 8, 8]);
+    let (send, _recv) = chan::bounded::<node::FetchResult>(1);
+
+    logger::init(log::Level::Debug);
+
+    // Send the first fetch.
+    alice.connect_to(&bob);
+    alice.command(Command::Fetch(rid, bob.id, send));
+
+    assert_matches!(alice.messages(bob.id).next(), Some(Message::Fetch { .. }));
+
+    // Send the 2nd fetch that will be queued.
+    let (send2, _recv2) = chan::bounded::<node::FetchResult>(1);
+    alice.command(Command::Fetch(rid2, bob.id, send2));
+
+    // Send the 3rd fetch that will be queued.
+    let (send3, _recv3) = chan::bounded::<node::FetchResult>(1);
+    alice.command(Command::Fetch(rid3, bob.id, send3));
+
+    // We shouldn't send out the 2nd, 3rd fetch while we're doing the 1st fetch.
+    assert_matches!(alice.messages(bob.id).next(), None);
+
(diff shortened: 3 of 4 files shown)


radicle-node/src/tests.rs
```diff
@@ -593,6 +593,52 @@ fn test_refs_announcement_no_subscribe() {
...
+            initiated: true,
+        },
+        Ok(vec![]),
+    );
+
+    // Now the 1st fetch is done, the messages Bob missed should be relayed to him.
+    let mut messages = alice.messages(bob.id);
+    assert_matches!(messages.next(), Some(Message::Fetch { .. }));
+    assert_matches!(messages.next(), Some(Message::Ping { .. }));
+}
```
cloudhead, line 640: Such a beautiful test :smiling_face_with_tear: 

I think we're missing an important detail though! We are checking that alice is indeed sending the queued `Fetch` message to Bob, but not that she's actually ready to fetch from Bob the second time (or the first time, but that's tested elsewhere I guess).

What we should do at the end of this test is additionally:
1. Have alice receive the `FetchOk` from bob
2. Expect that alice outputs an `Io::Fetch` action (not a message!), which is how the service gears of for fetching. You can get these via `alice.outbox()`.

My guess is that this will cause test failure:
```rust
            (session::State::Connected { protocol, .. }, Message::FetchOk { rid }) => {
                if *protocol
                    != (session::Protocol::Gossip {
                        requested: Some(rid),
                    })
                {
                    [...]
                    return Err(session::Error::Misbehavior);
                }

```

keepsimple1: thanks for the details!

radicle-node/src/tests.rs
```diff
@@ -593,6 +593,58 @@ fn test_refs_announcement_no_subscribe() {
...
+            namespaces: Namespaces::All,
+            remote: bob.id,
+            initiated: true,
+        },
+        Ok(vec![]),
+    );
+
+    // Now the 1st fetch is done, the 2nd fetch message is drained.
+    let mut messages = alice.messages(bob.id);
+    assert_matches!(messages.next(), Some(Message::Fetch { .. }));
```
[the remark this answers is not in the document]
cloudhead: They aren't the same thing though, the comment may refer to a block of code. I think the assertion message on its own doesn't help much as you always have to look at the test anyway.

radicle-node/src/tests.rs
```diff
@@ -600,6 +600,64 @@ fn test_refs_announcement_no_subscribe() {
...
+
+    // `FetchOk` for the 2nd fetch.
+    alice.receive(bob.id(), Message::FetchOk { rid: rid2 });
+
+    // The 2nd fetch should be in `Io` now.
+    let out_item = alice.outbox().last().unwrap();
+    assert_matches!(out_item, Io::Fetch(..));
+    if let Io::Fetch(fetch) = out_item {
+        assert_eq!(fetch.rid, rid2);
+    }
```
cloudhead, line 658: You can combine these statements like
```rust
assert_matches!(a, Io::Fetch(fetch) if fetch.rid == rid2)
```

radicle/src/test/arbitrary.rs
```diff
@@ -60,9 +60,11 @@ pub fn vec<T: Eq + Arbitrary>(size: usize) -> Vec<T> {
 
 pub fn nonempty_storage(size: usize) -> MockStorage {
     let mut storage = gen::<MockStorage>(size);
-    storage
-        .inventory
-        .insert(gen::<Id>(size), gen::<Doc<Verified>>(size));
+    for _ in 0..size {
+        storage
+            .inventory
+            .insert(gen::<Id>(size), gen::<Doc<Verified>>(size));
```
cloudhead, line 66:
```suggestion
            .insert(gen::<Id>(1), gen::<Doc<Verified>>(1));
```
The size parameter doesn't have any effect on things with a fixed size, such as ids or docs, so we usually set it to `1`

radicle-node/src/service.rs
```diff
@@ -571,12 +571,9 @@ where
...
                 } else {
-                    // TODO: If we can't fetch, it's because we're already fetching from
-                    // this peer. So we need to queue the request, or find another peer.
-                    error!(
-                        target: "service",
-                        "Dropping fetch for {rid} from {from}: another fetch is ongoing"
-                    );
+                    // This fetch will be queued in `Reactor.outbox`.
+                    let fetch = Message::Fetch { rid };
+                    self.reactor.write(session, fetch);
```
cloudhead, line 576: What if 2+ fetch messages are queued? When the initial fetch completes, both fetch messages will be drained from the outbox, and sent to the peer. The second fetch message will be dropped, as it's received during a fetch, and the peer will be disconnected for misbehavior :thinking:

radicle-node/src/tests.rs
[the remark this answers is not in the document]
slack-coder: Not always, though they can be leaned on to elaborate on the error and make the code clearer reducing the need for comments.

```
// We shouldn't send out the 2nd fetch to Bob while we're doing the 1st fetch.
assert_matches!(alice.messages(bob.id).next(), None);
```

vs.

```
assert_matches!(alice.messages(bob.id).next(), None, "only one fetch at a time");
```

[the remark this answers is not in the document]
cloudhead: Yeah I mean, I prefer the comment since it reads a lot better in your example.

radicle-node/src/service.rs
keepsimple1: Only one fetch message will be drained at a time, see [here](https://github.com/radicle-dev/heartwood/blob/keepsimple1/queue-fetch-342/radicle-node/src/service.rs#L1268). 

The remaining fetch messages will still be queued and drained the next time around. I am adding some test code to cover such cases.

radicle-node/src/tests.rs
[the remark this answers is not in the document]
keepsimple1: I've updated the diff to address the original concern. Please review.

cloudhead (commented):
The solution seems a little messy to me. We're switching to gossip, and writing gossip messages to the peer, but then sometimes we're actually not switching to gossip.. So which is it?

cloudhead:
I'm not sure the solution of using the outbox is the right one. The outbox was designed to buffer gossip messages during a fetch, but here it's also being used to buffer fetches, which seems to add a lot of complexity.

I think it would be worth exploring a different solution, where `Fetch` messages do not end up in the outbox at all, but are queued in their own queue that is processed separately from the outbox.

cloudhead:
Note also that fetches can run in parallel as long as they are over different connections.

keepsimple1:
> Note also that fetches can run in parallel as long as they are over different connections.

Do you mean different connections to the same peer,  or different connections to different peers?

keepsimple1:
> I think it would be worth exploring a different solution, where `Fetch` messages do not end up in the outbox at all, but are queued in their own queue that is processed separately from the outbox.

I will explore some other solution.

keepsimple1:
> Perhaps storing a `queue` object in `Session` would be an idea..

I tried this idea. It seems to be working.  One thing is that we do not support to send a pending fetch to a different peer yet. I think there is both pros and cons for doing that. Probably for another PR.

radicle-node/src/service/session.rs
```diff
@@ -154,6 +154,9 @@ pub struct Session {
     /// Last time a message was received from the peer.
     pub last_active: LocalTime,
 
+    /// Fetches queued due to another ongoing fetch.
+    pending_fetches: Vec<Id>,
```
keepsimple1, line 158: We can also make this one `pub(crate)` if needed. For this patch, I chose to define methods to access this field.

radicle-node/src/service.rs
```diff
@@ -568,12 +568,14 @@ where
...
-                    // TODO: If we can't fetch, it's because we're already fetching from
-                    // this peer. So we need to queue the request, or find another peer.
-                    error!(
+                    // If we can't fetch, it's because we're already fetching from
+                    // this peer. So we need to queue the request.
+                    // TODO: consider to find another peer.
+                    debug!(
                         target: "service",
-                        "Dropping fetch for {rid} from {from}: another fetch is ongoing"
+                        "Queuing fetch for {rid} from {from}: another fetch is ongoing"
```
cloudhead, line 576:
```suggestion
                        "Queueing fetch for {rid} from {from}: another fetch is ongoing"
```
I think?

radicle-node/src/service/session.rs
```diff
@@ -363,4 +368,16 @@ impl Session {
...
+    pub(crate) fn add_pending_fetch(&mut self, rid: Id) {
+        self.pending_fetches.push(rid);
+    }
+
+    pub(crate) fn next_pending_fetch(&mut self) -> Option<Id> {
+        if self.pending_fetches.is_empty() {
+            None
+        } else {
+            Some(self.pending_fetches.remove(0))
+        }
```
cloudhead, line 381: It would make more sense to use a `VecDeque` with `pop_front()` and `push_back()`

radicle-node/src/service.rs
FintanH: ✅

radicle-node/src/service/session.rs
```diff
@@ -363,4 +368,16 @@ impl Session {
         }
         Ok(())
     }
+
+    pub(crate) fn add_pending_fetch(&mut self, rid: Id) {
+        self.pending_fetches.push(rid);
+    }
+
```
cloudhead, line 375: I'd be in favor of shorter names: `queue_fetch` and `dequeue_fetch`

radicle-node/src/service/session.rs
cloudhead: Best to keep it private imo

radicle-node/src/service.rs
```diff
@@ -1277,6 +1281,16 @@ where
         }
     }
 
+    /// Execute the next pending fetch with `remote`, if any.
+    fn next_pending_fetch(&mut self, remote: &NodeId) {
```
cloudhead, line 1285: `next_*` sounds like it's returning a fetch; how about `process_fetch_queue`?

radicle-node/src/service.rs
```diff
@@ -651,6 +653,8 @@ where
                 // to the gossip protocol, otherwise the messages will
                 // be queued.
                 self.sync_and_announce();
+
```
cloudhead, line 656:
```suggestion
```

cloudhead:
This solution seems a lot cleaner to me :+1:

radicle-node/src/service.rs
keepsimple1: Will update.   It's interesting what chatGPT said:

```
Both "queuing" and "queueing" are correct spellings of the word, with "queuing" being the more common spelling in British English and "queueing" being more common in American English. However, both spellings are widely accepted and can be used interchangeably.
```

cloudhead: Oh interesting!

cloudhead (approved):
LGTM

Issue #400, opened 2023-03-07
Improve `rad patch` target selection


slack-coder:
When creating a new patch via `rad open`, it needs to know the target repository (or peer) and branch.  The branch is hard coded as the project's default branch (e.g. master).  The target therefor is defined by the peer (or repository).

The current implementation only works when one target is available, otherwise it fails with error (resulting in issue #380).

I propose setting the target by default to the delegate's master branch.  Radicle currently supports the case of only one delegate anyways.  In the uncommon case the user wants to patch against another peer's branch, he/she can do so via an optional argument.

slack-coder:
There was confusion about Patches and merge Targets.  Instead of the patch requiring a (target repository, branch) pair, it requires a target git reference only.

Currently in Rad there is one sane git reference to use, the quorum of the delegates master branch.

cloudhead:
Closed by #407

PR #402, opened 2023-03-07
cli: add merge target option to `rad patch`

slack-coder:
Providing a merge target as a command line option supports using the command in scripts.

```
rad patch open --target <peer>
```

### DIFF
--- a/radicle-cli/src/commands/patch/common.rs
+++ b/radicle-cli/src/commands/patch/common.rs
@@ -49,6 +49,8 @@ pub struct MergeTargets {
 }
 
 /// Find potential merge targets for the given head.
+///
+/// Only delegates can be merge targets.
 pub fn find_merge_targets(
     head: &Oid,
     branch: &git::RefStr,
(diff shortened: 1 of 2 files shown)


radicle-cli/src/commands/patch/common.rs
```diff
@@ -77,6 +77,7 @@ pub fn get_merge_target(
     project: &Project,
     storage: &Repository,
     head_branch: &git::raw::Branch,
+    options: &Options,
```
FintanH, line 80: Since we're only using the `target` in `Options`, I'd suggest passing that in:
```suggestion
    target: Option<&NodeId>,
```

FintanH (commented):
Nice :) Have a couple of suggestions :+1:

radicle-cli/src/commands/patch/common.rs
```diff
@@ -85,6 +86,18 @@ pub fn get_merge_target(
     let mut spinner = term::spinner("Analyzing remotes...");
     let targets = find_merge_targets(&head_oid, project.default_branch().as_refstr(), storage)?;
 
+    if let Some(target) = options.target {
+        if targets.merged.into_iter().any(|remote| remote.id == target) {
+            anyhow::bail!("'{target}' already has already included the commits");
```
cloudhead, line 91:
```suggestion
            anyhow::bail!("'{target}' already has the proposed changes");
```

slack-coder:
After reviewing the code, there's deeper changes which need to be made.  This PR is therefor being closed.

Related to https://github.com/radicle-dev/heartwood/issues/400

PR #404, opened 2023-03-08
cli: Update `track` command

### LINKED ISSUE #312: `rad-track`: Add ability to track a repo without having a clone/checkout
Currently `rad-track` only works from inside a radicle project repo, and only adds peer trackings. We should re-think this command a little bit so that it can modify the tracking policy for both peers and repos, and without having the repo checked out.

Perhaps something like:
```
rad track --nid <node-id> # works when inside a repo, uses repo as RID
rad track --rid <repo-id> # works anywhere, tracks repo
rack track --rid <repo-id> --nid <node-id> # works anywhere, tracks specific peer for given repo
```

I also think we should start moving to `--rid` and `--nid` argument names consistently in commands.

cloudhead:
Allows the `track` command to be used to track nodes and repos.

Closes #312

### DIFF
--- a/radicle-cli/src/commands/track.rs
+++ b/radicle-cli/src/commands/track.rs
@@ -1,36 +1,51 @@
 use std::ffi::OsString;
-use std::str::FromStr;
 
-use anyhow::{anyhow, Context as _};
+use anyhow::anyhow;
 
+use radicle::node::tracking::{Alias, Scope};
 use radicle::node::{Handle, NodeId};
-use radicle::storage::ReadStorage;
+use radicle::{prelude::*, Node};
 
 use crate::terminal as term;
 use crate::terminal::args::{Args, Error, Help};
 
 pub const HELP: Help = Help {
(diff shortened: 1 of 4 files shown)


radicle-cli/src/commands/track.rs
```diff
@@ -1,36 +1,51 @@
...
     name: "track",
-    description: "Manage project tracking policy",
+    description: "Manage repository and node tracking policy",
     version: env!("CARGO_PKG_VERSION"),
     usage: r#"
 Usage
 
-    rad track <peer> [--fetch] [--alias <name>]
+    rad track <did> [--fetch] [--alias <name>]
+    rad track <rid> [--fetch] [--scope <scope>]
```
FintanH, line 20:
```suggestion
    rad track <did> [--[no-]fetch] [--alias <name>]
    rad track <rid> [--[no-]fetch] [--scope <scope>]
```

PR #405, opened 2023-03-08
node: Handle `FetchOk` never being received

### LINKED ISSUE #382: `rad fetch` sometimes freezes
Running `rad fetch` can result in the fetch command 'fetching' forever (not exiting).

The context I've run into it is running it twice after the peer creates a patch.  This is likely due to an error when fetching the patch causing the remote peer to disconnect.

```
2023-03-03T10:29:10.140-05:00 DEBUG wire     Dropping read buffer with `[111, 100, 110, 101]`
2023-03-03T10:29:10.140-05:00 DEBUG wire     Disconnecting peer (fd=14): error: peer misbehaved
2023-03-03T10:29:10.140-05:00 DEBUG wire     Received transport handover (fd=14)
2023-03-03T10:29:10.140-05:00 DEBUG service  Disconnected from z6MkkTUSrCLYz9o8HQD7PJnC68eQrZ59fPhcHmUDsNjdd7sv (error: peer misbehaved)
```

cloudhead:
It's possible for a peer to disconnect before responding with `FetchOk`. In that case, we make sure to return to any pending fetch request with an error. Closes #382

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -710,7 +710,16 @@ where
         debug!(target: "service", "Disconnected from {} ({})", remote, reason);
 
         if let Some(session) = self.sessions.get_mut(&remote) {
-            session.to_disconnected(since);
+            // If the peer disconnected while we were waiting for a [`Message::FetchOk`],
+            // return a failure to any potential fetcher.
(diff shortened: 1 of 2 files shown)


FintanH (approved):
Makes sense :+1:

PR #407, opened 2023-03-09
cli: use canonical head for patch target

slack-coder:
Correct assumptions made about how patching work in `rad patch`.  Make the default Patch target a `git ref` in place of a target 'peer'.

Use the canonical head as the git reference.  This is head of the project's master branch which project delegates have concensus on.

The process is simplified by ignoring the case of other delegates having merged the changes.  This would be an rare circumstance which can be handled by the maintainers/contributor for the rare circumstance it happens in.

### DIFF
--- a/radicle-cli/examples/rad-patch.md
+++ b/radicle-cli/examples/rad-patch.md
@@ -29,7 +29,7 @@ $ rad patch open --message "Define power requirements" --message "See details."
 ✓ Pushing HEAD to storage...
 ✓ Analyzing remotes...
 
-z6MknSL…StBU8Vi/master (f2de534) <- z6MknSL…StBU8Vi/flux-capacitor-power (3e674d1)
+master (f2de534) <- z6MknSL…StBU8Vi/flux-capacitor-power (3e674d1)
 
 1 commit(s) ahead, 0 commit(s) behind
 
--- a/radicle-cli/examples/rad-patch.md
+++ b/radicle-cli/examples/rad-patch.md
@@ -29,7 +29,7 @@ $ rad patch open --message "Define power requirements" --message "See details."
 ✓ Pushing HEAD to storage...
 ✓ Analyzing remotes...
 
-master (f2de534) <- z6MknSL…StBU8Vi/flux-capacitor-power (3e674d1)
+master <- z6MknSL…StBU8Vi/flux-capacitor-power (3e674d1)
 
 1 commit(s) ahead, 0 commit(s) behind
 
--- a/radicle-cli/src/commands/patch/common.rs
+++ b/radicle-cli/src/commands/patch/common.rs
@@ -6,9 +6,7 @@ use radicle::cob::patch::{Clock, MergeTarget, Patch, PatchId, Patches};
 use radicle::git;
 use radicle::git::raw::Oid;
 use radicle::prelude::*;
-use radicle::storage;
 use radicle::storage::git::Repository;
-use radicle::storage::Remote;
 
 use crate::terminal as term;
 use crate::terminal::args::Error;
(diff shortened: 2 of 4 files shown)


radicle-cli/src/commands/patch/common.rs
```diff
@@ -39,68 +37,26 @@ pub fn confirm<D: fmt::Display>(prompt: D, options: &Options) -> bool {
...
-        _ => {
-            // TODO: Let user select which branch to use as a target.
-            todo!();
-        }
-    };
     // TODO: Tell user how many peers don't have this change.
     spinner.finish();
 
-    Ok((target_peer.clone(), *target_oid))
+    let (_, _, branch, _) = qualified_ref.non_empty_components();
```
slack-coder, line 58: Is this the correct way to get a branch name from a git:Qualified type?

radicle-cli/src/commands/patch/common.rs
```diff
@@ -39,68 +37,26 @@ pub fn confirm<D: fmt::Display>(prompt: D, options: &Options) -> bool {
...
-    let targets = find_merge_targets(&head_oid, project.default_branch().as_refstr(), storage)?;
+) -> anyhow::Result<(String, git::Oid)> {
+    let spinner = term::spinner("Analyzing remotes...");
+    let (qualified_ref, target_oid) = storage.canonical_head()?;
+    let head_oid = branch_oid(head_branch)?;
+    let merge_base = storage.raw().merge_base(*head_oid, *target_oid)?;
+
+    if head_oid == merge_base.into() {
+        anyhow::bail!("changes have already been merged");
+    }
```
slack-coder, line 53: This feels like it could belong in its own section for patch validation

radicle-cli/src/commands/patch/common.rs
slack-coder: Thanks @FintanH !

radicle-cli/src/commands/patch/common.rs
```diff
@@ -33,68 +31,35 @@ pub fn branch_oid(branch: &git::raw::Branch) -> anyhow::Result<git::Oid> {
...
-        }
+#[inline]
+fn get_branch(git_ref: git::Qualified) -> String {
+    let (_, _, head, tail) = git_ref.non_empty_components();
+    if tail.clone().count() == 0 {
+        return head.to_string();
     }
-    Ok(targets)
+    head.join(tail.collect::<git::RefString>()).to_string()
 }
```
cloudhead, line 41: This is kind of crazy, there must be a simpler way @FintanH ? Maybe via `RefStr::strip_prefix` and `lit`?

radicle-cli/src/commands/patch/common.rs
cloudhead: Yeah this makes sense here. I'd change the error though to something like `patch commits are already part of the canonical branch; nothing to do`

radicle-cli/src/commands/patch/create.rs
```diff
@@ -60,9 +58,8 @@ fn show_patch_commit_info(
 
     term::blank();
     term::info!(
-        "{}/{} ({}) <- {}/{} ({})",
-        term::format::dim(term::format::node(&target_peer.id)),
-        term::format::highlight(project.default_branch().to_string()),
+        "{} ({}) <- {}/{} ({})",
+        term::format::highlight(target_ref),
         term::format::secondary(term::format::oid(*target_oid)),
```
cloudhead, line 63: Maybe we shouldn't show the target_oid either, seems kind of random, wdyt?

radicle-cli/src/commands/patch/common.rs
slack-coder: The tail.collect() was bugging out when tail was empty

radicle-cli/src/commands/patch/create.rs
slack-coder: Sounds good if there is an easy way to inspect the ref of the canonical branch.  The canonical head may be different to one or more of the delegate's head.  Without an easy way to inspect it, it would be difficult to know what is exactly being patched against.

cloudhead: True but the target_oid is not patched against anyway, the only relevant info would be the merge base if you wanted to include that.

radicle-cli/src/commands/patch/common.rs
```diff
@@ -33,68 +31,35 @@ pub fn branch_oid(branch: &git::raw::Branch) -> anyhow::Result<git::Oid> {
...
-            targets.not_merged.push((remote, target_oid));
-        }
+#[inline]
+fn get_branch(git_ref: git::Qualified) -> git::RefString {
+    let (_, _, head, tail) = git_ref.non_empty_components();
+    if tail.clone().count() == 0 {
+        return head.to_ref_string();
     }
-    Ok(targets)
+    head.join(tail.collect::<git::RefString>())
```
FintanH, line 40: Here's what I did in surf :)

```rust
pub(crate) fn refstr_join<'a>(c: Component<'a>, cs: Components<'a>) -> RefString {
    std::iter::once(c).chain(cs).collect::<RefString>()
}
```

FintanH, line 40:
```suggestion
    std::iter::once(head).chain(tail).collect()
```

radicle-cli/src/commands/patch/create.rs
slack-coder: The patch is applied against the head of the master branch.  There could be merge conflicts for instance.  Integration tests should be run against the merged result of an update too.  This is where the project's canonical head commit is useful.

This is likely best supported by other parts of Radicle.

slack-coder: https://github.com/radicle-dev/heartwood/pull/407/commits

cloudhead: The canonical head isn't immutable though, which means that as time passes, "target_oid" can move, so it's not really a stable property of the patch, it's just the current head of the branch you're patching against.. Not sure that's super useful :shrug:

slack-coder: Maybe we should have the head easily inspectable or updated as its own git remote?

I've updated the code and removed references to both the patch creation and update assuming we were minimizing the content.  If not, poke me back, and I'll undo the change the for 'update'.

radicle-cli/examples/rad-patch.md
```diff
@@ -29,7 +29,7 @@ $ rad patch open --message "Define power requirements" --message "See details."
 ✓ Pushing HEAD to storage...
 ✓ Analyzing remotes...
 
-z6MknSL…StBU8Vi/master (f2de534) <- z6MknSL…StBU8Vi/flux-capacitor-power (3e674d1)
```
cloudhead, line 32: The ref on the right is important, it's the head of the branch you want to merge.

cloudhead (approved):

PR #409, opened 2023-03-09
httpd: Serve root handler on `/api/v1` instead of `/api`

xphoniex:
```console
$ curl localhost:8080/api | jq
{
  "path": "/api",
  "links": [
    "/v1"
  ]
}

$ curl localhost:8080/api/v1 | jq
{
  "message": "Welcome!",
  "service": "radicle-httpd",
  "version": "0.1.0-b2c2b65",
  "node": {
    "id": "z6MkqgfPWLDruEjqzY5enrrqSSSxpARdFsQK6DWJfBBV9AV4"
  },
  "path": "/api/v1",
  "links": [
    {
      "href": "/projects",
      "rel": "projects",
      "type": "GET"
    },
    {
      "href": "/node",
      "rel": "node",
      "type": "GET"
    },
    {
      "href": "/delegates/:did/projects",
      "rel": "projects",
      "type": "GET"
    },
    {
      "href": "/stats",
      "rel": "stats",
      "type": "GET"
    }
  ]
}
```

### DIFF
--- a/radicle-httpd/src/api.rs
+++ b/radicle-httpd/src/api.rs
@@ -4,7 +4,6 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
 
-use axum::extract::State;
 use axum::http::header::{AUTHORIZATION, CONTENT_TYPE};
 use axum::http::Method;
 use axum::response::{IntoResponse, Json};
@@ -76,12 +75,8 @@ impl Context {
 }
 
 pub fn router(ctx: Context) -> Router {
-    let root_router = Router::new()
-        .route("/", get(root_handler))
-        .with_state(ctx.clone());
-
     Router::new()
-        .merge(root_router)
+        .route("/", get(root_handler))
         .merge(v1::router(ctx))
         .layer(
             CorsLayer::new()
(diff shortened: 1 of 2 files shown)


rudolfs:
Just looking at the outputs you posted. Shouldn't _everything_ be under `/api/v1` including `message`, `service`, etc?

xphoniex:
> Just looking at the outputs you posted. Shouldn't _everything_ be under `/api/v1` including `{"message":"Welcome!","service":"radicle-httpd","version":"0.1.0-b2c2b65","node":{"id":"z6MkwPUeUS2fJMfc2HZN1RQTQcTTuhw4HhPySB8JeUg2mVvx"},"path":"/api","links":[{"canonical":"/v1"}]}`?
> 
> If we still have stuff under plain `/api` then that doesn't fix anything in the client lib for me.

Thanks for clarifying this, I wanted to ask about this. Should I revert @cloudhead ?

cloudhead:
Yeah, though both have merit, I would say that `/api` is not in itself an API and therefore should only provide the minimum links to the actual API, which is `/api/v1`.

rudolfs:
The links seem to be outdated or incomplete btw. I guess we write them manually?

xphoniex:
> The links seem to be outdated or incomplete btw. I guess we write them manually? 

Which other part is outdated/incomplete? `/sessions` is missing.

> Also, shouldn't the links be shown at the very top-level, i.e. `/`? People would have to guess that it's `/api` in the first place.

api simply starts from `/api`, and we shouldn't expect hitting `"/"` to give us api-related info, as that's not clearly api, and falls into git router area.

rudolfs:
> > The links seem to be outdated or incomplete btw. I guess we write them manually?
> 
> Which other part is outdated/incomplete? `/sessions` is missing.

I guess the rest of the project routes:
```
/projects/:project
/projects/:project/commits
/projects/:project/commits/:sha
/projects/:project/activity
/projects/:project/tree/:sha
/projects/:project/tree/:sha/*path
/projects/:project/remotes
/projects/:project/remotes/:peer
/projects/:project/blob/:sha/*path
/projects/:project/readme/:sha
/projects/:project/issues
/projects/:project/issues/:id
/projects/:project/patches
/projects/:project/patches/:id
```

> > Also, shouldn't the links be shown at the very top-level, i.e. `/`? People would have to guess that it's `/api` in the first place.
> 
> api simply starts from `/api`, and we shouldn't expect hitting `"/"` to give us api-related info, as that's not clearly api, and falls into git router area.

ah, forgot that we serve git from it. 👍

xphoniex:
I thought these were supposed to give high level view of routes, do we want that granularity @cloudhead ?

xphoniex:
> Do we need the routes showing up there?

Probably not, unless interface needs it? @rudolfs 

This was copied from `http-api`.

radicle-httpd/src/api.rs
```diff
@@ -98,35 +93,10 @@ pub fn router(ctx: Context) -> Router {
...
-                "type": "GET"
-            },
-            {
-                "href": "/v1/stats",
-                "rel": "stats",
-                "type": "GET"
-            }
-        ]
+        "path": "/api",
+        "links": ["/v1"]
```
cloudhead, line 99: Maybe use the same format as the in /v1? Eg.  with href/rel/type?

rudolfs:
> 

I don't need the routes.

PR #412, opened 2023-03-10
node: Add eventing system to handler

xphoniex:
Context: https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/Emitting.20events.20from.20.60node.60

### DIFF
--- a/radicle-node/src/runtime.rs
+++ b/radicle-node/src/runtime.rs
@@ -3,6 +3,7 @@ mod handle;
 use std::io::{BufRead, BufReader};
 use std::os::unix::net::UnixListener;
 use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
 use std::{fs, io, net, thread, time};
 
 use crossbeam_channel as chan;
@@ -22,7 +23,7 @@ use crate::address;
 use crate::control;
 use crate::crypto::Signer;
 use crate::node::{routing, NodeId};
-use crate::service::tracking;
+use crate::service::{tracking, Event};
 use crate::wire;
 use crate::wire::Wire;
 use crate::worker;
@@ -59,6 +60,39 @@ pub enum Error {
     /// A git version error.
     #[error("git version error: {0}")]
     GitVersion(#[from] git::VersionError),
+}
+
+/// Publishes events to subscribers.
+#[derive(Debug, Clone)]
+pub struct Emitter<T> {
+    pub(crate) subscribers: Arc<Mutex<Vec<chan::Sender<T>>>>,
+}
+
+impl<T> Default for Emitter<T> {
+    fn default() -> Emitter<T> {
+        Emitter {
+            subscribers: Default::default(),
+        }
+    }
+}
+
+impl<T: Clone> Emitter<T> {
+    /// Emit event to subscribers and drop those who can't receive it.
+    pub(crate) fn emit(&self, event: T) {
+        self.subscribers
+            .lock()
+            .unwrap()
+            .retain(|s| s.try_send(event.clone()).is_ok());
+    }
+
+    /// Subscribe to events stream.
+    pub fn events(&mut self) -> chan::Receiver<T> {
+        let (sender, receiver) = chan::unbounded();
+        let mut subs = self.subscribers.lock().unwrap();
+        subs.push(sender);
+
+        receiver
+    }
 }
 
 /// Holds join handles to the client threads, as well as a client handle.
@@ -113,6 +147,7 @@ impl<G: Signer + Ecdh + 'static> Runtime<G> {
 
         log::info!(target: "node", "Default tracking policy set to '{}'", &config.policy);
         log::info!(target: "node", "Initializing service ({:?})..", network);
+        let emitter: Emitter<Event> = Default::default();
         let service = service::Service::new(
             config,
             clock,
@@ -122,6 +157,7 @@ impl<G: Signer + Ecdh + 'static> Runtime<G> {
             tracking,
             signer.clone(),
             rng,
+            emitter.clone(),
         );
 
         let (worker_send, worker_recv) = chan::unbounded::<worker::Task<G>>();
@@ -138,7 +174,7 @@ impl<G: Signer + Ecdh + 'static> Runtime<G> {
             log::info!(target: "node", "Listening on {local_addr}..");
         }
         let reactor = Reactor::named(wire, popol::Poller::new(), id.to_human())?;
-        let handle = Handle::new(home.clone(), reactor.controller());
+        let handle = Handle::new(home.clone(), reactor.controller(), emitter);
         let atomic = git::version()? >= git::VERSION_REQUIRED;
 
         if !atomic {
--- a/radicle-node/src/runtime/handle.rs
+++ b/radicle-node/src/runtime/handle.rs
@@ -13,8 +13,10 @@ use crate::crypto::Signer;
 use crate::identity::Id;
 use crate::node::{Command, FetchResult};
 use crate::profile::Home;
+use crate::runtime::Emitter;
 use crate::service;
 use crate::service::tracking;
+use crate::service::Event;
 use crate::service::{CommandError, QueryState};
 use crate::service::{NodeId, Sessions};
 use crate::wire;
@@ -64,6 +66,20 @@ pub struct Handle<G: Signer + Ecdh> {
 
     /// Whether a shutdown was initiated or not. Prevents attempting to shutdown twice.
     shutdown: Arc<AtomicBool>,
+
+    /// Publishes events to subscribers.
+    emitter: Emitter<Event>,
+}
+
+impl<G: Signer + Ecdh> Handle<G> {
+    /// Subscribe to events stream.
+    pub fn events(&mut self) -> chan::Receiver<Event> {
+        let (sender, receiver) = chan::unbounded();
+        let mut subs = self.emitter.subscribers.lock().unwrap();
+        subs.push(sender);
+
+        receiver
+    }
 }
 
 impl<G: Signer + Ecdh> fmt::Debug for Handle<G> {
@@ -78,16 +94,22 @@ impl<G: Signer + Ecdh> Clone for Handle<G> {
             home: self.home.clone(),
             controller: self.controller.clone(),
             shutdown: self.shutdown.clone(),
+            emitter: self.emitter.clone(),
         }
     }
 }
 
 impl<G: Signer + Ecdh + 'static> Handle<G> {
-    pub fn new(home: Home, controller: reactor::Controller<wire::Control<G>>) -> Self {
+    pub fn new(
+        home: Home,
+        controller: reactor::Controller<wire::Control<G>>,
+        emitter: Emitter<Event>,
+    ) -> Self {
         Self {
             home,
             controller,
             shutdown: Arc::default(),
+            emitter,
         }
     }
 
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -29,6 +29,7 @@ use crate::node;
 use crate::node::routing;
 use crate::node::{Address, Features, FetchResult, Seed, Seeds};
 use crate::prelude::*;
+use crate::runtime::Emitter;
 use crate::service::message::{Announcement, AnnouncementMessage, Ping};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
 use crate::service::reactor::FetchDirection;
@@ -203,6 +204,8 @@ pub struct Service<R, A, S, G> {
     last_announce: LocalTime,
     /// Time when the service was initialized.
     start_time: LocalTime,
+    /// Publishes events to subscribers.
+    emitter: Emitter<Event>,
 }
 
 impl<R, A, S, G> Service<R, A, S, G>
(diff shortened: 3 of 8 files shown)


radicle-node/src/runtime/handle.rs
```diff
@@ -58,12 +58,65 @@ impl<T> From<chan::SendError<T>> for Error {
     }
 }
 
+/// Event enum to emit.
+#[derive(Clone)]
+pub enum Event {
```
xphoniex, line 63: Two notes:

- The [Event](https://github.com/radicle-dev/heartwood/blob/master/radicle-node/src/service.rs#L83) you showed here seems to be the type `node` sends to reactor which is an interface to network. We might not necessarily want the same type for local events.

- This can be a struct, but the downside is a lot of impl would change, etc. So I went with an `enum` here and implement `From` in order to get eq. of `.type`.

cloudhead: It sends it to reactor because the thought was that the reactor would then send it to the user via an `Io::Event`. There is an advantage in consolidating all of the effects of service under a single enum (`reactor::Io`).

radicle-node/src/runtime/handle.rs
```diff
@@ -58,12 +58,65 @@ impl<T> From<chan::SendError<T>> for Error {
...
+}
+
 pub struct Handle<G: Signer + Ecdh> {
     pub(crate) home: Home,
     pub(crate) controller: reactor::Controller<wire::Control<G>>,
 
     /// Whether a shutdown was initiated or not. Prevents attempting to shutdown twice.
     shutdown: Arc<AtomicBool>,
+    /// Map of event types, and subscribers interested in them.
+    subscribers: Arc<Mutex<HashMap<EventType, Vec<chan::Sender<Event>>>>>,
```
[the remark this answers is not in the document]
cloudhead: I don't think there's any need to have topics, since listeners can simply receive all events and choose which events to react to.

radicle-node/src/runtime/handle.rs
xphoniex: Oh, also, we can use `Event` for subscribing as well, but I thought it wouldn't pass the review. So I created two types, `Event` and `EventType`, but if we can look past that, we can merge them into one, and use empty `Event` as event_type for subscribe requests as well.

radicle-node/src/runtime/handle.rs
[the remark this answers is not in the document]
xphoniex: gotcha

radicle-node/src/runtime/handle.rs
cloudhead: Not sure I follow -- there shouldn't need to be a subscribe request event.

cloudhead:
Am I missing something or there's currently no way for the service to emit events that can be subscribed to?

xphoniex:
> A description in the commit and pull request would be helpful to explain the why and some of the how of the PR.
> 
> It would be helpful to others to keep in sync about what's happening even though there may have been discussion out of band.

Search function is your friend. https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/Emitting.20events.20from.20.60node.60

cloudhead:
> Search function is your friend. https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/Emitting.20events.20from.20.60node.60

That's not good enough though. It helps to have the context where the review is taking place.

radicle-node/src/service.rs
```diff
@@ -595,11 +600,18 @@ where
...
                             rid,
                             updated: updated.clone(),
-                        });
+                        };
+                        self.reactor.event(event.clone());
+                        self.emitter
+                            .subscribers
+                            .lock()
+                            .unwrap()
+                            .retain(|s| s.try_send(event.clone()).is_ok());
```
xphoniex, line 613: I didn't change the emit type yet, though.

radicle-node/src/runtime/handle.rs
```diff
@@ -64,6 +65,24 @@ pub struct Handle<G: Signer + Ecdh> {
...
     /// Whether a shutdown was initiated or not. Prevents attempting to shutdown twice.
     shutdown: Arc<AtomicBool>,
+
+    /// Publishes events to subscribers.
+    emitter: Emitter<Event>,
+}
+
+trait EventEmitter {
+    fn subscribe(&mut self) -> chan::Receiver<Event>;
+}
```
cloudhead, line 75: Can we avoid the trait?

xphoniex: it's there to suppress:
```rust
warning: associated function `subscribe` is never used
   --> radicle-node/src/runtime/handle.rs:100:8
    |
100 |     fn subscribe(&mut self) -> chan::Receiver<Event> {
    |        ^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

radicle-node/src/service.rs
cloudhead: Yeah this should be a method on `Emitter` called `publish`

radicle-node/src/service.rs
```diff
@@ -595,11 +600,18 @@ where
...
                         log::debug!(target: "service", "Fetched {rid} from {remote}");
 
-                        self.reactor.event(Event::RefsFetched {
+                        let event = Event::RefsFetched {
                             remote,
                             rid,
                             updated: updated.clone(),
-                        });
+                        };
+                        self.reactor.event(event.clone());
```
cloudhead, line 608:
```suggestion
```

xphoniex: what happens to the reactor then? are we removing it? or does reactor subscribe and then forwards the events?

radicle-node/src/service.rs
xphoniex: went with [`emit`](https://github.com/radicle-dev/heartwood/compare/151868ed3c0ec50054de690b24126c57aa5e247a..4a65b33f3ac878437c41bafce9fa3993e0e357ce) instead since it's what [nakamoto](https://github.com/cloudhead/nakamoto/blob/master/net/src/event.rs#L43) uses as well.

radicle-node/src/service.rs
xphoniex: removed it [here](https://github.com/radicle-dev/heartwood/compare/151868ed3c0ec50054de690b24126c57aa5e247a..4a65b33f3ac878437c41bafce9fa3993e0e357ce), waiting for the substitute.

xphoniex: I think [test](https://github.com/radicle-dev/heartwood/actions/runs/4416542559/jobs/7741336583) is failing because of this, need your input re this.

xphoniex: I tried a few things just to experiment, such as putting reactor inside emitter, but couldn't get it to compile. I guess it's harder than it looks, maybe we should abandon reactor `Vec` model, and use `Receiver` in another thread?

radicle-node/src/runtime/handle.rs
cloudhead: We can just move it to `Emitter` and make it public no?

radicle-node/src/service.rs
cloudhead: The reactor is not needed for events anymore, since you are sending them over another channel. We just need to read from that channel in the tests.

radicle-node/src/runtime/handle.rs
xphoniex: https://github.com/radicle-dev/heartwood/pull/412/commits/1fe751e262acbe8c687420cd4ff8357099494e9e

radicle-node/src/runtime.rs
```diff
@@ -61,6 +62,39 @@ pub enum Error {
...
+    }
+
+    /// Subscribe to events stream.
+    pub fn subscribe(&mut self) -> chan::Receiver<T> {
+        let (sender, receiver) = chan::unbounded();
+        let mut subs = self.subscribers.lock().unwrap();
+        subs.push(sender);
+
+        receiver
+    }
```
xphoniex, line 95: I added `subscribe` method on `Emitter` so that from `Service`, one can subscribe as well. (as well as from `Handler` which could come in handy for anything which interacts with `node`)

radicle-node/src/service.rs
```diff
@@ -340,6 +345,11 @@ where
         &self.signer
     }
 
+    /// Subscriber to inner `Emitter`.
+    pub fn subscribe(&mut self) -> chan::Receiver<Event> {
```
xphoniex, line 349: is `subscribe` name for fn descriptive enough?

radicle-node/src/test/peer.rs
```diff
@@ -319,10 +324,9 @@ where
         msgs.into_iter()
     }
 
-    /// Get a draining iterator over the peer's emitted events.
-    pub fn events(&mut self) -> impl Iterator<Item = Event> + '_ {
-        self.outbox()
-            .filter_map(|io| if let Io::Event(e) = io { Some(e) } else { None })
+    /// Get a stream of the peer's emitted events.
+    pub fn events(&mut self) -> chan::Receiver<Event> {
```
xphoniex, line 328: maybe I should rename this to `subscribe` too?

also what should I do with the `fn` after this, `pub fn outbox`? is that still relevant?

xphoniex:
also I skipped adding something like [`wait`](https://github.com/cloudhead/nakamoto/blob/master/net/src/event.rs#L97) as that seems more appropriate for a `Handle` usage vs. in a test.

radicle-node/src/runtime.rs
```diff
@@ -61,6 +55,39 @@ pub enum Error {
     GitVersion(#[from] git::VersionError),
 }
 
+/// Publishes events to subscribers.
+#[derive(Debug, Clone)]
+pub struct Emitter<T> {
```
xphoniex, line 60: also this doesn't have to be generic over `T`

radicle-node/src/address/types.rs
```diff
@@ -1,8 +1,8 @@
 use std::ops::{Deref, DerefMut};
 
 use nonempty::NonEmpty;
-use radicle::node;
-use radicle::node::Address;
+
+use radicle::node::{self, Address};
```
cloudhead, line 5: I try to avoid using `self` in imports.

cloudhead:
Can you revert all unnecessary import changes? Then ping me so I can review.

xphoniex:
> Can you revert all unnecessary import changes? Then ping me so I can review.

Reverted. I briefly discussed this with you, and I've explained why I made the re-formats in this other PR https://github.com/radicle-dev/heartwood/pull/441 . You can close that if you don't find it useful, but personally I find it much easier on the eyes.

radicle-node/src/address/types.rs
xphoniex: Agreed, but we've went with `self` elsewhere, so I just wanted to stick to the same formatting.

radicle-node/src/test/peer.rs
cloudhead: I'd actually rename the other `Handle::subscribe` function to `Handle::events`, because it's clearer that you're getting events.

`outbox` is still useful as it's the `Io` actions that are returned from there.

radicle-node/src/tests.rs
```diff
@@ -1098,8 +1100,8 @@ fn test_push_and_pull() {
         .unwrap()
         .is_some());
     assert_matches!(
-        sim.events(&bob.id).next(),
-        Some(service::Event::RefsFetched { remote, .. })
+        bob_events.recv().unwrap(),
```
cloudhead, line 1103:
```suggestion
        bob_events.recv(),
```

radicle-node/src/service/reactor.rs
```diff
@@ -81,11 +81,6 @@ pub struct Reactor {
 }
 
 impl Reactor {
-    /// Emit an event.
-    pub fn event(&mut self, event: Event) {
-        self.io.push_back(Io::Event(event));
-    }
-
```
cloudhead, line 88: You can also remove the `Io::Event` variant.

xphoniex: [removed](https://github.com/radicle-dev/heartwood/compare/096b533df2b3e727e32ed084c3f5e301bafd5a6d..350133ebabb47b1ed79128eb691c771235b1253f#diff-d8b3517be454cad4430e080cbba053e9c346237f8f5ea1e4e6142a9e70ec62c8L25-L26)

radicle-node/src/tests.rs
xphoniex: good catch, [fixed](https://github.com/radicle-dev/heartwood/compare/096b533df2b3e727e32ed084c3f5e301bafd5a6d..350133ebabb47b1ed79128eb691c771235b1253f#diff-31022284a049ee7a04a54dfb02126be0826bfb351765db0437e86edf20af7190R1104)

radicle-node/src/test/peer.rs
xphoniex: I renamed all `subscribe`s to `event`, [see](https://github.com/radicle-dev/heartwood/compare/096b533df2b3e727e32ed084c3f5e301bafd5a6d..350133ebabb47b1ed79128eb691c771235b1253f#diff-85df6dd835d2ce99d0219ad5e696f3c00dc35d94ccf1876ec98d3da15a36883fR89)

cloudhead (approved):
:+1:

PR #414, opened 2023-03-11
node: Fetch missing inventory 

### LINKED ISSUE #364: When we track a repo for the first time, it should be fetched

cloudhead:
Periodically, make sure we try to fetch inventory that is tracked
but missing from local storage.

We decide not to fetch immediately on `TrackRepo`, as this limits
the command's flexibility. For example, when `rad clone` is used,
we want to be able to fetch manually after tracking.

Closes #364


FintanH (approved):
Nice, makes sense :+1:

PR #415, opened 2023-03-11
node: Rename `tracked_repo` -> `repo_policies`

cloudhead:
I was finding the current naming misleading, as the returned entries could be for blocked repos/nodes.

### DIFF
--- a/radicle-node/src/control.rs
+++ b/radicle-node/src/control.rs
@@ -131,7 +131,7 @@ fn command<H: Handle<Error = runtime::HandleError>>(
                 }
             }
         }
-        CommandName::TrackedRepos => match handle.tracked_repos() {
+        CommandName::RepoPolicies => match handle.repo_policies() {
             Ok(ts) => {
                 for t in ts.into_iter() {
                     serde_json::to_writer(&mut writer, &t)?;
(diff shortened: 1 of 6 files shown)


radicle-node/src/control.rs
```diff
@@ -131,7 +131,7 @@ fn command<H: Handle<Error = runtime::HandleError>>(
                 }
             }
         }
-        CommandName::TrackedRepos => match handle.tracked_repos() {
+        CommandName::TrackedRepos => match handle.repo_policies() {
```
FintanH, line 134: It would make sense to rename the commands too, I would say :)
```suggestion
        CommandName::RepoPolicies => match handle.repo_policies() {
```

radicle-node/src/runtime/handle.rs
```diff
@@ -136,10 +136,10 @@ impl<G: Signer + Ecdh + 'static> radicle::node::Handle for Handle<G> {
         receiver.recv().map_err(Error::from)
     }
 
-    fn tracked_repos(&self) -> Result<Self::TrackedRepos, Self::Error> {
+    fn repo_policies(&self) -> Result<Self::TrackedRepos, Self::Error> {
```
FintanH, line 139: And the associated types too
```suggestion
    fn repo_policies(&self) -> Result<Self::RepoPolicies, Self::Error> {
```

FintanH (commented):
Agree with the reasoning, left a couple of suggestions :)

radicle-node/src/control.rs
cloudhead: Oh good catch!

PR #416, opened 2023-03-12
Flesh out `CONTRIBUTING.md`

cloudhead:
Add code guidelines and instructions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,20 +3,155 @@
 Contributions are very welcome. When contributing code, please follow these
 simple guidelines.
 
-* Make sure you run `rustfmt` on your code. Also ensure all trailing whitespace is trimmed.
-* Run the tests with `cargo test --all`.
-* Before adding any code dependencies, check with the maintainers if this is okay.
-* Write properly formatted comments: they should be English sentences, eg:
+* Follow the coding guidelines when proposing code changes ([see below](#code-style)).
+* Write properly formatted git commits messages ([see below](#writing-git-commit-messages)).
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -3,20 +3,135 @@
...
+    let r = RefString::try_from(node.to_string()).unwrap();
+
+Use `expect` only if the function expects certain invariants that were not met,
+either due to bad inputs, or a problem with the environment; and include the
+expectation in the message. For example:
+
+    logger::init(log::Level::Debug)
+        .expect("logger must only be initialized once");
+
+### Module imports
```
FintanH, line 60: What about `mod` declarations?

CONTRIBUTING.md
```diff
@@ -3,20 +3,135 @@
...
+    }
+
+Use the most descriptive names for globals:
+
+    pub const KEEP_ALIVE_DELTA: LocalDuration = LocalDuration::from_secs(30);
+
+### Function naming
+
+Stay concise. Use the function doc comment to describe
+what the function does, not the name. Keep in mind function are in the
```
FintanH, line 97:
```suggestion
what the function does, not the name. Keep in mind functions are in the
```

CONTRIBUTING.md
FintanH: Aye, if that's what we do then it'd be good to state it :)

slack-coder:
What about a note re: checking issues + proposing changes there first?  Discussing the changes before putting the work in can save significant time and effort

cloudhead:
Yeah good point although I'd like to save that for when we can actually point people to `rad issue` etc.

FintanH (approved):
LGTM

PR #418, opened 2023-03-12
httpd: Basic alias support for git

cloudhead:
Allows cloning via git using an alias instead of the full RID.

Eg. `git clone https://seed.radicle.xyz/heartwood.git`


FintanH (commented):
Could we not also build the aliases from the projects themselves? :thinking:

cloudhead:
Yeah but that's trickier due to potentially duplicate names. This is also a bit more flexible, allowing multiple aliases per project as well as aliases different from the project name.

FintanH:
Ah of course, makes sense :+1:

PR #419, opened 2023-03-12
cli: Polish `assign` and `unassign` commands

cloudhead:
Just some code and ux improvements.

### DIFF
--- a/radicle-cli/src/commands/assign.rs
+++ b/radicle-cli/src/commands/assign.rs
@@ -2,61 +2,65 @@ use std::ffi::OsString;
 use std::str::FromStr;
 
 use anyhow::anyhow;
+use nonempty::NonEmpty;
 use radicle::prelude::Did;
 
 use crate::terminal as term;
-use crate::terminal::args;
+use crate::terminal::args::{Args, Error, Help};
 use radicle::cob;
 use radicle::cob::issue;
 use radicle::storage::WriteStorage;
 
-pub const HELP: args::Help = args::Help {
+pub const HELP: Help = Help {
     name: "assign",
-    description: "assign an issue",
+    description: "Assign an issue",
     version: env!("CARGO_PKG_VERSION"),
     usage: r#"
 Usage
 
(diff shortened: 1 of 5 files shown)


radicle-cli/src/commands/assign.rs
```diff
@@ -5,58 +5,58 @@ use anyhow::anyhow;
...
+    --to <did>    Assignee to add to the issue
+    --help        Print help
 "#,
 };
 
 #[derive(Debug)]
 pub struct Options {
     pub id: issue::IssueId,
-    pub peer: Did,
+    pub to: Did,
```
FintanH, line 32: Should we allow `Vec<Did>` while we're at it?

radicle-cli/src/commands/unassign.rs
```diff
@@ -12,23 +12,24 @@ use radicle::storage::WriteStorage;
...
+    --from <did>     Assignee to remove from the issue
+    --help           Print help
 "#,
 };
 
 #[derive(Debug)]
 pub struct Options {
     pub id: issue::IssueId,
-    pub peer: Did,
+    pub from: Did,
```
FintanH, line 32: Same here

FintanH (approved):
Looks good but suggesting that we allow for multiple `Did`s

radicle-cli/src/commands/assign.rs
cloudhead: Actually I was thinking multiple issues would be more useful to start, eg. assigning someone to multiple issues: `rad assign id1 id2 id3 --to finto`

cloudhead: As usually we only have one person assigned per issue. But yeah, at the COB level what you proposed is "free".

FintanH: Ya, I presumed the assigning can have multiple people since the code was using `[]`

FintanH (approved):
:kissing_cat: :ok_hand:

Issue #397, opened 2023-03-07
Using Force for Fetch Refspecs


FintanH:
The `Namespaces` type provides refspecs for the fetching. At the moment, these refspecs do not use `+` for forcing non-fast-forwards. 

On one hand, we would like to use fast-forwards when a peer has force-pushed changes to their reference set, otherwise, these would currently be rejected.

On the other hand, for example, if we wanted Eve's changes and we fetched them from Alice and Bob, where Alice has the most up-to-date state and Bob is behind, then we will end up in a behind state if we fetch from Bob after Alice. Unless... Would we reject the fetch from Bob because the signed refs are not a fast-forward?

So rubberducking as I write this... The invariant is that signed refs _always_ need to be a fast-forward, which we should fetch and verify first, and then all references in that set are force fetched because we know it's safe. In the case above, we reject the fetch from Bob because we know the signed refs are ahead -- we just fetched the latest from Alice.

cloudhead:
:bulb: !

FintanH:
In order for this to work correctly we need to split the fetching from a node into two stages:
1. Fetch the `rad` references -- these are needed to determine the validity of the identities and the signed refs.
2. For each remote we are fetching from, fetch the valid signed refs using the forced (`+`) refspec.

Something to be aware of is that step 1. can fail verification, for example, if a malicious actor is trying to fake signed refs. We don't want to overwrite the existing references so we need to have some kind of staging process. In the prior art, i.e. `link-replication`, these references were fetched into an in-memory refdb for verification purposes. However, this requires being able to tap into a custom protocol. For our case, it should be enough to fetch into a `staging` area, e.g. `refs/namespaces/staging`. If any of those references fail to verify then they can be pruned.

For each staged set of signed refs, we can ensure that they are a fast-forward of our currently seen signed refs, if they are not they can be discarded as they are out of date. If they are a fast-forward we can move forward the real signed refs reference and for each reference in its payload create a forced refspec. We send those refspecs off for fetching, completing phase 2, and the whole fetch process.

cloudhead:
Related: #14

PR #420, opened 2023-03-13
fix namespaces_for in scope::Trusted scenarios and its test

keepsimple1:
I noticed that [`test_fetch_trusted_remotes`](https://github.com/radicle-dev/heartwood/blob/f8aa1daedb391069847c66e14b2da6cc2e5669ae/radicle-node/src/tests/e2e.rs#L225-L227) was not working (ignored currently), and then I looked into it.  Hence this fix. 

Also updated other related tests.

This fix is more or less best effort and a learning experience for me. I felt there should be more refactoring we can do to improve this.

### DIFF
--- a/radicle-cli/tests/commands.rs
+++ b/radicle-cli/tests/commands.rs
@@ -496,6 +496,15 @@ fn test_replication_via_seed() {
         .rad("track", &[&bob.id.to_human()], working.join("alice"))
         .unwrap();
 
+    // Track nodes as the default scope is Trusted.
+    alice
+        .handle
+        .track_node(bob.id, Some("bob".to_string()))
+        .unwrap();
+    bob.handle
+        .track_node(alice.id, Some("alice".to_string()))
+        .unwrap();
+
     alice.routes_to(&[(rid, alice.id), (rid, seed.id)]);
     seed.routes_to(&[(rid, alice.id), (rid, seed.id)]);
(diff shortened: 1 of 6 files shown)


FintanH:
The issue is actually with how the fetching should work, as I detailed here https://github.com/radicle-dev/heartwood/issues/397

The TL;DR is that fetching should happen in multiple steps:
* Fetch trusted `rad` branches
* Fetch trusted references based on signed refs found under `rad/sigrefs`

In the case of the repository not existing, we need to fetch the `rad` branches to determine the trusted set, i.e. the delegates. I think you're right though that we can do this by fetching the remote we're talking to get these initial `rad` references and determine the rest from there.

That being said, I don't think this change is what we need.

keepsimple1:
> In the case of the repository not existing, we need to fetch the `rad` branches to determine the trusted set, i.e. the delegates. I think you're right though that we can do this by fetching the remote we're talking to get these initial `rad` references and determine the rest from there.

Yes, this is the case that this patch tries to solve. Fetching from the initial remote/seed is the best thing I can find at this point. But I might missed something better as I am still learning the fetch logic. I will merge this patch with https://github.com/radicle-dev/heartwood/pull/370 as I'm trying to reuse `namespaces_for` over there.

keepsimple1:
Covered by #370 .

PR #421, opened 2023-03-13
radicle-node: add tracking-scope option

FintanH:
The radicle-node was forced into using Scope::Trusted for its default scope.

Allow the Scope to be specified at the command line using a `--tracking-scope` option.


sebastinez (approved):
LGTM! Thanks @FintanH for the quick response!

cloudhead (approved):

PR #424, opened 2023-03-13
cli: Implement `tag` and `untag`

### LINKED ISSUE #344: radicle-cli: Add ability to tag issues
Currently it isn't possible to assign, unassign someone or add, remove a tag from an issue in the CLI.

I propose to:
- Add the ability to add assignees and tags to the `rad issue open` command.
- Add new commands `rad issue assign <did>`, `rad issue unassign <did>`
- Add new commands `rad issue tag <tag>`, `rad issue untag <tag>`

We could also use a `rad issue edit <id>` where we could edit assignees and tags somehow, I imagine maybe show inside the chosen editor a frontmatter formatted text where we take edits, and then apply them to the corresponding issue COB.

cloudhead:
These commands are added at the top level because they will eventually also work on patches.

We can also eventually implement an alias `rad issue tag` that just calls `rad tag` but forces the argument to be an issue.

Closes #344

### DIFF
--- a/radicle-cli/src/commands/untag.rs
+++ b/radicle-cli/src/commands/untag.rs
@@ -1,0 +1,91 @@
+use std::ffi::OsString;
+use std::str::FromStr;
+
+use anyhow::anyhow;
+use nonempty::NonEmpty;
+
+use crate::terminal as term;
+use crate::terminal::args::{Args, Error, Help};
+use radicle::cob;
+use radicle::cob::common::Tag;
+use radicle::cob::issue;
+use radicle::storage::WriteStorage;
+
+pub const HELP: Help = Help {
+    name: "untag",
+    description: "Untag an issue",
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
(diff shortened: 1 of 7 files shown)


radicle-cli/src/commands/untag.rs
```diff
@@ -0,0 +1,91 @@
...
+    let (_, id) = radicle::rad::cwd()?;
+    let repo = profile.storage.repository_mut(id)?;
+    let mut issues = issue::Issues::open(&repo)?;
+    let mut issue = issues.get_mut(&options.id).map_err(|e| match e {
+        cob::store::Error::NotFound(_, _) => anyhow!("issue {} not found", options.id),
+        _ => e.into(),
+    })?;
+    let signer = term::signer(&profile)?;
+
+    issue.tag([], options.tags.into_iter(), &signer)?;
```
FintanH, line 88: I hope we never mess the order of these parameters up :P

FintanH (approved):
Nice

PR #426, opened 2023-03-13
Safe Reaction

FintanH:
The Reaction type needs to ensure that the `char` stored for its value
meets certain requirements -- via its `new` constructor.

To ensure Reaction is always constructed correctly, the `pub` accessor
is removed, and the Deserialize implementation is implemented by hand
-- as opposed to using derive.


sebastinez (approved):
LGTM, have to still look why the httpd doesn't want to return reactions, but maybe this helps :+1:

cloudhead:
The name of this PR makes me think of an optional feature "safe reactions" that hides unsafe reactions from the user :smile:

PR #427, opened 2023-03-13
node: Fix reconnection logic

cloudhead:
*See commits*.

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -637,14 +637,15 @@ where
         // Inbound connection attempt.
     }
 
-    pub fn attempted(&mut self, id: NodeId, addr: &Address) {
-        debug!(target: "service", "Attempted connection to {id} ({addr})");
+    pub fn attempted(&mut self, nid: NodeId, addr: &Address) {
+        debug!(target: "service", "Attempted connection to {nid} ({addr})");
 
-        let persistent = self.config.is_persistent(&id);
-        self.sessions
-            .entry(id)
-            .and_modify(|sess| sess.to_connecting())
-            .or_insert_with(|| Session::connecting(id, persistent, self.rng.clone()));
+        if let Some(sess) = self.sessions.get_mut(&nid) {
+            sess.to_attempted();
+        } else {
+            #[cfg(debug_assertions)]
+            panic!("Service::attempted: unknown session {nid}@{addr}");
+        }
     }
 
     pub fn connected(&mut self, remote: NodeId, link: Link) {
@@ -666,9 +667,8 @@ where
                     );
                 }
                 Entry::Vacant(e) => {
-                    let peer = e.insert(Session::connected(
+                    let peer = e.insert(Session::inbound(
                         remote,
-                        Link::Inbound,
                         self.config.is_persistent(&remote),
                         self.rng.clone(),
                         self.clock,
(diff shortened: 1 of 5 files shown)


radicle-node/src/service.rs
```diff
@@ -630,14 +637,15 @@ where
...
-        self.sessions
-            .entry(id)
-            .and_modify(|sess| sess.to_connecting())
-            .or_insert_with(|| Session::connecting(id, persistent, self.rng.clone()));
+        if let Some(sess) = self.sessions.get_mut(&nid) {
+            sess.to_attempted();
+        } else {
+            #[cfg(debug_assertions)]
+            panic!("Service::attempted: unknown session {nid}@{addr}");
+        }
```
slack-coder, line 648: let () else {}?
```
let Some(sess) = self.sessions.get_mut(..) else {
          #[cfg(debug_assertions)]
            panic!("Service::attempted: unknown session {nid}@{addr}");
}
sess.to_attempted();
```

radicle-node/src/service.rs
```diff
@@ -690,31 +697,30 @@ where
...
+                    self.sessions.remove(&remote);
                 }
             } else {
                 self.sessions.remove(&remote);
                 self.maintain_connections();
             }
+        } else {
+            #[cfg(debug_assertions)]
+            panic!("Service::disconnected: unknown session {remote}");
         }
```
slack-coder, line 724: `let Some(..) = () else {}`?

radicle-node/src/service/session.rs
```diff
@@ -38,7 +38,9 @@ impl Default for Protocol {
 #[allow(clippy::large_enum_variant)]
 pub enum State {
     /// Initial state for outgoing connections.
-    Connecting,
+    Initial,
+    /// Connection attempted successfully.
+    Attempted,
```
slack-coder, line 43: Problem with Attempted is attempted things can have failed.
What about?
 - Negotiating or
 - Attempted => Connected, Connected => Established

radicle-node/src/service.rs
```diff
@@ -1433,6 +1436,33 @@ where
...
+        for (nid, session) in self.sessions.disconnected_mut() {
+            if let Some(addr) = self.config.peer(nid) {
+                if let session::State::Disconnected { retry_at, .. } = &mut session.state {
+                    // TODO: Try to reconnect only if the peer was attempted. A disconnect without
+                    // even a successful attempt means that we're unlikely to be able to reconnect.
+
+                    if now >= *retry_at {
+                        // FIXME: Make sure we don't attempt two concurrent outgoing connections
+                        // to the same peer.
+                        reconnect.push((*nid, addr.clone(), session.attempts()));
```
FintanH: Could you easily ensure this by using:
```rust
let mut reconnect = BTreeMap<NodeId, (Addr, usize)>::new();
// ...
reconnect.insert(*nid, (addr.clone(), session.attempts()));
```

radicle-node/src/service.rs
```diff
@@ -436,6 +440,9 @@ where
             self.reactor.wakeup(PRUNE_INTERVAL);
             self.last_prune = now;
         }
+
+        // Always check whether there are persistent peers that need reconnecting.
+        self.maintain_persistent();
```
slack-coder, line 445: Would better naming the method remove the need for the comment?

radicle-node/src/service/session.rs
cloudhead: The reason they are called this is to match up with the callback functions of service, which are `attempted` -> `connected`. If we rename these states, we'd also have to rename those functions, and I'd rather leave that for a different patch.

radicle-node/src/service.rs
cloudhead: Don't think so. I left the comment because it seems "random" for this one method to be called down here after all the `if{}` statements, and I want to explain why we're calling it here.

radicle-node/src/service.rs
cloudhead: That's not the problem actually, the  problem is if this function gets called multiple times in a row; or right after a `Command::Connect` is received.

cloudhead: Hence the state of what is being attempted needs to be service-global and can't be local to this function.

radicle-node/src/service.rs
FintanH: `sess` won't be in scope then because it would need to be returned in the `else`

cloudhead: Yeah indeed, was thinking we need a macro that panics in debug mode and logs in release mode..

PR #428, opened 2023-03-14
httpd: Fix status code in commit history endpoint

rudolfs:
It seems like this was set to 302 by mistake. We're not setting any location headers, so when querying this endpoint with with `fetch` from a JS frontend and then do `.json()` on the result, it throws an error.

```
{
  error: TypeError: terminated
...
    [cause]: DOMException [AbortError]: The operation was aborted.
...
  }
}
```

### DIFF
--- a/radicle-httpd/src/api/v1/projects.rs
+++ b/radicle-httpd/src/api/v1/projects.rs
@@ -138,12 +138,12 @@ async fn history_handler(
         per_page,
     } = qs;
 
-    let (sha, fallback_to_head) = match parent {
-        Some(commit) => (commit, false),
+    let sha = match parent {
+        Some(commit) => commit,
         None => {
             let info = ctx.project_info(project)?;
 
-            (info.head.to_string(), true)
+            info.head.to_string()
         }
     };
 
@@ -200,10 +200,6 @@ async fn history_handler(
         "commits": commits,
         "stats":  repo.stats()?,
     });
-
-    if fallback_to_head {
-        return Ok::<_, Error>((StatusCode::FOUND, Json(response)));
-    }
 
     Ok::<_, Error>((StatusCode::OK, Json(response)))
 }
@@ -689,7 +685,7 @@ mod routes {
         let app = super::router(test::seed(tmp.path()));
         let response = get(&app, format!("/projects/{RID}/commits")).await;
 
-        assert_eq!(response.status(), StatusCode::FOUND);
+        assert_eq!(response.status(), StatusCode::OK);
         assert_eq!(
             response.json().await,
             json!({


xphoniex:
It's not by mistake, this is copied from [radicle-dev/radicle-client-services](https://github.com/radicle-dev/radicle-client-services/blame/29416ea9dc207074b4d9c84648fc648f46ce0bf2/http-api/src/lib.rs#L793) and the status code was chosen by @sebastinez .

rudolfs:
I'm not a 100% sure, but what I think is happening is that we're not adhering to the HTTP spec here and that's why the frontend fetch breaks.

> The server SHOULD generate a [Location](https://httpwg.org/specs/rfc9110.html#field.location) header field in the response containing a URI reference for the different URI. The user agent MAY use the Location field value for automatic redirection. The server's response content usually contains a short hypertext note with a hyperlink to the different URI(s).

Screenshot 2023-03-14 at 17 51 13

It seems that the `Location` header is missing. So if we want to keep the `302` we should implement it correctly.

xphoniex:
Ah, I see. That'd unnecessarily make other returns longer as well since they'd have to return `HeaderMap` too, so `StatusCode::OK` seems like a simpler solution.

xphoniex (approved):
lgtm

rudolfs:
For reference, here's how we can reproduce this error on the frontend (node v18.15.0):

```
node -e " ( async () => {const result = await globalThis.fetch('http://0.0.0.0:8080/api/v1/projects/rad:zKtT7DmF9H34KkvcKj9PHW19WzjT/commits', { method: 'GET', headers: {}, body: undefined, credentials: 'include', signal: undefined }); const json = await result.json(); console.log(json);})() "
```

cloudhead:
@sebastinez do you remember why we went with FOUND? I think it was to differentiate the fallback head from the one provided? But was there a reason?

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -202,7 +202,7 @@ async fn history_handler(
     });
 
     if fallback_to_head {
-        return Ok::<_, Error>((StatusCode::FOUND, Json(response)));
+        return Ok::<_, Error>((StatusCode::OK, Json(response)));
     }
 
     Ok::<_, Error>((StatusCode::OK, Json(response)))
```
sebastinez, line 208: If we change the status codes to OK 200 I suggest to remove the entire `fallback_to_head` if branch

```diff
diff --git a/radicle-httpd/src/api/v1/projects.rs b/radicle-httpd/src/api/v1/projects.rs
index d81b756..852ca48 100644
--- a/radicle-httpd/src/api/v1/projects.rs
+++ b/radicle-httpd/src/api/v1/projects.rs
@@ -201,10 +201,6 @@ async fn history_handler(
         "stats":  repo.stats()?,
     });
 
-    if fallback_to_head {
-        return Ok::<_, Error>((StatusCode::FOUND, Json(response)));
-    }
-
     Ok::<_, Error>((StatusCode::OK, Json(response)))
 }
```

rudolfs: fba250fa40a9764b67e9691e8bc4e5ff24c0d691

sebastinez:
> @sebastinez do you remember why we went with FOUND? I think it was to differentiate the fallback head from the one provided? But was there a reason?

Yes exactly, If I remember correctly we wanted to differentiate the status code we used, if we have to fallback to head.
But it seemed I messed implementing the location header.

rudolfs:
> > @sebastinez do you remember why we went with FOUND? I think it was to differentiate the fallback head from the one provided? But was there a reason?
> 
> Yes exactly, If I remember correctly we wanted to differentiate the status code we used, if we have to fallback to head. But it seemed I messed implementing the location header.

I don't think we differentiate by the status code anywhere in the app.

sebastinez:
> > > @sebastinez do you remember why we went with FOUND? I think it was to differentiate the fallback head from the one provided? But was there a reason?
> 
> > 
> 
> > Yes exactly, If I remember correctly we wanted to differentiate the status code we used, if we have to fallback to head. But it seemed I messed implementing the location header.
> 
> 
> 
> I don't think we differentiate by the status code anywhere in the app.

Yeah that's true it was more an http-api thing, the interface was happy with the body either way

PR #433, opened 2023-03-15
node: Move tracking store to `radicle` crate

cloudhead:
Simplifies code by allowing the CLI direct access to the routing and tracking databases.


radicle-node/src/control.rs
```diff
@@ -131,15 +131,6 @@ fn command<H: Handle<Error = runtime::HandleError>>(
...
         }
-        CommandName::RepoPolicies => match handle.repo_policies() {
-            Ok(ts) => {
-                for t in ts.into_iter() {
-                    serde_json::to_writer(&mut writer, &t)?;
-                    writeln!(writer)?;
-                }
-            }
-            Err(e) => return Err(CommandError::Runtime(e)),
-        },
```
FintanH, line 142: Oh, so we're not keeping these to allow for the node to also report them?

cloudhead: Nah it seems redundant to me. Better only have one way of doing this I think.

FintanH (approved):

PR #436, opened 2023-03-15
node: Fail fast when repository can't be read

cloudhead:
It's not a good idea to have fallible operations run after `FetchOk`
is received, since there's no easy way to communicate the error
to the remote at that point. Hence, we compute namespaces earlier
in the process, and store the namespaces in the session.

### DIFF
--- a/radicle-node/src/service.rs
+++ b/radicle-node/src/service.rs
@@ -32,6 +32,7 @@ use crate::prelude::*;
 use crate::service::message::{Announcement, AnnouncementMessage, Ping};
 use crate::service::message::{NodeAnnouncement, RefsAnnouncement};
 use crate::service::reactor::FetchDirection;
+use crate::service::session::GossipState;
 use crate::service::tracking::Scope;
 use crate::storage;
 use crate::storage::{Namespaces, ReadStorage};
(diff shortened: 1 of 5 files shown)


radicle-node/src/service.rs
```diff
@@ -538,8 +538,22 @@ where
...
+                                reason: err.to_string(),
+                            })
+                            .ok();
+                        }
+                        return;
+                    }
+                };
                 self.reactor.write(session, fetch);
-                session.to_requesting(rid);
+                session.to_requesting(rid, namespaces);
```
FintanH, line 556: We could move these lines into the `Ok` match arm

radicle-node/src/service/session.rs
```diff
@@ -18,10 +20,10 @@ pub enum PingState {
 }
 
 /// Session protocol.
-#[derive(Debug, Copy, PartialEq, Eq, Clone)]
+#[derive(Debug, PartialEq, Eq, Clone)]
 pub enum Protocol {
     /// The default message-based gossip protocol.
-    Gossip { requested: Option<Id> },
+    Gossip { requested: Option<(Id, Namespaces)> },
```
FintanH, line 26: I think I need some explanation on why `Namespaces` is being shoved in here :thinking:

PR #438, opened 2023-03-15
node: Fix state mismatch between service and wire

cloudhead:
Due to never removing peers from the peer map in the wire protocol module, it was possible for there to be multiple connected peers with the same Node ID. Some might be in a "connected" state, while others might be "disconnected".

This caused a bug where a peer, although connected would be considered disconnected by the wire module, because when searching for a peer with that Node ID, the disconnected one would return first.

To fix this, we make sure to remove peers from the map once the connection is dropped.


FintanH (approved):
Makes sense :+1:

PR #441, opened 2023-03-16
node: Re-organize imports for better readability

xphoniex:
I realize `radicle-node` does not adhere to how we import (`use`) other crates, and this was harder on eyes because the sorting doesn't take levels into account, for example, at a first glance it's harder to find `wire` import in the following:

```rust
use crate::address;
use crate::control;
use crate::crypto::Signer;
use crate::node::{routing, NodeId};
use crate::service::tracking;
use crate::wire;
use crate::wire::Wire;
use crate::worker;
```

versus

```rust
use crate::crypto::Signer;
use crate::node::{routing, NodeId};
use crate::service::{self, tracking};
use crate::wire::{self, Wire};
use crate::{address, control, worker, LocalTime};
```

I agree that importing `std::io::{self,..}` is not as easy to read as `std::{io, ..}` but we've went with former in the rest of places.


xphoniex:
also there seems to be some unnecessary imports like [this](https://github.com/radicle-dev/heartwood/pull/441/files#diff-31022284a049ee7a04a54dfb02126be0826bfb351765db0437e86edf20af7190L29)

cloudhead:
> also there seems to be some unnecessary imports like [this](https://github.com/radicle-dev/heartwood/pull/441/files#diff-31022284a049ee7a04a54dfb02126be0826bfb351765db0437e86edf20af7190L29)

The logging imports are used for testing, and the `allow(unused)` means we don't have to worry about whether it's imported or not.

cloudhead:
Closing this, since it's inconsistent with our other crates.

PR #443, opened 2023-03-16
node: Use correct scope when getting namespaces

cloudhead:
We were using the default scope instead of the entry scope.


FintanH (approved):
Nice catch :+1:

PR #451, opened 2023-03-17
Patch fixes

cloudhead:
*See commits.*

### DIFF
--- a/radicle-cli/src/commands/patch/checkout.rs
+++ b/radicle-cli/src/commands/patch/checkout.rs
@@ -1,38 +1,85 @@
-use crate::terminal as term;
 use anyhow::anyhow;
-use radicle::cob::patch::{self, PatchId};
-use radicle::git::{self, RefString};
+
+use radicle::cob::patch;
+use radicle::cob::patch::{Patch, PatchId};
+use radicle::git;
+use radicle::git::RefString;
 use radicle::storage::git::Repository;
+use radicle::storage::ReadRepository;
+
+use crate::terminal as term;
 
 pub fn run(
-    storage: &Repository,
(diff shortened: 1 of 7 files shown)


radicle-cli/src/commands/patch/checkout.rs
```diff
@@ -1,38 +1,77 @@
...
+            );
+            let url = git::Url::from(stored.id).with_namespace(author);
+
+            // Fetch only the ref pointing to the patch revision.
+            working.remote_anonymous(url.to_string().as_str())?.fetch(
+                &[&format!("{refstr}:{remote_branch}")],
+                None,
+                None,
+            )?;
+            working.find_commit(patch_head.into())?
```
slack-coder, line 58: Maybe move this code into a private method?  That way the condition and resulting actions are clear.

slack-coder:
Otherwise :+1:

radicle-cli/src/commands/patch/checkout.rs
```diff
@@ -1,38 +1,85 @@
...
+use radicle::git::RefString;
 use radicle::storage::git::Repository;
+use radicle::storage::ReadRepository;
+
+use crate::terminal as term;
 
 pub fn run(
-    storage: &Repository,
-    git_workdir: &git::raw::Repository,
+    stored: &Repository,
```
FintanH, line 13: Why the rename, out of interest?

cloudhead: `storage` is confusing when it doesn't refer to `Storage`, no?

I think a nice convention if there are two repositories is `stored` for the stored copy and `working` for the working copy. That was the idea.

Issue #337, opened 2023-02-19
Document `/mnt/radicle/heartwood` in docker-compose.yml


gsaslis:
I tried cloning the repo and running `docker-compose up` and the images built correctly, but when they actually started I got an error message because `/mnt/radicle/heartwood` didn't exist on my machine. 

I tried replacing that with an empty folder, but: 

* radicle-node fails with: 
```
radicle-node     | 2023-02-19T11:44:21.387Z ERROR node     Fatal: key not found in '/root/.radicle/keys'
```

* and radicle-httpd fails with: 
```
radicle-httpd    | 2023-02-19T11:44:21.353758Z ERROR radicle_httpd: Fatal: no profile found at the filepath '/root/.radicle'
```

I guess there are some files I am missing in that path ? Do I need to run anything in particular to initialise that folder **before** running `docker-compose up` ?

gsaslis:
I was able to work around this by first running `rad auth` (with the latest pre-built binary) and then pointing the mount point to `~/.radicle` (`/Users/$USER/.radicle` on macOS)

gsaslis:
(but the issue about documenting that path still remains, I think)

cloudhead:
@adaszko wdyt?

adaszko:
Oh, it most definitely should be a changed to a different directory in your running it on your machine.  The /mnt directory was meant for server use and is specific to our infra.  I think a comment recommending editing this path should suffice

cloudhead:
Yeah, we don't really want users running via docker locally I think.

gsaslis:
Yes, what wasn't clear was what is expected to be contained in that directory. 

From that point onwards, it sounds like more of a user choice how they run the node (and there's a point to be made that if the Radicle team deploys theirs via docker, that's a good place to start... ;) )

adaszko:
If we could somehow embed the git binary inside of radicle-node and radicle-httpd, we’d be on a way to ship just a single self-contained binary which I think would be as simple as it gets.  Is such a thing even feasible? I.e. could we use libgit in places where we currently use the git executable?

cloudhead:
> If we could somehow embed the git binary inside of radicle-node and radicle-httpd, we’d be on a way to ship just a single self-contained binary which I think would be as simple as it gets. Is such a thing even feasible? I.e. could we use libgit in places where we currently use the git executable?

I wish...

[...]

PR #452, opened 2023-03-17
Run with docker-compose

gsaslis:
This PR addresses some of the issues raised in #337 and additionally adds some instructions about how to run radicle-node and radicle-httpd in containers.

### DIFF
--- a/RUNNING_IN_CONTAINERS.md
+++ b/RUNNING_IN_CONTAINERS.md
@@ -21,70 +21,12 @@ dependencies and then compile the code:
 docker-compose build
 ```
 
-
-## 3. Choose scenario
-
-Choose your scenario:
-
-* [Run a Radicle node on my machine, using docker](#run-a-single-radicle-node)
-* [Run multiple Radicle nodes on my machine, to try things out / learn how Radicle works.](#run-multiple-radicle-nodes-locally) 
-
-### 3A. Run a single Radicle node 
+## 3. Start the containers 
 
 This is as simple as using the existing `docker-compose.yml` and simply passing in 
 the path to the **parent** folder of `RAD_HOME`, where you previously created the 
-profile.
+profile, as per the instructions in #1 above.
 
 ```bash
 RAD_ROOT=~/radicle/profiles/bob docker-compose up radicle-node radicle-httpd 
 ```
-
-### 3B. Run Multiple Radicle Nodes Locally
-
-You would probably use this setup if you want to play around with multiple nodes and
-study, understand and learn how Radicle works and how projects (and their data) are 
-replicated across nodes. **This setup is not recommended for production use!**
-
-The setup is slightly different between the first node you run and the additional ones. 
-
-#### Run First Node
-
-For the first node you want to start on your machine, you can:
-
-1. Create a `.env.$nodename` file that will store all your environment variables:
-   ```yaml
-   # these options are especially useful in a development setting - probably not for production use
-   RADICLE_NODE_OPTIONS=--tracking-policy track --tracking-scope all
-   # Note the difference between RAD_ROOT vs. RAD_HOME.
-   RAD_ROOT=~/radicle/profiles/bob
-   # ensure these ports are free on your machine
-   RADICLE_API_PORT=8888
-   RADICLE_NODE_PORT=8778
-   ```
-1. Start the containers:
-   ```bash
-   # we don't need to start the included `caddy` service
-   docker-compose --env-file=.env.bob --project-name=bob up radicle-node radicle-httpd
-   ```
-
-#### Run additional nodes in your network
-
-For each additional node: 
-1. Create a new profile in a new directory, using the steps in the "Create a profile" section above. 
-1. Create a `.env.$nodename` file that will store all your environment variables:
-   ```dotenv
-   # IMPORTANT: substitute `<FIRST_NODE_ID>` with the node id of your **first** node ( `RAD_HOME=<path_to_first_node_profile> rad self | grep "Node ID"` ). 
-   # IMPORTANT: substitute `<FIRST_NODE_PROJECT_NAME>` with the `--project-name` value you used in your first node. In our example, this would be `bob`.
-   RADICLE_NODE_OPTIONS=--tracking-policy track --tracking-scope all  --connect <FIRST_NODE_ID>@<FIRST_NODE_PROJECT_NAME>_radicle-node_1.radicle-services:8776
-   # Use the new profile directory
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -34,16 +33,17 @@ services:
       RUST_LOG: debug
       RUST_BACKTRACE: 1
     volumes:
-      - /mnt/radicle/heartwood:/root/
+      - ${RAD_ROOT:-/mnt/radicle/heartwood}:/root/
```
adaszko, line 36: Just to make this more convenient for a "laptop" use, I wouldn't mind if we picked a sensible default, e.g. `${RAD_ROOT:-$HOME/.radicle/}` (as an aside: I'd be in favor or `~/.local/state/radicle` to conform with XDG specs).  We could flip it such that I'd override the default on the server via `.env.seed` file.  One fewer environment variable to configure that way.

RUNNING_IN_CONTAINERS.md
```diff
@@ -0,0 +1,62 @@
...
+   - your profile should be created in `~/radicle/profiles/bob/.radicle`. 
+
+## Build the container images
+
+This takes a couple of minutes - depending on your machine - as it needs 
+to download the parent container images, and also all the rust 
+dependencies and then compile the code: 
+
+```bash
+docker-compose build
```
adaszko, line 22: Or we could open up our GCP containers repository to be public to make this easier.  It's got pre-built images that can be simply downloaded and run.  CC @cloudhead

RUNNING_IN_CONTAINERS.md
```diff
@@ -0,0 +1,62 @@
...
+```bash
+# we don't need to start the included `caddy` service
+docker-compose --env-file=.env.bob --project-name=bob up radicle-node radicle-httpd
+```
+
+## Run additional nodes in your network
+
+For each additional node: 
+1. Create a new profile in a new directory, using the steps in the "Create a profile" section above. 
+1. Create a `.env.$nodename` file that will store all your environment variables:
```
adaszko, line 46: Perhaps this is just a bit *too* flexible?  Wouldn't `.env` work fine?  The additional option `--env-file=.env.bob` could be eliminated that way.

RUNNING_IN_CONTAINERS.md
```diff
@@ -0,0 +1,62 @@
...
+## Run First Node 
+1. Create a `.env.$nodename` file that will store all your environment variables:
+```yaml
+# these options are especially useful in a development setting - probably not for production use
+RADICLE_NODE_OPTIONS=--tracking-policy track --tracking-scope all
+# Note the difference between RAD_ROOT vs. RAD_HOME.
+RAD_ROOT=~/radicle/profiles/bob
+# ensure these ports are free on your machine
+RADICLE_API_PORT=8888
+RADICLE_NODE_PORT=8778
```
adaszko, line 34: Maybe we could rely on the defaults in this file to reduce the cognitive overhead and add a section with optional steps about overriding default port values in case there's a conflict?  Only a minority of users are going to need to configure that I imagine so we might as well sidetrack it.

adaszko:
Nice job overall 👍.  Let me know if you have some ideas on who we could lower the bar to deployment for end users

gsaslis:
@adaszko the thing is that running multiple nodes on your machine has value when trying to understand how the p2p network works and how a repo (and all its data) is replicated across nodes. You want to see it happening to understand how it works (... and to verify that it ... does work! 🙃 )

Because of that, I think this is an important scenario to capture, BUT I think you also have a point in that, if someone only wants to run a single node on their machine, the current instructions do add quite a bit of mental overhead. 

Perhaps I could split the markdown in 2 different sections? 🤔 

One for running a single node (that's much simpler) and one that caters for the current multi-node scenario ?

adaszko:
> @adaszko the thing is that running multiple nodes on your machine has value when trying to understand how the p2p network works and how a repo (and all its data) is replicated across nodes. You want to see it happening to understand how it works (... and to verify that it ... does work! 🙃 )

I see, didn't think of that use case.

> Because of that, I think this is an important scenario to capture, BUT I think you also have a point in that, if someone only wants to run a single node on their machine, the current instructions do add quite a bit of mental overhead.
> 
> Perhaps I could split the markdown in 2 different sections? 🤔
> 
> One for running a single node (that's much simpler) and one that caters for the current multi-node scenario ?

Braking it down sounds great 👍

gsaslis:
@adaszko i just pushed a new commit. Please feel free to take another look when you get a chance. ;) 👍🏼

docker-compose.yml
```diff
@@ -8,14 +8,13 @@ services:
...
     environment:
       RUST_LOG: debug
-      RAD_PASSPHRASE: seed
+      RAD_PASSPHRASE: ${RAD_PASSPHRASE:-seed}
       RUST_BACKTRACE: 1
       GIT_TRACE: 1
       GIT_TRACE_PACKET: 1
     volumes:
-      - /mnt/radicle/heartwood:/root/
+      - ${RAD_ROOT:-/mnt/radicle/heartwood}:/root/
```
cloudhead, line 16: Hmm this should just be `$RAD_HOME` I think?

gsaslis: well, I originally thought the same, and this is why I opened #337

adaszko (approved):
We could iterate further on eliminating the build step (just rely on downloading images) and improving the default value of `RAD_ROOT` but it looks simpler now 👍

cloudhead:
I definitely think we need documentation on how to run radicle node in docker, but I'm not sure this is the place for documenting multi-node setups in docker. This is not a setup we currently use, and therefore the documentation would eventually get outdated.

gsaslis:
@cloudhead ok, I can move the docs in **3B** to some other place - no probs. 

would the rest still make sense as part of this repo ?

cloudhead:
Yes, it makes sense, but I think if you remove the multi-node stuff, the rest needs to be adapted too, eg. it doesn't make sense to talk about profiles and "bob" etc.

docker-compose.yml
gsaslis: I am resolving this, as I think we've agreed on this in the main discussion of the PR.

gsaslis:
@cloudhead does this look better? anything else you'd like to see changed? 

I think the profile creation section helps in providing a start-to-finish guide with no missing pieces, so could we keep that here?

cloudhead:
I'm going to close this as I don't think it's what we need in terms of container docs. We can discuss further and open a patch on radicle at that time.

PR #461, opened 2023-03-20
Enable `rad merge` to work on concurrently updated patches

slack-coder:
`rad merge` originally takes an optional revision number to define which version of a patch to merge.  The problem is a patch can be updated concurrently, multiple versions of the patch could exist with the same version number.

Therefor use the patch's globally unique revision -id- instead of its index number.

Related to https://github.com/radicle-dev/heartwood/issues/148

### DIFF
--- a/radicle-cli/examples/workflow/5-patching-maintainer.md
+++ b/radicle-cli/examples/workflow/5-patching-maintainer.md
@@ -54,3 +54,36 @@ a07ef77 R1 (27857ec) -> R2 (f6484e0)
 
 ✓ Patch a07ef77 updated to 737dc47d9eba730c5db8a8e33c41c7955f9093de
 ```
+
+Great, all fixed up, lets merge the code.
+
+```
+$ git checkout master
+Your branch is up to date with 'rad/master'.
+$ rad merge 737dc47d9eba730c5db8a8e33c41c7955f9093de
+Merging a07ef77 R2 (f6484e0) by did:key:z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk into master (f2de534) via fast-forward...
+Running `git merge --ff-only f6484e0f43e48a8983b9b39bf9bd4cd889f1d520`...
+Updating f2de534..f6484e0
+Fast-forward
+ README.md       | 0
+ REQUIREMENTS.md | 0
+ 2 files changed, 0 insertions(+), 0 deletions(-)
+ create mode 100644 README.md
+ create mode 100644 REQUIREMENTS.md
+✓ Updated master f2de534 -> f6484e0 via fast-forward
+✓ Patch state updated, use `rad push` to publish
+```
+
+The patch is now merged and closed :).
+
+```
+$ rad patch
+╭───────────────────────────────────────────────────────────────────────────────────────────────╮
+│ Define power requirements a07ef77 R2 f6484e0 (flux-capacitor-power, master) ahead 3, behind 0 │
+├───────────────────────────────────────────────────────────────────────────────────────────────┤
+│ ● opened by did:key:z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk 3 months ago             │
+│ ↑ updated to a07ef7743a32a2e902672ea3526d1db6ee08108a (3e674d1) 3 months ago                  │
+│ ↑ updated to 4f15eb5c994edd0bb6be29cb4801aa74308cc628 (27857ec) 3 months ago                  │
+│ ✓ merged by did:key:z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi (you) 3 months ago       │
+╰───────────────────────────────────────────────────────────────────────────────────────────────╯
+```
--- a/radicle-cli/src/commands/merge.rs
+++ b/radicle-cli/src/commands/merge.rs
@@ -1,14 +1,12 @@
 use std::ffi::OsString;
 use std::fmt;
 use std::fmt::Write;
-use std::str::FromStr;
 
 use anyhow::{anyhow, Context};
 
 use crate::git::Rev;
 use crate::terminal as term;
 use crate::terminal::args::{string, Args, Error, Help};
-use radicle::cob::patch::RevisionIx;
 use radicle::cob::patch::{Patch, PatchId, Patches};
 use radicle::git;
 use radicle::prelude::*;
@@ -21,14 +19,14 @@ pub const HELP: Help = Help {
     usage: r#"
 Usage
 
-    rad merge [<id>] [<option>...]
+    rad merge [<revision-id>] [<option>...]
 
-    To specify a patch to merge, use the fully qualified patch id.
+    To specify a patch revision to merge, use the fully qualified revision id.
 
 Options
 
+    -f, --force               Force merging an older patch revision
     -i, --interactive         Ask for confirmations
-    -r, --revision <number>   Revision number to merge, defaults to the latest
         --help                Print help
 "#,
 };
@@ -71,11 +69,12 @@ pub enum State {
     Open,
     Merged,
 }
+
 #[derive(Debug)]
 pub struct Options {
-    pub id: Rev,
+    pub revision_id: Rev,
+    pub force: bool,
     pub interactive: bool,
-    pub revision: Option<RevisionIx>,
 }
 
 impl Args for Options {
@@ -83,8 +82,8 @@ impl Args for Options {
         use lexopt::prelude::*;
 
         let mut parser = lexopt::Parser::from_args(args);
-        let mut id: Option<Rev> = None;
-        let mut revision: Option<RevisionIx> = None;
+        let mut force = false;
+        let mut revision_id = None;
         let mut interactive = false;
 
         while let Some(arg) = parser.next()? {
@@ -92,30 +91,28 @@ impl Args for Options {
                 Long("help") => {
                     return Err(Error::Help.into());
                 }
+                Long("force") | Short('f') => {
+                    force = true;
+                }
                 Long("interactive") | Short('i') => {
                     interactive = true;
                 }
-                Long("revision") | Short('r') => {
-                    let value = parser.value()?;
-                    let id =
-                        RevisionIx::from_str(value.to_str().unwrap_or_default()).map_err(|_| {
-                            anyhow!("invalid revision number `{}`", value.to_string_lossy())
-                        })?;
-                    revision = Some(id);
-                }
                 Value(val) => {
                     let val = string(&val);
-                    id = Some(Rev::from(val));
+                    revision_id = Some(Rev::from(val));
                 }
                 _ => return Err(anyhow::anyhow!(arg.unexpected())),
             }
         }
 
+        let revision_id =
+            revision_id.ok_or_else(|| anyhow!("a revision id to merge must be provided"))?;
+
         Ok((
             Options {
-                id: id.ok_or_else(|| anyhow!("a patch id to merge must be provided"))?,
+                revision_id,
+                force,
                 interactive,
-                revision,
             },
             vec![],
         ))
@@ -144,10 +141,27 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
     //
     // Get patch information
     //
-    let patch_id = options.id.resolve(&repository.backend)?;
+    let revision_id = options.revision_id.resolve(&repository.backend)?;
+    let (patch_id, patch, revision) = patches.find_by_revision(&revision_id)?.ok_or(anyhow!(
+        "no open patch with revision `{}` could be found",
+        &revision_id
+    ))?;
+    if !patch.is_open() {
+        anyhow::bail!(
+            "revision's patch must be open for merging, but it is `{}`",
+            patch.state()
+        );
+    }
+    let (last_revision_id, _) = patch
+        .latest()
+        .ok_or(anyhow!("patch must have atleast one unredacted revision"))?;
+    if !options.force && revision_id != *last_revision_id {
+        anyhow::bail!("refusing to merge old patch revision");
+    }
+
     let mut patch = patches
         .get_mut(&patch_id)
-        .map_err(|e| anyhow!("couldn't find patch {} locally: {e}", &options.id.clone()))?;
+        .map_err(|e| anyhow!("couldn't find patch {} locally: {e}", &id))?;
 
     let head = repo.head()?;
     let branch = head
@@ -156,11 +170,6 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
     let head_oid = head
         .target()
         .ok_or_else(|| anyhow!("cannot merge into detatched head; aborting"))?;
-    let revision_ix = options.revision.unwrap_or_else(|| patch.version());
-    let (revision_id, revision) = patch
-        .revisions()
-        .nth(revision_ix)
-        .ok_or_else(|| anyhow!("revision R{} does not exist", revision_ix))?;
 
     //
     // Analyze merge
@@ -221,6 +230,11 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
             .to_string(),
     };
 
+    // SAFETY: The patch has already been fetched by its revision_id.
+    let revision_ix = patch
+        .revisions()
+        .position(|(id_, _)| id_ == &revision_id)
+        .unwrap();
     term::info!(
         "{} {} {} {} by {} into {} {} via {}...",
         term::format::bold("Merging"),
@@ -261,7 +275,7 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
     // Update patch COB
     //
     // TODO: Don't allow merging the same revision twice?
-    patch.merge(*revision_id, head_oid.into(), &signer)?;
+    patch.merge(revision_id, head_oid.into(), &signer)?;
 
     term::success!(
         "Patch state updated, use {} to publish",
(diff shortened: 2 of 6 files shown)


cloudhead:
Can we get a proper PR title?

slack-coder:
@cloudhead what should be done about the use of revision indexes?  I believe the revision id's should take their place, but in what format should that happen?

radicle-cli/examples/rad-patch.md
```diff
@@ -52,6 +52,7 @@ $ rad patch
 Define power requirements 191a14e520f R0 3e674d1 (flux-capacitor-power) ahead 1, behind 0
 └─ * opened by did:key:z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi (you) [..]
 └─ * patch id 191a14e520f2eeff7c0e3ee0a5523c5217eecb89
+└─ * revision id 191a14e520f2eeff7c0e3ee0a5523c5217eecb89
```
cloudhead, line 55: Keep in mind the first revision's id is equal to the patch's id. So I wouldn't show it here unless they are different.

radicle-cli/src/commands/merge.rs
```diff
@@ -95,35 +102,31 @@ impl Args for Options {
...
-        Ok((
-            Options {
-                id: id.ok_or_else(|| anyhow!("a patch id to merge must be provided"))?,
-                interactive,
-                revision,
-            },
-            vec![],
-        ))
+        let str_id =
+            str_id.ok_or_else(|| anyhow!("a patch or revision id to merge must be provided"))?;
```
cloudhead, line 115: Now that I think about it more, I think we should only support revision-ids, since the patch id is the id of the first revision!

radicle/src/cob/patch.rs
```diff
@@ -849,6 +849,20 @@ impl<'a> Patches<'a> {
...
+    /// Find the patch containing the given `Revision`.
+    pub fn find_with_revision(
+        &self,
+        id: &RevisionId,
+    ) -> Result<Option<(PatchId, Patch)>, store::Error> {
+        let all = self.all()?;
+        let result = all
+            .into_iter()
+            .filter_map(|result| result.ok())
+            .find(|(_, p, _)| p.revisions().any(|(i, _)| i == id))
```
cloudhead, line 861: I would first filter for open patches, no need to search archived patches.
Also, for patches with a single revision, we can just do `self.get(id)`; so I'd check that first, before iterating over all patches.

radicle-cli/examples/workflow/5-patching-maintainer.md
```diff
@@ -37,6 +37,7 @@ Nothing to show.
 Define power requirements a07ef7743a3 R1 27857ec ahead 2, behind 0
 └─ * opened by did:key:z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk [..]
 └─ * patch id a07ef7743a32a2e902672ea3526d1db6ee08108a
+└─ * revision id 4f15eb5c994edd0bb6be29cb4801aa74308cc628
 
 ```
```
cloudhead, line 43: Make sure to rebase, there's been changes to the formatting of the output.

radicle-cli/src/commands/merge.rs
slack-coder: Indeed, Ill update the changes to account for it

radicle-cli/examples/rad-patch.md
slack-coder: Where do we want to place the revision in the 'patch list' output?

Here's the output as of now with the patch id.

```
 Define power requirements 191a14e520f2eeff7c0e3ee0a5523c5217eecb89 R0 3e674d1 (flux-capacitor-power) ahead 1, behind 0

  ● opened by did:key:z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk (you) 3 months ago                               

```

We need somewhere to include the Revision ID, the revision index is irrelevant now. Here is a suggestion where we use shortened ID's to suppress noise.

```
 Define power requirements 191a14e Rev  0f2eeff 3e674d1 (flux-capacitor-power) ahead 1, behind 0

  ● opened by did:key:z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk (you) 3 months ago      
```

cloudhead: Hmm that's too confusing, I have a better idea but this should be done in a separate PR

radicle/src/cob/patch.rs
cloudhead: > Would this not result in a 'revision not found' when someone tries merging a recently closed patch?

Yes, but you could have an error like "No open patches with revision <id> found".

> Is there overhead for 'revisions().any(..)' when there's only one revision? We'd still need to check against the revision ID.

Yes, because you might end up searching all revisions (linear time) before finding the one; vs. using `get` which is constant time.

radicle-cli/examples/rad-patch.md
cloudhead: image

In #478

radicle/src/cob/patch.rs
slack-coder: I feel like I'm missing something.

> > Would this not result in a 'revision not found' when someone tries merging a recently closed patch?
>
> Yes, but you could have an error like "No open patches with revision {id} found".
>
A more specific message would be helpful to the user.  How costly could the search be?  Would marking the method as 'slow' help?

> > Is there overhead for 'revisions().any(..)' when there's only one revision? We'd still need to check against the revision ID.
>
> Yes, because you might end up searching all revisions (linear time) before finding the one; vs. using `get` which is constant time.
Are other patch's revisions being included in the iteration?  I'm confused about where the overhead is coming from, because the assumption is a patch having one revision would mean one iteration.

cloudhead:
There is one small conflict, I suggest we keep what's on master, ie. showing the revision index, for now, since the revision id is shown in other places now.

radicle-cli/examples/workflow/5-patching-maintainer.md
```diff
@@ -61,3 +61,18 @@ a07ef7743a3 R1 (27857ec) -> R2 (f6484e0)
...
+Merging a07ef7743a3 Ra07ef7743a32a2e902672ea3526d1db6ee08108a (3e674d1) by did:key:z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk into master (f2de534) via fast-forward...
+Running `git merge --ff-only 3e674d1a1df90807e934f9ae5da2591dd6848a33`...
+Updating f2de534..3e674d1
+Fast-forward
+ REQUIREMENTS | 0
+ 1 file changed, 0 insertions(+), 0 deletions(-)
+ create mode 100644 REQUIREMENTS
+✓ Updated master f2de534 -> 3e674d1 via fast-forward
+✓ Patch state updated, use `rad push` to publish
+```
```
cloudhead, line 78: We should test that the merge actually happened :)

slack-coder:
I've reverted to displaying the revision index in place of the revision id after a merge, as done originally.  The PR is done to specification now.

A few things:

 - what is the future of the revision index?  It has problems defining a specific revision due to be concurrent updates.
 - because the revision id is being used (good), its easy for a user to mistakenly give the patch id for `rad merge`, merging the first revision of the patch (bad).  This is due to the patch ID and revision ID being the same for the first revision.
 - does searching only proposed/open patches for the revision help?  It appears the method patches::proposed() will iterate over them anyways.
 
 ```
     /// Get proposed patches.
    pub fn proposed(
        &self,
    ) -> Result<impl Iterator<Item = (PatchId, Patch, clock::Lamport)>, Error> {
        let all = self.all()?;

        Ok(all
            .into_iter()
            .filter_map(|result| result.ok())
            .filter(|(_, p, _)| p.is_proposed()))
    }
 ```

cloudhead:
> what is the future of the revision index? It has problems defining a specific revision due to be concurrent updates. It can happen different revisions will display the same revision number.

I need to think about this more. We may just remove it.

> because the revision id is being used (good), its easy for a user to mistakenly give the patch id for rad merge, merging the first revision of the patch (bad). This is due to the patch ID and revision ID being the same for the first revision.

True -- I think the way to deal with this is generally not merge old revisions, ie. when there is a newer revision, require ` --force` flag to merge the old one.

> does searching only proposed/open patches for the revision help? It appears the method patches::proposed() will iterate over them anyways.

This isn't an optimization, it's preventing the user from merging an archived patch.

radicle/src/cob/patch.rs
cloudhead: I would add ` Patch::revision(id: &RevisionId) -> Option<Revision>` function so we don't have to iterate over all patch revisions.

radicle-cli/src/git.rs
```diff
@@ -44,6 +45,12 @@ impl Rev {
         let object = repo.revparse_single(self.as_str())?;
         Ok(ObjectId::from(object.id()))
     }
+
+    /// Resolve the revision to an [`EntryId`].
+    pub fn resolve_entry(&self, repo: &git2::Repository) -> Result<EntryId, git2::Error> {
+        let object = repo.revparse_single(self.as_str())?;
+        Ok(EntryId::from(object.id()))
+    }
 }
```
cloudhead, line 54: IT should be possible to make `resolve` generic and not need two functions here. If we return `Result<T, _>` where `T: From<git2::Oid>`

radicle/src/cob/patch.rs
```diff
@@ -885,6 +885,27 @@ impl<'a> Patches<'a> {
         Ok(state_groups)
     }
 
+    /// Find the proposed patch containing the given `Revision`.
+    pub fn find_proposed_by_revision(
+        &self,
+        id: &RevisionId,
+    ) -> Result<Option<(PatchId, Patch)>, Error> {
```
cloudhead, line 892: This function could be improved by taking a `state` parameter, it's currently overly specific for an API function.

radicle-cli/src/commands/merge.rs
```diff
@@ -144,10 +134,24 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
-        .map_err(|e| anyhow!("couldn't find patch {} locally: {e}", &options.id.clone()))?;
+        .map_err(|e| anyhow!("couldn't find patch {} locally: {e}", &id))?;
+    let revision_ix = patch
+        .revisions()
+        .position(|(id_, _)| *id_ == revision_id)
+        .expect("patch revision exists");
+    let (revision_id, revision) = patch
+        .revisions()
+        .nth(revision_ix)
+        .expect("patch revision exists");
```
cloudhead, line 154: Please use `unwrap` and document the safety of it, it like in other places

radicle-cli/src/commands/merge.rs
```diff
@@ -21,14 +19,13 @@ pub const HELP: Help = Help {
     usage: r#"
 Usage
 
-    rad merge [<id>] [<option>...]
+    rad merge [<revision id>] [<option>...]
```
cloudhead, line 22:
```suggestion
    rad merge [<revision-id>] [<option>...]
```

radicle-cli/examples/workflow/5-patching-maintainer.md
FintanH: You could do:
```
$ git show-ref refs/heads/master
f6484e0f43e48a8983b9b39bf9bd4cd889f1d520 refs/heads/master
```

radicle-cli/src/commands/merge.rs
```diff
@@ -144,10 +134,27 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+            "revision's patch must be proposed for merging, but it is `{}`",
+            patch.state()
+        );
+    }
+
     let mut patch = patches
         .get_mut(&patch_id)
-        .map_err(|e| anyhow!("couldn't find patch {} locally: {e}", &options.id.clone()))?;
+        .map_err(|e| anyhow!("couldn't find patch {} locally: {e}", &id))?;
+    // The patch is fetched by its revision_id above, and therefor expected to exist.
```
FintanH, line 152:
```suggestion
    // SAFETY: The patch is fetched by its revision_id above and therefore expected to exist.
```

radicle/src/cob/patch.rs
```diff
@@ -885,6 +885,23 @@ impl<'a> Patches<'a> {
         Ok(state_groups)
     }
 
+    /// Find the proposed patch containing the given `Revision`.
+    pub fn find_by_revision(&self, id: &RevisionId) -> Result<Option<(PatchId, Patch)>, Error> {
+        // First Patch `RevisionId` is the same as the `PatchId`.
```
cloudhead, line 890:
```suggestion
        // The first patch revision has the same id as the patch.
```

slack-coder:
> True -- I think the way to deal with this is generally not merge old revisions, ie. when there is a newer revision, require  --force flag to merge the old one.

This has been [added](https://github.com/radicle-dev/heartwood/pull/461/commits/86917c10f4307184aefb7f4828976a0954c4752e)

radicle-cli/src/commands/merge.rs
```diff
@@ -71,51 +69,50 @@ pub enum State {
...
 }
 
 impl Args for Options {
     fn from_args(args: Vec<OsString>) -> anyhow::Result<(Self, Vec<OsString>)> {
         use lexopt::prelude::*;
 
         let mut parser = lexopt::Parser::from_args(args);
-        let mut id: Option<Rev> = None;
-        let mut revision: Option<RevisionIx> = None;
+        let mut force = true;
```
cloudhead, line 85: `force` flags usually default to `false`

radicle/src/cob/patch.rs
cloudhead: Still hoping for this :point_up:

radicle-cli/src/commands/merge.rs
slack-coder: Yup, not good.  Had to rebase to update this with recent changes, and [corrected the above patch](https://github.com/radicle-dev/heartwood/pull/461/commits/1dd4d7f777845deb18c5551b987a68154d5c804f)

slack-coder: :+1: I've rebased + squashed due to merge conflicts against master.  The [valule in question has been fixed](https://github.com/radicle-dev/heartwood/pull/461/commits/8603bfcab4f1b105727338b7374d55ffe8ed6c9d)

radicle/src/cob/patch.rs
slack-coder: [done](https://github.com/radicle-dev/heartwood/pull/461/commits/4bf75d58c0e2e7cd36f66bf9f0553a1fbf405d38) and resolved merge conflicts

radicle/src/cob/patch.rs
```diff
@@ -216,6 +216,15 @@ impl Patch {
...
     }
 
+    /// Get the `Revision` by its `RevisionId`.
+    ///
+    /// None is returned if the `Revision` has been redacted (deleted).
+    pub fn revision(&self, id: &RevisionId) -> Option<&Revision> {
+        self.revisions()
+            .find(|(id_, _)| id_ == &id)
+            .map(|(_, rev)| rev)
+    }
```
cloudhead, line 226: The reason for adding this function was to have O(1) access to any revision; but this implementation just iterates over all revisions instead of just getting the revision from the map.

radicle-cli/src/commands/merge.rs
```diff
@@ -221,6 +230,10 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
             .to_string(),
     };
 
+    let revision_ix = patch
+        .revisions()
+        .position(|(id_, _)| id_ == &revision_id)
+        .unwrap();
```
cloudhead, line 236: Safety comment missing

radicle-cli/src/commands/merge.rs
```diff
@@ -144,10 +141,27 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+    if !options.force && revision_id != *last_revision_id {
+        anyhow::bail!("refusing to merge old patch revision");
+    }
+
     let mut patch = patches
         .get_mut(&patch_id)
-        .map_err(|e| anyhow!("couldn't find patch {} locally: {e}", &options.id.clone()))?;
+        .map_err(|e| anyhow!("couldn't find patch {} locally: {e}", &id))?;
+    // SAFETY: The patch is fetched by its revision_id above, and therefore expected to exist.
+    let revision = patch.revision(&revision_id).unwrap();
```
cloudhead, line 164: Seems like it would make more sense if `find_by_revision` also returned the revision?

radicle-cli/src/commands/merge.rs
```diff
@@ -144,10 +141,27 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+        "no open patch with revision `{}` could be found",
+        &revision_id
+    ))?;
+    if !patch.is_open() {
+        anyhow::bail!(
+            "revision's patch must be open for merging, but it is `{}`",
+            patch.state()
+        );
+    }
+    let (last_revision_id, _) = patch.latest().expect("patch has one or more revisions");
```
cloudhead, line 155: We don't use `expect` that way, please use `unwrap()` plus a "safety" comment, or return the error with `?`. For this case, it's possible for a patch not to have a latest, so it would be better to return an error.

radicle/src/cob/patch.rs
cloudhead: You're still iterating over all patch revisions..

slack-coder: fixed

radicle-cli/src/commands/merge.rs
slack-coder: done

PR #462, opened 2023-03-20
radicle-httpd: Add commit diff endpoint

sebastinez:
Adds a `diff` endpoint to get the diff and commit listing between two commits.
Very useful for patches, but eventually also for other places.

Pending:
- [x] I want to check what happens when multiple commits are in between the merge base and the head commit.

### DIFF
--- a/radicle-httpd/src/api/v1/projects.rs
+++ b/radicle-httpd/src/api/v1/projects.rs
@@ -33,6 +33,7 @@ pub fn router(ctx: Context) -> Router {
         .route("/projects/:project", get(project_handler))
         .route("/projects/:project/commits", get(history_handler))
         .route("/projects/:project/commits/:sha", get(commit_handler))
+        .route("/projects/:project/diff/:base/:oid", get(diff_handler))
         .route(
             "/projects/:project/activity",
             get(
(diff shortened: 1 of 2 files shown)


radicle-httpd/src/api/v1/projects.rs
```diff
@@ -230,6 +238,47 @@ async fn commit_handler(
...
+) -> impl IntoResponse {
+    let DiffQueryString { page, per_page } = qs;
+    let storage = &ctx.profile.storage;
+    let repo = Repository::open(paths::repository(storage, &project))?;
+    let base = repo.commit(base)?;
+    let commit = repo.commit(oid)?;
+    let diff = repo.diff(base.id, commit.id)?;
+
+    // If a pagination is defined, we do not want to paginate the commits, and we return all of them on the first page.
+    let page = page.unwrap_or(0);
```
cloudhead, line 256: It seems weird that we would paginate the commits, but return the full diff. Could we just remove pagination for now? Or do we think we need it?

sebastinez: I wanted to have a similar commit listing as GH, on their PRs and the diff to be the full one, since we would show them in different tabs.
But I think we can allow us to not have commit pagination for now.

cloudhead: Ah is this in the "compare" view? I didn't realize they paginated commits there.

sebastinez: Yes I thought about adding a commit listing like GitHub does under the commit tab
Bildschirm­foto 2023-03-21 um 13 14 32

But I finally found a big PR and it seems they don't paginate..
https://github.com/torvalds/linux/pull/797/commits

So I'll leave the commit pagination out for now

cloudhead: Aha but they also don't show all commits! " This pull request is big! We’re only showing the most recent 250 commits. "

So yeah, let's return all commits and deal with the case of "too many commits" if it comes up.

sebastinez:
Fixed open issues with last commit 1f5b2a3

PR #464, opened 2023-03-20
radicle: Fix profile::canonicalize_home test on MacOS

sebastinez:
On MacOS the following error appears when running the test suite

```
thread 'profile::test::canonicalize_home' panicked at 'assertion failed: `(left == right)`
  left: `"/private/var/folders/r9/dcp4ksl90dnfj2kqg7g7nbd40000gn/T/.tmpuewkuK/Home/Radicle"`,
 right: `"/var/folders/r9/dcp4ksl90dnfj2kqg7g7nbd40000gn/T/.tmpuewkuK/Home/Radicle"`', radicle/src/profile.rs:253:9
 ```

To make the tests pass we should strip the `/private` prefix on MacOS, if it exists.
It shouldn't be an issue since MacOS provides a `/var` symlink to mimic Unix behaviour

### DIFF
--- a/radicle/src/profile.rs
+++ b/radicle/src/profile.rs
@@ -227,6 +227,7 @@ impl Home {
 #[cfg(test)]
 mod test {
     use std::fs;
+    use std::path::Path;
 
     use super::Home;
 
@@ -242,7 +243,7 @@ mod test {
         fs::create_dir_all(path.clone()).unwrap();
 
         let last = tmp.path().components().last().unwrap();
-        let home = Home::new(
+        let mut home = Home::new(
             tmp.path()
                 .join("..")
                 .join(last)
@@ -250,6 +251,11 @@ mod test {
                 .join("Radicle"),
         )
         .unwrap();
+        if cfg!(target_os = "macos") {
+            home.path =
+                Path::new("/").join(home.path.strip_prefix("/private").unwrap_or(&home.path))
+        };
+
         assert_eq!(home.path, path);
     }
 }

### CODE REVIEW

File: radicle/src/profile.rs [hunk 1]
```diff
@@ -250,6 +250,13 @@ mod test {
                 .join("Radicle"),
         )
         .unwrap();
-        assert_eq!(home.path, path);
+        // Handle MacOS test case with '/private' tmp.path prefix.
+        let home_path = if let Ok(path) = home.path.strip_prefix("/private") {
+            ("/".to_owned() + path.to_str().unwrap()).into()
+        } else {
+            home.path
+        };
```
FintanH, line 258: I'm not 100% sure if this code will work but we could definitely use parts of it if it doesn't.
I added a check for the OS being `macos` since it only affects that OS, so it's a good signal for testing only on that platform.
Secondly, I opted to use `Path::new("/").join(path)` instead of string manipulation.
Last, I think if we could mark `home` as `mut` then perhaps we can just assign the new `home.path`. If this doesn't work we can use the new `home_path` variable and have an `#[cfg(not(target_os = "macos"))]` branch of execution and do `home_path = home.path`.
```suggestion
        #[cfg(target_os = "macos")]
        // Handle MacOS test case with '/private' tmp.path prefix.
        if let Ok(path) = home.path.strip_prefix("/private") {
            home.path = Path::new("/").join(path);
        };
```

sebastinez: Thanks @FintanH!

I also tried a few of your comments, like mutate `home` and add platform specific branch execution and mutate the variable inside of a platform specific tag didn't worked yesterday, but today it seems like the compiler says yes 😅

FintanH: Unfortunately, the linter says no :joy: I think your original proposal is fine with the `if ... else ...` but let's add the `join`.

sebastinez: hmm it seems so, ok will revert so the linter says yes

radicle/src/profile.rs
```diff
@@ -250,6 +251,13 @@ mod test {
...
         )
         .unwrap();
-        assert_eq!(home.path, path);
+        // Handle MacOS test case with '/private' tmp.path prefix.
+        let home_path = if let Ok(path) = home.path.strip_prefix("/private") {
+            Path::new("/").join(path)
+        } else {
+            home.path
+        };
+
```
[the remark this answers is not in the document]
sebastinez: Ok targetted `macos` again with b764f28, and converted `home` back to mut, since I think it seems nicer.

radicle/src/profile.rs
```diff
@@ -242,14 +243,18 @@ mod test {
...
+        let mut home = Home::new(
             tmp.path()
                 .join("..")
                 .join(last)
                 .join("Home")
                 .join("Radicle"),
         )
         .unwrap();
+        if cfg!(target_os = "macos") {
+            home.path = Path::new("/").join(home.path.strip_prefix("/private").unwrap())
```
FintanH, line 255: I'd say use `unrwap_or` just in case the `/private` doesn't show up
```suggestion
            home.path = Path::new("/").join(home.path.strip_prefix("/private").unwrap_or(home.path))
```

FintanH: Note that the join won't add a second `/` https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fa1c7b98ace3cf2f2174df02472163b8

sebastinez: > I'd say use `unrwap_or` just in case the `/private` doesn't show up
> 
> ```suggestion
> 
>             home.path = Path::new("/").join(home.path.strip_prefix("/private").unwrap_or(home.path))
> 
> ```

Good point 👍

sebastinez: > Note that the join won't add a second `/` https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fa1c7b98ace3cf2f2174df02472163b8

Yeah the join is due to `strip_prefix` removing the first `/` so we add it back
see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ad6cb1f45ca68d20f88fd5ef0f365748

sebastinez: > Sure, what I meant was that if the path already starts with `/` in the case of using `unwrap_or` we're not adding an extra one :)

Ah nice good to know

PR #469, opened 2023-03-22
Skip hidden files in repository listing (`rad ls`)

cloudhead:
*See commits*

### DIFF
--- a/radicle/src/storage.rs
+++ b/radicle/src/storage.rs
@@ -20,7 +20,7 @@ use crate::git::{Qualified, RefError, RefString};
 use crate::identity;
 use crate::identity::doc::DocError;
 use crate::identity::Did;
-use crate::identity::{Id, IdError, IdentityError};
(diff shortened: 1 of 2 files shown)


slack-coder:
Code looks good.

Are hidden files and non-directory files expected in this directory?  Otherwise if neither are expected, why would an unexpected directory (one with a bad ID) cause a failure, while a file does not.

FintanH:
> Code looks good.
> 
> Are hidden files and non-directory files expected in this directory? Otherwise if neither are expected, why would an unexpected directory (one with a bad ID) cause a failure, while a file does not.

It showed up on a Mac machine https://radicle.zulipchat.com/#narrow/stream/374620-cli/topic/unknown.20base.20code.20.2E :upside_down_face:

FintanH (approved):

sebastinez (approved):
Code works for me 👍

PR #471, opened 2023-03-22
httpd: Add `content` property to `CodeLocation` in serialization

sebastinez:
For the web client to be able to show file content based on a Oid we need to provide the web client with the file in question.

~~It gets a bit messy IMO, since we need to drill down until we get to the `CodeLocation` to be able to serialize the Blob Oid, into its content, adding a new field..~~

~~I think this would be nicer if instead or additionally to passing a Blob Id we store the file path in the `CodeLocation` struct, and then query the `/raw` endpoint for the files content.~~

### DIFF
--- a/radicle-httpd/src/api/v1/projects.rs
+++ b/radicle-httpd/src/api/v1/projects.rs
@@ -2089,7 +2089,8 @@ mod routes {
           "inline": [
             {
               "location": {
-                "blob": HEAD,
+                "blob": "82eb77880c693655bce074e3dbbd9fa711dc018b",
+                "path": "./README.md",
                 "commit": HEAD,
                 "lines": {
                     "start": 1,
@@ -2148,7 +2149,8 @@ mod routes {
                         "inline": [
                           {
                             "location": {
-                              "blob": HEAD,
+                              "blob": "82eb77880c693655bce074e3dbbd9fa711dc018b",
+                              "path": "./README.md",
                               "commit": HEAD,
                               "lines": {
                                 "start": 1,
--- a/radicle/src/cob/patch.rs
+++ b/radicle/src/cob/patch.rs
@@ -2,6 +2,7 @@
 use std::fmt;
 use std::ops::Deref;
 use std::ops::Range;
+use std::path::PathBuf;
 use std::str::FromStr;
 
 use once_cell::sync::Lazy;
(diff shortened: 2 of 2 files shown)


radicle-httpd/src/api/v1/projects.rs
```diff
@@ -2117,6 +2117,26 @@ mod routes {
             format!("/projects/{CONTRIBUTOR_RID}/patches/{CONTRIBUTOR_PATCH_ID}"),
         )
         .await;
+        // Separated this serialization due to hitting the recursion limit
```
cloudhead, line 2120: lol

cloudhead:
If we could somehow have the file path inside the COB state, but *not* in the COB actions, that would be interesting.. Otherwise it's redundant, but redundancy is maybe not the worse..

radicle/src/cob/patch.rs
```diff
@@ -464,8 +464,8 @@ impl fmt::Display for Verdict {
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CodeLocation {
-    /// File being commented on.
-    pub blob: git::Oid,
+    /// Path of file being commented on.
+    pub path: String,
```
cloudhead, line 468:
```suggestion
    pub path: PathBuf,
```

radicle/src/cob/patch.rs
FintanH, line 1: I was baffled that these are the only changes, but I suppose we don't have any code that uses the `Review` code path yet :sweat_smile:

radicle/src/cob/patch.rs
cloudhead: I can't think of a reason why we'd need the blob-id if we have commit+path, right @FintanH ?

cloudhead: It should be equivalent, since `commit -> tree -> path -> blob`

FintanH: Indeed! It might be worth keeping in the case of improving `surf` down the line though. Much quicker to find the blob by its id over looking up via a commit's tree.

cloudhead: Right. Let's keep both then I suggest.

sebastinez:
Ok did address the comments and rebased/squashed it

PR #473, opened 2023-03-22
node: Fix periodic inventory announcement

cloudhead:
1. We were announcing inventory every 30 seconds. This is now changed to 1 hour.
2. We were using a variable to track whether our inventory has been announced to the network, but this var was set based on what we tracked, not the inventory we had. This var has been removed.


radicle-node/src/service.rs
```diff
@@ -1148,7 +1143,7 @@ where
         )
     }
 
-    /// Sync, and if needed, announce our local inventory.
+    /// Update our routing table based on our inventory.
     fn sync_inventory(&mut self) -> Result<Vec<Id>, Error> {
```
slack-coder, line 1147: Suggestion: "Update our routing table with this node's inventory"

slack-coder:
Otherise :+1:

PR #474, opened 2023-03-23
cli: Enable `rad tag` and `rad untag` for patches as well

sebastinez:
Whats on the tin.

Instead of just allowing issues to be tagged and untagged, we should also allow patches to be tagged.


cloudhead (approved):
LGTM

Issue #450, opened 2023-03-16
Deleting local storage is non-recoverable


cloudhead:
If a user deletes `.radicle/storage` for whatever reason, their `rad/sigrefs` branch gets deleted.
Then, if they try to recover the repo by cloning from a seed, their namespace will be omitted.
Then, if they do anything, eg. create a commit and try to push, their sigrefs will be rejected by the remote, since the remote sigrefs is divergent.

PR #476, opened 2023-03-23
node: multistep fetching logic

### LINKED ISSUE #14: Multi-step git fetch
Currently, we fetch all refs in one go and then verify them.

We should use a multi-step process where:
1. Identity branches for all remotes are fetched and verified
2. Signed refs are fetched and verified against the identity branches
3. All other refs are fetched and verified against the signed refs

### LINKED ISSUE #318: On clone, only fetch delegate refs
Currently, when calling `rad clone`, we fetch all refs (`Namespaces::All`) of all remotes. This is not optimal, and instead we should only fetch the delegates at first. Later if the user tracks additional peers, he can fetch the additional refs on demand, or perhaps having an `--all-remotes` option to `rad-clone` could force that behavior. The default though should be delegates-only.

This is a little tricky because we don't know who the delegates are, apriori. It therefore requires the identity branches of *all* (?) remotes to be fetched first, so that the delegates may be discovered, and then the source code refs can be fetched only for those delegates.

### LINKED ISSUE #397: Using Force for Fetch Refspecs

FintanH:
The previous fetching logic had some pitfalls regarding its logic:

* Fetched 'rad' references could end up not being verifiable and leaving the storage in a corrupt state
* Forced pushes would be prohibited, since they would be rejected by the client side due to the refspecs being used.

The new approach is to stage a fetch in 2 steps. The first step is to fetch the relevant `rad/id` and `rad/sigrefs` for verification. For all remotes that are verified, fetch all references listed in their respective `rad/sigrefs`, using the `+` marker in the refspec, allowing to update for any force pushes -- note that this becomes a safe operation since the `sigrefs` are signed by the remote that created them.

To prevent unverifiable 'rad' references from polluting storage, all fetch negotiations are done in a staging repository set up in a temporary directory. All successful fetches are then transferred, using the 'file://' protocol, from the temporary directory into the radicle storage.

Depends on https://github.com/radicle-dev/heartwood/pull/488
Fixes #14
Fixes #318 
Fixes #397 
Fixed https://github.com/radicle-dev/heartwood/issues/450

### DIFF
--- a/radicle-node/src/tests/e2e.rs
+++ b/radicle-node/src/tests/e2e.rs
@@ -196,6 +196,56 @@ fn test_replication() {
 }
 
 #[test]
+fn test_migrated_clone() {
+    logger::init(log::Level::Debug);
+
+    let tmp = tempfile::tempdir().unwrap();
+    let mut alice = Node::init(tmp.path());
+    let bob = Node::init(tmp.path());
+    let acme = alice.project("acme", "");
+
+    let mut alice = alice.spawn(service::Config::default());
+    let mut bob = bob.spawn(service::Config::default());
+
+    alice.connect(&bob);
+    converge([&alice, &bob]);
+
+    let tracked = bob.handle.track_repo(acme, Scope::All).unwrap();
+    assert!(tracked);
+
+    let result = bob.handle.fetch(acme, alice.id).unwrap();
+    assert!(result.is_success());
+
+    log::debug!(target: "test", "Fetch complete with {}", alice.id);
+
+    // Simulate alice deleting the project and cloning it again
+    {
+        let path = alice.storage.path().join(acme.canonical());
+        std::fs::remove_dir_all(path).unwrap();
+    }
+    assert!(!alice.storage.contains(&acme).unwrap());
+    let result = alice.handle.fetch(acme, bob.id).unwrap();
+    assert!(result.is_success());
+
+    let alice_repo = alice.storage.repository(acme).unwrap();
+    let bob_repo = bob.storage.repository(acme).unwrap();
+
+    let alice_refs = alice_repo
+        .references()
+        .unwrap()
+        .collect::<Result<Vec<_>, _>>()
+        .unwrap();
+    let bob_refs = bob_repo
+        .references()
+        .unwrap()
+        .collect::<Result<Vec<_>, _>>()
+        .unwrap();
+
+    assert_eq!(alice_refs, bob_refs);
+    assert_matches!(alice.storage.repository(acme).unwrap().verify(), Ok(()));
+}
+
+#[test]
 fn test_dont_fetch_owned_refs() {
     logger::init(log::Level::Debug);
 
--- a/radicle-node/src/tests/e2e.rs
+++ b/radicle-node/src/tests/e2e.rs
@@ -223,7 +223,6 @@ fn test_dont_fetch_owned_refs() {
 }
 
 #[test]
-#[ignore = "failing"]
 fn test_fetch_trusted_remotes() {
     logger::init(log::Level::Debug);
 
@@ -272,8 +271,22 @@ fn test_fetch_trusted_remotes() {
         .unwrap()
         .collect::<Result<HashSet<_>, _>>()
         .unwrap();
-    // TODO: This fails because we are fetching all namespaces at the moment.
-    assert_eq!(bob_remotes.len(), trusted.len() + 1);
+    assert_eq!(bob_remotes, Some(alice.id).into_iter().collect());
+
+    // TODO(finto): we have to fetch again to get the other trusted remotes.
+    // At the moment, the existing Namespaces enum does not allow us
+    // to pass on what nodes are tracked, if there is no existing
+    // repository. Thus, the first fetch only attempts to clone the
+    // delegate.
+    bob.handle.fetch(acme, alice.id).unwrap();
+    assert!(result.is_success());
+    let bob_remotes = bob_repo
+        .remote_ids()
+        .unwrap()
+        .collect::<Result<HashSet<_>, _>>()
+        .unwrap();
+
+    assert!(bob_remotes.len() == trusted.len() + 1);
     assert!(bob_remotes.is_superset(&trusted));
     assert!(bob_remotes.contains(&alice.id));
 }
--- a/radicle-node/src/tests/e2e.rs
+++ b/radicle-node/src/tests/e2e.rs
@@ -171,7 +171,7 @@ fn test_replication() {
             panic!("Fetch failed from {}: {reason}", bob.id);
         }
     };
-    assert_eq!(*updated, vec![]);
+    assert!(!updated.is_empty());
 
     log::debug!(target: "test", "Fetch complete with {}", bob.id);
 
--- a/radicle-node/src/worker.rs
+++ b/radicle-node/src/worker.rs
@@ -1,3 +1,5 @@
+mod fetch;
+
 use std::io::{prelude::*, BufReader};
 use std::ops::ControlFlow;
 use std::thread::JoinHandle;
@@ -8,9 +10,9 @@ use cyphernet::Ecdh;
 use netservices::tunnel::Tunnel;
 use netservices::{AsConnection, NetSession, SplitIo};
 
-use radicle::crypto::{PublicKey, Signer};
+use radicle::crypto::Signer;
 use radicle::identity::{Id, IdentityError};
-use radicle::storage::{Namespaces, ReadRepository, RefUpdate, WriteRepository, WriteStorage};
+use radicle::storage::{Namespaces, ReadRepository, RefUpdate};
 use radicle::{git, Storage};
 use reactor::poller::popol;
 
@@ -54,6 +56,12 @@ pub enum FetchError {
     Upload(#[from] UploadError),
     #[error("remote aborted fetch")]
     RemoteAbortedFetch,
+    #[error(transparent)]
+    StagingInit(#[from] fetch::error::Init),
+    #[error(transparent)]
+    StagingTransition(#[from] fetch::error::Transition),
+    #[error(transparent)]
+    StagingTransfer(#[from] fetch::error::Transfer),
 }
 
 impl FetchError {
@@ -100,7 +108,6 @@ pub struct TaskResult<G: Signer + Ecdh> {
 
 /// A worker that replicates git objects.
 struct Worker<G: Signer + Ecdh> {
-    local: PublicKey,
     storage: Storage,
     tasks: chan::Receiver<Task<G>>,
     daemon: net::SocketAddr,
@@ -153,27 +160,15 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
         drain: Vec<u8>,
         mut session: WireSession<G>,
     ) -> (WireSession<G>, Result<Vec<RefUpdate>, FetchError>) {
-        let rid = fetch.rid;
         match &fetch.direction {
             FetchDirection::Initiator { namespaces } => {
                 log::debug!(target: "worker", "Worker processing outgoing fetch for {}", fetch.rid);
 
-                let mut tunnel =
-                    match Tunnel::with(session, net::SocketAddr::from(([0, 0, 0, 0], 0))) {
-                        Ok(tunnel) => tunnel,
-                        Err((session, err)) => return (session, Err(err.into())),
-                    };
-                let result = self.fetch(rid, namespaces, &mut tunnel);
-                let mut session = tunnel.into_session();
-
+                let (session, result) = self.fetch(fetch.rid, namespaces, session);
                 if let Err(err) = &result {
                     log::error!(target: "worker", "Fetch error: {err}");
                 }
-                log::debug!(target: "worker", "Sending `done` packet to remote..");
 
-                if let Err(err) = pktline::done(&mut session) {
-                    log::error!(target: "worker", "Fetch error: error sending `done` packet: {err}");
-                }
                 (session, result)
             }
             FetchDirection::Responder => {
@@ -190,70 +185,13 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
                     }
                 };
                 let mut pktline_r = pktline::Reader::new(drain, &mut stream_r);
-
-                match self.upload_pack(fetch, &mut pktline_r, &mut stream_w) {
-                    Ok(()) => {
-                        log::debug!(target: "worker", "Upload of {} to {} exited successfully", fetch.rid, fetch.remote);
-
-                        (WireSession::from_split_io(stream_r, stream_w), Ok(vec![]))
-                    }
-                    Err(err) => {
-                        log::error!(target: "worker", "Upload error for {rid}: {err}");
-
-                        // If we exited without receiving a `done` packet, wait for it here.
-                        // It's possible that the daemon exited first, or the remote crashed.
-                        log::debug!(target: "worker", "Waiting for `done` packet from remote..");
-                        let mut header = [0; pktline::HEADER_LEN];
-
-                        // Set the read timeout for the `done` packet to twice the configured
-                        // value that is used for the fetching (initiator) side.
-                        //
-                        // This is because the uploader always waits for the `done` packet;
-                        // so in case the fetch is aborted by the uploader, eg. if
-                        // it can't connect with the daemon, it will wait long enough for the
-                        // fetcher to timeout before timing out itself, and will thus receive
-                        // the `done` packet.
-                        pktline_r
-                            .stream()
-                            .as_connection()
-                            .set_read_timeout(Some(self.timeout * 2))
-                            .ok();
-
-                        loop {
-                            match pktline_r.read_done_pktline(&mut header) {
-                                Ok(()) => {
-                                    log::debug!(target: "worker", "Received `done` packet from remote");
-
-                                    // If we get the `done` packet, we exit with the original
-                                    // error.
-                                    return (
-                                        WireSession::from_split_io(stream_r, stream_w),
-                                        Err(err.into()),
-                                    );
-                                }
-                                Err(e) if e.kind() == io::ErrorKind::InvalidInput => {
-                                    // If we get some other packet, because the fetch request
-                                    // is still sending stuff, we simply keep reading until we
-                                    // get a `done` packet.
-                                    continue;
-                                }
-                                Err(_) => {
-                                    // If we get any other error, eg. a timeout, we abort.
-                                    log::error!(
-                                        target: "worker",
-                                        "Upload of {} to {} aborted: missing `done` packet from remote",
-                                        fetch.rid,
-                                        fetch.remote
-                                    );
-                                    return (
-                                        WireSession::from_split_io(stream_r, stream_w),
-                                        Err(FetchError::RemoteAbortedFetch),
-                                    );
-                                }
-                            }
-                        }
-                    }
-                }
+                // Nb. two fetches are expected to happen, one for
+                // the `rad` refs, followed by the refs listed in
+                // signed refs.
+                let result = self.upload_pack(fetch, &mut pktline_r, &mut stream_w);
+                let result =
+                    result.and_then(|_| self.upload_pack(fetch, &mut pktline_r, &mut stream_w));
+                (WireSession::from_split_io(stream_r, stream_w), result)
             }
         }
     }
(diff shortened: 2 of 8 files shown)


radicle-node/src/worker.rs
```diff
@@ -100,7 +104,6 @@ pub struct TaskResult<G: Signer + Ecdh> {
 
 /// A worker that replicates git objects.
 struct Worker<G: Signer + Ecdh> {
-    local: PublicKey,
```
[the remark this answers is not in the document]
cloudhead: * If we force-fetch `rad/id`, then yes, the refs won't be rejected.
* If we don't use `--atomic`, a rejected ref won't fail the fetch

These are the two reasons I can think of, but I'm not sure they are good reasons, ie. I think we may not want to force `rad/id` and we may want to use `--atomic`.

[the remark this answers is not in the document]
FintanH: `--atomic` only exists using the `git` command line though iirc, but in our final step we're using `git2` for local transfer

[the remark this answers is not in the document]
cloudhead: True yes, it only applies to the fetch into the staging area.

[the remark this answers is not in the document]
FintanH: Right, so this is a good reason _not_ to include `rad/id` in the sigrefs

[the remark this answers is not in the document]
cloudhead: Oh interesting :thinking: -- this is something we discussed on zulip right? And then thought we'd leave it as it "doesn't hurt".

The only thing is, I think parts of the code expect that *all* refs are signed in sigrefs. It makes verification simpler, with two rules: (1) all refs in sigrefs must exist, and (2) all refs that exist must be in sigrefs.

Removing rad/id would add an annoying exception to that rule.

Another option is to leave rad-id in there, but special case it to not allow force-push.

But this makes me wonder: is there any harm in force-pushing rad/id? I guess we'd have to change the "longest branch" logic.. :thinking:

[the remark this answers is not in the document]
FintanH: Hmmmm I guess it can't go wrong because we've fetched `rad/id` anyway, so it should essentially be a no-op because it would be hard (and useless) to have a different `sigrefs` entry.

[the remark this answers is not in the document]
cloudhead: So are you proposing to leave it in sigrefs because that'll fail on verification if it is force-pushed?

radicle-node/src/worker.rs
```diff
@@ -403,6 +291,155 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
             }
         }
     }
+
+    fn _upload_pack(
+        &self,
+        fetch: &Fetch,
+        pktline_r: &mut pktline::Reader<WireReader>,
+        stream_w: &mut WireWriter<G>,
+    ) -> Result<Vec<RefUpdate>, FetchError> {
```
slack-coder, line 300: What do methods with preceding '_' signify?  I notice process(..) calls _process(..) but _upload_pack(..) calls upload_pack(..).

radicle-node/src/worker.rs
```diff
@@ -190,153 +189,42 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
...
+            self.tunnel_fetch(&staging.repo, staging.refspecs(namespaces.clone()), session);
+
+        let staging = match res.and_then(|()| {
+            staging
+                .into_staging_refs(namespaces.clone())
+                .map_err(FetchError::from)
+        }) {
+            Ok(staging) => staging,
+            Err(e) => return (session, Err(e)),
+        };
```
slack-coder, line 219: This -might- be easier to grok if the and_then was split into a second `let .. match(.)`.  It looks like both would be returning on error.

FintanH: Ya, I'm avoiding having to repeat matching on the errors. Is it's current form confusing?

radicle-node/src/worker.rs
FintanH: Thanks for catching this. I've updated the code by swapping them around :)

FintanH:
[Test failure](https://github.com/radicle-dev/heartwood/actions/runs/4541762111/jobs/8004345038?pr=476) might be a flake because it's not happening locally for me :thinking:

radicle-node/src/worker/fetch.rs
```diff
@@ -0,0 +1,333 @@
...
+    pub fn refspecs(
+        &self,
+        namespaces: Namespaces,
+    ) -> Vec<Refspec<git::PatternString, git::PatternString>> {
+        let id = git::PatternString::from(IDENTITY_BRANCH.clone().into_refstring());
+        match self.repo {
+            StagedRepository::Cloning(_) => Refspec {
+                src: id.clone(),
+                dst: id,
+                force: true,
```
slack-coder, line 120: Why would `force` be true if this is cloning?  My belief being there's nothing to force write over, since something is being cloned.

FintanH: Ya, you're right. If `force: false` fails then something went wrong so it should be set to `false` in the cloning case :+1:

slack-coder:
Unrelated to this PR.

What do you think about renaming worker to something job specific?  As it stands, it focused git cloning, 'worker' is too generic.

radicle-node/src/worker.rs
```diff
@@ -153,27 +160,19 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
...
         mut session: WireSession<G>,
     ) -> (WireSession<G>, Result<Vec<RefUpdate>, FetchError>) {
-        let rid = fetch.rid;
         match &fetch.direction {
             FetchDirection::Initiator { namespaces } => {
                 log::debug!(target: "worker", "Worker processing outgoing fetch for {}", fetch.rid);
+                let staging = match fetch::StagingRad::new(&self.storage, fetch.rid) {
+                    Ok(staging) => staging,
+                    Err(err) => return (session, Err(err.into())),
+                };
```
slack-coder, line 169: How would you feel about moving this into the 'fetch(..)'?  It would reduce the fetch() arguments to commonly types.  StagingRad and StagingRefs appear to be stages of the fetch action.

radicle-node/src/worker.rs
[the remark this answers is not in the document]
cloudhead: Ah I see, so we fetch without force, and we keep rad/id inside sigrefs.

cloudhead:
> What do you think about renaming worker to something job specific? As it stands, it focused git cloning, 'worker' is too generic.

An alternative is to move it to a submodule: `node::git::worker`. Otherwise I'm not sure what I'd call it.

slack-coder:
> > What do you think about renaming worker to something job specific? As it stands, it focused git cloning, 'worker' is too generic.
> 
> An alternative is to move it to a submodule: `node::git::worker`. Otherwise I'm not sure what I'd call it.

That could work

radicle-node/src/tests/e2e.rs
```diff
@@ -272,8 +271,20 @@ fn test_fetch_trusted_remotes() {
         .unwrap()
         .collect::<Result<HashSet<_>, _>>()
         .unwrap();
-    // TODO: This fails because we are fetching all namespaces at the moment.
-    assert_eq!(bob_remotes.len(), trusted.len() + 1);
+    assert_eq!(bob_remotes, Some(alice.id).into_iter().collect());
+
+    // TODO(finto): we have to fetch again to get the other trusted remotes
+    // We need to be able to thread through tracking in the fetch to
+    // look up the trusted peers -- or possibly some other method.
```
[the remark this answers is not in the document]
slack-coder: Ok cool I'll take a look at the PR after this one is merged.  The comments confusing because of the 'thread through term', and where the 'trusted remotes' information is coming from.  As it stands its a little confusing.

radicle-node/src/worker.rs
```diff
@@ -262,84 +200,96 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
...
+                // This is because the uploader always waits for the `done` packet;
+                // so in case the fetch is aborted by the uploader, eg. if
+                // it can't connect with the daemon, it will wait long enough for the
+                // fetcher to timeout before timing out itself, and will thus receive
+                // the `done` packet.
+                pktline_r
+                    .stream()
+                    .as_connection()
+                    .set_read_timeout(Some(self.timeout * 2))
+                    .ok();
```
slack-coder, line 259: What is the '.ok()' for?

radicle-node/src/tests/e2e.rs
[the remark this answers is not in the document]
FintanH: Sorry, I'll write a better comment! :) It was likely just a stream of consciousness as I thought about how to solve the problem

radicle-node/src/worker.rs
FintanH: That's pretty common for channel communications, I believe. If the channel is gone away then reading will throw an error anyway and there's nothing we can do about it anyway -- so this `Err` is silence, so to say.

radicle-node/src/tests/e2e.rs
```diff
@@ -272,8 +321,22 @@ fn test_fetch_trusted_remotes() {
...
         .unwrap();
-    // TODO: This fails because we are fetching all namespaces at the moment.
-    assert_eq!(bob_remotes.len(), trusted.len() + 1);
+    assert_eq!(bob_remotes, Some(alice.id).into_iter().collect());
+
+    // TODO(finto): we have to fetch again to get the other trusted remotes.
+    // At the momemnt, the existing Namespaces enum does not allow us
+    // to pass on what nodes are tracked, if there is no existing
+    // repository. Thus, the first fetch only attempts to clone the
+    // delegate.
```
cloudhead, line 330: Interesting, so for clones, we are missing logic for fetching delegates + trusted in a single fetch.

radicle-node/src/worker.rs
```diff
@@ -403,6 +355,90 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
...
+    ) -> (WireSession<G>, Result<(), FetchError>)
+    where
+        Specs: fetch::AsRefspecs,
+    {
+        let mut tunnel = match Tunnel::with(session, net::SocketAddr::from(([0, 0, 0, 0], 0))) {
+            Ok(tunnel) => tunnel,
+            Err((session, err)) => return (session, Err(err.into())),
+        };
+        let res = self._fetch(repo, specs, &mut tunnel);
+        let mut session = tunnel.into_session();
```
cloudhead, line 373:
```suggestion

        let mut session = tunnel.into_session();
```
Just some breathing room to logicallty break up the function

radicle-node/src/worker.rs
```diff
@@ -403,6 +355,90 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
...
+        log::debug!(target: "worker", "Sending `done` packet to remote..");
+        if let Err(err) = pktline::done(&mut session) {
+            log::error!(target: "worker", "Fetch error: error sending `done` packet: {err}");
+        }
+        (session, res)
+    }
+
+    fn _fetch<Specs>(
```
cloudhead, line 381:
```suggestion
    fn _fetch<S>(
```
I tend to prefer this because when reading the function parameters it makes it clearer what is a generic vs. concrete type.

radicle-node/src/worker/fetch.rs
```diff
@@ -0,0 +1,330 @@
...
+use radicle::storage::{ReadRepository, ReadStorage, WriteRepository, WriteStorage};
+use radicle::{git, storage, Storage};
+
+/// The initial staging area to perform a fetch from a remote.
+///
+/// The [`StagingRad::refpsecs`] generated are to fetch the `rad/id`
+/// and/or `rad/sigrefs` references from the remote end.
+///
+/// It is then expected to convert this into [`StagingRefs`] using
+/// [`StagingRad::into_staging_refs`] to continue the the rest of the
```
cloudhead, line 28:
```suggestion
/// [`StagingRad::into_staging_refs`] to continue the rest of the
```

radicle-node/src/worker/fetch.rs
```diff
@@ -0,0 +1,330 @@
...
+impl<'a> StagingRad<'a> {
+    /// Construct a [`StagingRad`] which sets up its
+    /// [`StagedRepository`] in a new, temporary directory.
+    pub fn new(
+        production: &'a Storage,
+        rid: Id,
+        namespaces: Namespaces,
+    ) -> Result<Self, error::Create> {
+        let tmp = tempfile::TempDir::new()?;
+        log::debug!(target: "fetch", "Staging fetch in {:?}", tmp.path());
```
cloudhead, line 104:
```suggestion
        log::debug!(target: "worker", "Staging fetch in {:?}", tmp.path());
```
I'd rather stick to fewer targets or it gets confusing to know where logs are coming from

radicle-node/src/worker/fetch.rs
```diff
@@ -0,0 +1,330 @@
...
+    pub(super) repo: StagedRepository,
+    /// The staging [`Storage`] for staging the `repo`.
+    staging: Storage,
+    /// The original [`Storage`] we are finalising changes into.
+    production: &'a Storage,
+    /// The remotes that the fetch is being performed for. These are
+    /// discovered after performing the fetch for [`StagingRad`].
+    remotes: NonEmpty<RemoteId>,
+    _tmp: tempfile::TempDir,
+}
```
cloudhead, line 82: It seems strange to me that these are two types. I would expect either an `enum` with two variants, or a struct with an `enum` inside of it :thinking: 

Having the two types makes it a bit too "open ended" as to the relationships between them, and how they are to be used.

radicle-node/src/tests/e2e.rs
FintanH: Aye, that's why I created this issue https://github.com/radicle-dev/heartwood/issues/494 and followed up with PR https://github.com/radicle-dev/heartwood/pull/495.

I wasn't sure how complex the follow-up would be but it turned out simple enough.

radicle-node/src/worker.rs
```diff
@@ -190,70 +185,13 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
...
-                                        Err(FetchError::RemoteAbortedFetch),
-                                    );
-                                }
-                            }
-                        }
-                    }
-                }
+                // XXX: two fetches are expected to happen, one for
+                // the `rad` refs, followed by the refs listed in
+                // signed refs.
```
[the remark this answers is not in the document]
FintanH: `XXX` is basically an `Nb.` or a heads up. So it's like, "Hey, be careful if you're changing this because of the expected two fetched".

Not necessarily a TODO, but possibly something that can be improved in the future :)

radicle-node/src/worker.rs
FintanH: Sure :+1:

radicle-node/src/worker/fetch.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    fn deref(&self) -> &Self::Target {
+        match self {
+            StagedRepository::Cloning(repo) => repo,
+            StagedRepository::Fetching(repo) => repo,
+        }
+    }
+}
+
+/// The second staging area to perform a fetch from a remote.
```
cloudhead, line 60: Area these actually different areas? I thought the refs shared the same staging area (repo)

radicle-node/src/worker/fetch.rs
FintanH: Interesting. I don't think an `enum` would make much sense because then, for example, when we call `transfer` and check that it's in the `rad` fetching phase we have to error.

I thought this was clearer because it splits out the steps into two separate structs -- granted they share a majority of fields. The API _should_ be clear in that `StagingRefs` has no constructor, other than via `StagingRad`, and `StagingRad` is constructed via `new`, and `StagingRefs` has a consuming method `transfer` -- so the path is `StagingRad::new -> StagingRad::into_staging_refs -> StagingRefs::transfer`.

radicle-node/src/worker/fetch.rs
FintanH: You're right :) It's the second _phase_ of fetching

radicle-node/src/worker.rs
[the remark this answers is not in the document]
cloudhead: Oh, then please change it to `Nb.`, because that's what we use in the codebase :)

radicle-node/src/worker/fetch.rs
```diff
@@ -0,0 +1,330 @@
...
+/// an existing repository.
+///
+/// It is then expected to finalise the process by transferring the
+/// fetched references into the production storage, via
+/// [`StagingRefs::transfer`].
+pub struct StagingRefs<'a> {
+    /// The inner [`Repository`] for staging fetches into.
+    pub(super) repo: StagedRepository,
+    /// The staging [`Storage`] for staging the `repo`.
+    staging: Storage,
```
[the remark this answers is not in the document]
FintanH: Good shout :+1:

FintanH:
Thanks for the feedback! Should be ready for another round of :eyes:

radicle-node/src/worker/fetch.rs
```diff
@@ -0,0 +1,322 @@
...
+            Ok(true) => {
+                let url = url::File::new(production.path_of(&rid)).to_string();
+                log::debug!(target: "worker", "Setting up fetch for existing repository: {}", url);
+                let to = storage::git::paths::repository(&staging, &rid);
+                let copy = git::raw::build::RepoBuilder::new()
+                    .bare(true)
+                    .clone_local(git::raw::build::CloneLocal::Local)
+                    .clone(&url, &to)?;
+                // Nb.: need to clean up origin created during clone
+                copy.remote_delete("origin")?;
```
cloudhead, line 175: Any harm in leaving it? It is annoying that it gets created :thinking: -- but if it has no consequences, I wonder if we can leave it because it seems random to delete a remote here that we didn't manually create.

radicle-node/src/worker/fetch/refspecs.rs
```diff
@@ -0,0 +1,166 @@
...
+use std::fmt::Write as _;
+
+use radicle::crypto::PublicKey;
+use radicle::git;
+use radicle::git::refs::storage::{IDENTITY_BRANCH, SIGREFS_BRANCH};
+use radicle::storage;
+use radicle::storage::git::NAMESPACES_GLOB;
+use radicle::storage::{Namespaces, Remote};
+
+// TODO(finto): this should go into radicle-git-ext/git-ref-format
```
cloudhead, line 11: Maybe open an issue?Seems like something pretty straightforward to fix

radicle-node/src/tests/e2e.rs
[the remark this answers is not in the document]
slack-coder: All good :)

radicle-node/src/worker/fetch.rs
FintanH: It was initially causing me issues -- I can't remember why but it was something to do with `HEAD` or `FETCH_HEAD`. Everything works removing it though

radicle-node/src/worker/fetch/refspecs.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.into_iter()
+            .flat_map(AsRefspecs::into_refspecs)
+            .collect()
+    }
+}
+
+impl AsRefspecs for Remote {
+    fn as_refspecs(&self) -> Vec<Refspec<git::PatternString, git::PatternString>> {
+        let ns = remote(&self.id);
+        // XXX: the references in Refs are expected to be Qualified
```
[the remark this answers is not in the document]
FintanH: It's an `Nb.` :)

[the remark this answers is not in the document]
cloudhead: After doing some research, it looks like you are using it correctly, and I not; but regardless I prefer more descriptive headers so we know if it's a todo or not etc. :)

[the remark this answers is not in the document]
cloudhead: Funny enough, I picked up `Nb.` from Kim :thinking:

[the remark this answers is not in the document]
FintanH: I guess one can use both :smile:

cloudhead (approved):
Great work!

PR #477, opened 2023-03-23
Fix some node re-connection issues

cloudhead:
*See commits*

### DIFF
--- a/radicle-node/src/service/session.rs
+++ b/radicle-node/src/service/session.rs
@@ -5,7 +5,6 @@ use radicle::storage::Namespaces;
 
 use crate::service::message;
 use crate::service::message::Message;
-use crate::service::storage;
 use crate::service::{Id, LocalTime, NodeId, Reactor, Rng};
 use crate::Link;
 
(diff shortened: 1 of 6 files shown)


radicle-node/src/tests.rs
```diff
@@ -900,16 +900,12 @@ fn test_persistent_peer_reconnect_attempt() {
...
 
-    // A transient error such as this will cause Alice to attempt a reconnection.
-    let error = Arc::new(io::Error::from(io::ErrorKind::ConnectionReset));
-
-    // A non-transient disconnect, such as one requested by the user will not trigger
-    // a reconnection.
-    alice.disconnected(eve.id(), &DisconnectReason::Dial(error.clone()));
-    assert_matches!(alice.outbox().next(), None);
+    // A non-transient disconnect, such as one due to peer misbehavior will still strigger a
+    // a reconnection, since this is a persistent peer.
```
slack-coder, line 904: strigger => trigger

radicle-node/src/service/session.rs
```diff
@@ -132,6 +134,7 @@ impl Error {
     pub fn is_transient(&self) -> bool {
         match self {
             Self::InvalidTimestamp(_) => false,
+            Self::ProtocolMismatch => true,
```
slack-coder, line 137: How come a protocol mismatch is considered transient?

FintanH (approved):

radicle-node/src/service/session.rs
cloudhead: Keep in mind this is a mismatch between the git and gossip protocols.

slack-coder: Maybe a better naming of the error or the error's comment/description?  By the naming alone, it can be confused with incompatible protocol versions, or a mismatch with other protocols (connecting to SSH somehow).  Also, in the later case the error is persistent (not transient).

cloudhead: Yeah, can add a doc comment :+1:

PR #478, opened 2023-03-23
More patch output improvements

cloudhead:
*See commits*


slack-coder:
Looks like 'rad-patch.md' example needs to be updated, otherwise :+1:

cloudhead:
Thanks!

PR #479, opened 2023-03-24
httpd: Add reactions to issue and patch json output

sebastinez:
This PR adds the reactions array to `json::issue` and `json::patch` output.

**Remarks**
- @FintanH I made the `cob::issue::Issue` properties public, similar to `cob::patch::Patch`, if we don't want to do that, maybe I can find a workaround to get the `thread` property some other way..

### DIFF
--- a/radicle-httpd/src/api/json.rs
+++ b/radicle-httpd/src/api/json.rs
@@ -8,8 +8,9 @@ use serde_json::{json, Value};
 
 use radicle::cob::issue::{Issue, IssueId};
 use radicle::cob::patch::{Patch, PatchId};
-use radicle::cob::thread::{self, CommentId};
-use radicle::cob::{ActorId, Author, Timestamp};
+use radicle::cob::thread;
+use radicle::cob::thread::{CommentId, Thread};
+use radicle::cob::{ActorId, Author, Reaction, Timestamp};
 use radicle::git::RefString;
 use radicle::storage::{git, refs, ReadRepository};
 use radicle_surf::blob::Blob;
@@ -98,7 +99,10 @@ pub(crate) fn issue(id: IssueId, issue: Issue) -> Value {
         "title": issue.title(),
         "state": issue.state(),
         "assignees": issue.assigned().collect::<Vec<_>>(),
-        "discussion": issue.comments().collect::<Comments>(),
+        "discussion": issue
+          .comments()
+          .map(|(id, comment)| Comment::new(id, comment, issue.thread()))
+          .collect::<Vec<_>>(),
         "tags": issue.tags().collect::<Vec<_>>(),
     })
 }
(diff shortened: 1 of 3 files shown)


FintanH:
> I made the cob::issue::Issue properties public, similar to cob::patch::Patch, if we don't want to do that, maybe I can find a workaround to get the thread property some other way..

My general approach to thinking about these things is this: if some data type is expected to be mutated/changed via an API then `pub` fields should be avoided. In this case, an `Issue` is expected to be mutated by applying operations and so I think `pub` fields would be incorrect. It's convenient to make them `pub` so that they can be accessed but it breaks the API.

So my suggestion would be a `pub fn thread(&self)` method.

That being said, if `Patch` does it too then I'm not sure what our "style" should be going forward :smile:

radicle-httpd/src/api/json.rs
```diff
@@ -136,33 +142,24 @@ fn name_in_path(path: &str) -> &str {
...
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
-struct Comment {
+struct Comment<'a> {
     id: CommentId,
     author: Author,
-    body: String,
-    reactions: [String; 0],
+    body: &'a str,
+    reactions: Vec<(&'a ActorId, &'a Reaction)>,
```
FintanH, line 149: I'm curious as to why the lifetimes are needed :)

sebastinez: The lifetimes are needed due to `thread.reactions(&EntryId)` returning `Iterator<Item = (&ActorId, &Reaction)>` so AFAIK we either clone the tuples by mapping over them, or we pass these lifetimes into the `Comment` struct.
But maybe I'm missing something, I'm a lifetime amateur hehe

sebastinez: And the `body: &'a str` lifetime I added to avoid converting `comment.body()` into a `String`

sebastinez: The lifetimes are needed due to thread.reactions(&EntryId) returning Iterator<Item = (&ActorId, &Reaction)> so AFAIK we either clone the tuples by mapping over them, or we pass these lifetimes into the Comment struct.
But maybe I'm missing something, I'm a lifetime amateur hehe
And the body: &'a str lifetime I added to avoid converting comment.body() into a String

sebastinez:
> > I made the cob::issue::Issue properties public, similar to cob::patch::Patch, if we don't want to do that, maybe I can find a workaround to get the thread property some other way..
> 
> My general approach to thinking about these things is this: if some data type is expected to be mutated/changed via an API then `pub` fields should be avoided. In this case, an `Issue` is expected to be mutated by applying operations and so I think `pub` fields would be incorrect. It's convenient to make them `pub` so that they can be accessed but it breaks the API.
> 
> So my suggestion would be a `pub fn thread(&self)` method.
> 
> That being said, if `Patch` does it too then I'm not sure what our "style" should be going forward 😄

Interesting, yeah that makes sense, ok will look into adding a method, and we can see later if we want to limit `Patch` as well

sebastinez:
@FintanH added a method to `Issue` to get the thread, and avoid making the properties public in 0536df2

radicle-httpd/src/api/json.rs
FintanH: Cool, ya that makes sense :+1:

cloudhead (approved):

PR #480, opened 2023-03-24
httpd: Add patch head branches to patch output

sebastinez:
This PR adds the `RefStrings` of the branches that have the patch head.

### DIFF
--- a/radicle-httpd/src/api/json.rs
+++ b/radicle-httpd/src/api/json.rs
@@ -9,10 +9,12 @@ use serde_json::{json, Value};
 use radicle::cob::issue::{Issue, IssueId};
 use radicle::cob::patch::{Patch, PatchId};
 use radicle::cob::thread::{self, CommentId};
-use radicle::cob::{Author, Timestamp};
+use radicle::cob::{ActorId, Author, Timestamp};
+use radicle::git::RefString;
+use radicle::storage::{git, refs, ReadRepository};
 use radicle_surf::blob::Blob;
 use radicle_surf::tree::Tree;
-use radicle_surf::{Commit, Stats};
+use radicle_surf::{Commit, Oid, Stats};
 
 use crate::api::auth::Session;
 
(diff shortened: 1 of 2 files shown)


radicle-httpd/src/api/json.rs
```diff
@@ -134,6 +137,24 @@ fn name_in_path(path: &str) -> &str {
...
+    head: &Oid,
+) -> Result<Vec<RefString>, refs::Error> {
+    let remote = repo.remote(id)?;
+    let branches = remote
+        .refs
+        .iter()
+        .filter_map(|(name, o)| match o == head {
+            true => Some(name.to_owned()),
+            false => None,
+        })
```
cloudhead, line 152: Maybe you can use an `if` instead of the `match`?

radicle-httpd/src/api/json.rs
```diff
@@ -117,6 +119,7 @@ pub(crate) fn patch(id: PatchId, patch: Patch) -> Value {
                 "description": rev.description(),
                 "base": rev.base,
                 "oid": rev.oid,
+                "branches": get_branches(repo, patch.author().id(), &rev.oid).unwrap_or(vec![]),
```
cloudhead, line 122: So one important detail is that these may not all be branches. I think tags could be returned as well here, or refs that aren't branches.

I'd suggest calling this key `refs` for now, to be correct. Later, we could filter it to only include branches, but it's a bit trickier.

cloudhead: Also would change the function to `get_refs`

radicle-httpd/src/api/json.rs
sebastinez: fixed in e438753

cloudhead: Hehe it's a pretty common pattern :shrug: 

There's also https://doc.rust-lang.org/std/primitive.bool.html#method.then_some

cloudhead (approved):

radicle-httpd/src/api/json.rs
sebastinez: Oh nice! Will jot that down

PR #482, opened 2023-03-24
node: Fix fetch requests code

cloudhead:
We were not correctly deleting the fetch request, since only one was kept per RID.


FintanH (approved):
Ah, good catch +1

PR #487, opened 2023-03-24
Reduce pub accessors in patch and issue cob structs

sebastinez:
Replaces pub accessors to struct fields with methods.
Also creates some methods to compute necessary information

As talked about here https://github.com/radicle-dev/heartwood/pull/479#issuecomment-1482503388
and here https://github.com/radicle-dev/heartwood/pull/479#issuecomment-1482731408

There are also other places where we eventually could reduce public exposure, but one PR at a time

### DIFF
--- a/radicle/src/cob/patch.rs
+++ b/radicle/src/cob/patch.rs
@@ -130,20 +130,20 @@ pub enum MergeTarget {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Patch {
     /// Title of the patch.
-    pub title: LWWReg<Max<String>>,
+    title: LWWReg<Max<String>>,
     /// Patch description.
-    pub description: LWWReg<Max<String>>,
+    description: LWWReg<Max<String>>,
     /// Current state of the patch.
-    pub state: LWWReg<Max<State>>,
+    state: LWWReg<Max<State>>,
     /// Target this patch is meant to be merged in.
-    pub target: LWWReg<Max<MergeTarget>>,
+    target: LWWReg<Max<MergeTarget>>,
     /// Associated tags.
-    pub tags: LWWSet<Tag>,
+    tags: LWWSet<Tag>,
     /// List of patch revisions. The initial changeset is part of the
     /// first revision.
-    pub revisions: GMap<RevisionId, Redactable<Revision>>,
+    revisions: GMap<RevisionId, Redactable<Revision>>,
     /// Timeline of operations.
-    pub timeline: GSet<(Lamport, EntryId)>,
+    timeline: GSet<(Lamport, EntryId)>,
 }
 
 impl Semilattice for Patch {
@@ -172,18 +172,22 @@ impl Default for Patch {
 }
 
 impl Patch {
+    /// Title of the patch.
     pub fn title(&self) -> &str {
         self.title.get().get()
     }
 
+    /// Current state of the patch.
     pub fn state(&self) -> State {
         *self.state.get().get()
     }
 
+    /// Target this patch is meant to be merged in.
     pub fn target(&self) -> MergeTarget {
         *self.target.get().get()
     }
 
+    /// Timestamp of the first revision of the patch.
     pub fn timestamp(&self) -> Timestamp {
         self.revisions()
             .next()
@@ -192,14 +196,17 @@ impl Patch {
             .timestamp
     }
 
+    /// Associated tags.
     pub fn tags(&self) -> impl Iterator<Item = &Tag> {
         self.tags.iter()
     }
 
+    /// Patch description.
     pub fn description(&self) -> &str {
         self.description.get().get()
     }
 
+    /// Author of the first revision of the patch.
     pub fn author(&self) -> &Author {
         &self
             .revisions()
@@ -209,6 +216,8 @@ impl Patch {
             .author
     }
 
+    /// List of patch revisions. The initial changeset is part of the
+    /// first revision.
     pub fn revisions(&self) -> impl DoubleEndedIterator<Item = (&RevisionId, &Revision)> {
         self.timeline.iter().filter_map(|(_, id)| {
             self.revisions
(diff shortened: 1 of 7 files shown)


radicle/src/cob/patch.rs
```diff
@@ -394,10 +394,34 @@ impl Revision {
...
+        &self.base
+    }
+
+    pub fn oid(&self) -> &git::Oid {
+        &self.oid
+    }
+
+    pub fn timestamp(&self) -> &Timestamp {
+        &self.timestamp
+    }
```
cloudhead, line 411: `Timestamp` is `Copy` I believe, So you can just return `Timestamp`

radicle/src/cob/patch.rs
```diff
@@ -394,10 +394,34 @@ impl Revision {
...
 
+    pub fn author(&self) -> &Author {
+        &self.author
+    }
+
+    pub fn base(&self) -> &git::Oid {
+        &self.base
+    }
+
+    pub fn oid(&self) -> &git::Oid {
```
cloudhead, line 405: We should call this `head` I think, would be more obvious what oid it refers to.

radicle/src/cob/patch.rs
```diff
@@ -394,10 +394,34 @@ impl Revision {
...
+        &self.discussion
+    }
+
     pub fn merges(&self) -> impl Iterator<Item = &Merge> {
         self.merges.iter().map(|m| m.get())
     }
 
+    pub fn merges_is_empty(&self) -> bool {
+        self.merges.is_empty()
+    }
```
cloudhead, line 423: This function is too specific to be useful. If you return `impl ExactSizeIterator` from `merges()`, then you can just use `merges().is_empty()`

radicle/src/cob/patch.rs
```diff
@@ -427,11 +451,25 @@ impl fmt::Display for State {
...
-    pub timestamp: Timestamp,
+    timestamp: Timestamp,
+}
+
+impl Merge {
+    pub fn node(&self) -> &NodeId {
+        &self.node
+    }
+
+    pub fn commit(&self) -> &git::Oid {
```
cloudhead, line 466:
```suggestion
    pub fn commit(&self) -> git::Oid {
```

radicle/src/cob/patch.rs
```diff
@@ -427,11 +451,25 @@ impl fmt::Display for State {
...
+        &self.commit
+    }
+
+    pub fn timestamp(&self) -> &Timestamp {
```
cloudhead, line 470:
```suggestion
    pub fn timestamp(&self) -> Timestamp {
```

radicle/src/cob/patch.rs
```diff
@@ -818,9 +870,9 @@ impl<'a, 'g> Deref for PatchMut<'a, 'g> {
 #[derive(Default, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct PatchCounts {
-    pub proposed: usize,
-    pub draft: usize,
-    pub archived: usize,
+    proposed: usize,
+    draft: usize,
+    archived: usize,
 }
```
cloudhead, line 877: I wouldn't change `PatchCounts`, `CodeLocation`, `Merge` and `IssueCounts`, since they are immutable, and likely to remain that way.

radicle/src/cob/patch.rs
cloudhead: Ah, perhaps we need to implement that on `LWWSet` then. Want to try?

sebastinez: Ok I tried a few times to get this working, but in the end the major point is that `ExactSizeIterator::is_empty()` is still part of the nightly_only API.
https://doc.rust-lang.org/stable/std/iter/trait.ExactSizeIterator.html#method.is_empty

So we need still would need to rely on checking if the length is equal to 0, which doesn't require any new implementation.

Maybe we can revisit this once `ExactSizeIterator::is_empty()` is stable?

sebastinez: Yeah I tried, but on one hand the `is_empty()` method on `ExactSizeIterator` is still nightly only and I'm not able to implement `ExactSizeIterator` successfully on `LWWSet`

radicle/src/cob/patch.rs
```diff
@@ -461,6 +481,20 @@ pub struct Merge {
...
+
+    pub fn commit(&self) -> git::Oid {
+        self.commit
+    }
+
+    pub fn timestamp(&self) -> Timestamp {
+        self.timestamp
+    }
+}
+
```
cloudhead, line 497: Can be removed

radicle/src/cob/patch.rs
```diff
@@ -532,24 +566,38 @@ impl Ord for CodeLocation {
...
+        &self.location
+    }
+
+    pub fn comment(&self) -> &str {
+        &self.comment
+    }
+
+    pub fn timestamp(&self) -> &Timestamp {
+        &self.timestamp
+    }
```
cloudhead, line 587: Can you copy over the field comments to the methods?

radicle/src/cob/patch.rs
```diff
@@ -532,24 +566,38 @@ impl Ord for CodeLocation {
...
+    verdict: LWWReg<Option<Verdict>>,
     /// Review general comment.
-    pub comment: LWWReg<Option<Max<String>>>,
+    comment: LWWReg<Option<Max<String>>>,
     /// Review inline code comments.
-    pub inline: LWWSet<Max<CodeComment>>,
+    inline: LWWSet<Max<CodeComment>>,
     /// Review timestamp.
-    pub timestamp: Max<Timestamp>,
+    timestamp: Max<Timestamp>,
```
cloudhead, line 600: Can you copy the field docs to the methods?

radicle/src/cob/patch.rs
sebastinez: You mean the entire impl block of the `Merge` struct?

sebastinez: We need the Merge getters in different places

cloudhead: The methods; Merge already has public fields

sebastinez: Ah right, got it

PR #488, opened 2023-03-24
Canonical `refs/rad/id`

FintanH:
Introduce a canonical reference for refs/rad/id to aid in discovering
information about a repository without knowing a peer -- e.g. if a
peer wants to clone a new repository it needs to find out which peers
are delegates for that project, but does not know what their NIDs are
up front.

The approach mirrors the canonical `head` for a repository and
introduces useful constructors for `Doc`/`DocAt`.

### DIFF
--- a/radicle-node/src/worker.rs
+++ b/radicle-node/src/worker.rs
@@ -328,6 +328,8 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
             log::debug!(target: "worker", "Fetch for {} exited successfully", rid);
             let head = repo.set_head()?;
             log::debug!(target: "worker", "Head for {} set to {head}", rid);
+            let head = repo.set_identity_head()?;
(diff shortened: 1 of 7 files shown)


radicle-node/src/worker.rs
```diff
@@ -327,7 +327,9 @@ impl<G: Signer + Ecdh + 'static> Worker<G> {
         if result.success() {
             log::debug!(target: "worker", "Fetch for {} exited successfully", rid);
             let head = repo.set_head()?;
-            log::debug!(target: "worker", "Head for {} set to {head}", rid);
+            log::debug!(target: "worker", "HEAD for {} set to {head}", rid);
```
cloudhead, line 330: It's not just `HEAD`, it's also the default branch ref :)

cloudhead (approved):
Looks good.

radicle-node/src/worker.rs
FintanH: Ah fair enough! I'll revert :)

PR #489, opened 2023-03-25
node: systemd file

slack-coder:
Support users with systemd integration by providing a systemd service file.  The file includes basic instructions for how to adapt to the users context.

There are two scenarios it would be run in, as a system service, or as a user service.  The scenario as a user service is problematic. `radicle-node` relies on ssh-agent or a clear text RAD_PASSPHRASE.


radicle-node/radicle-node.service
```diff
@@ -0,0 +1,26 @@
...
+#
+#   [Service]
+#   Environment="RAD_PASSPHRASE=[passphrase]"
+
+[Unit]
+After=syslog.target network.target
+Description=Radicle server daemon
+
+[Service]
+ExecStart=/usr/sbin/radicle-node
```
cloudhead, line 21:
```suggestion
ExecStart=%h/.radicle/bin/radicle-node
```
We won't be installing this in system folders, so this would make the most sense I think.

radicle-node/radicle-node.service
```diff
@@ -0,0 +1,26 @@
...
+#   cp radicle-node.service $HOME/.local/share/systemd/user/radicle-node.service
+#
+# You will have to set the RAD_PASSPHRASE for the node to start.
+#
+#   [Service]
+#   Environment="RAD_PASSPHRASE=[passphrase]"
+
+[Unit]
+After=syslog.target network.target
+Description=Radicle server daemon
```
cloudhead, line 18:
```suggestion
Description=Radicle Node
```

radicle-node/radicle-node.service
```diff
@@ -0,0 +1,26 @@
...
+
+[Service]
+ExecStart=/usr/sbin/radicle-node
+KillMode=process
+Restart=always
```
cloudhead, line 23: I'd set `RestartSec=1` as well to not flood the logs if there's a problem.

radicle-node/radicle-node.service
slack-coder: How about "./local/bin"? It is recommended in [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).  Though better places may be `/usr/local/sbin` or `/opt/radicle/bin/radicle`.

"User-specific executable files may be stored in $HOME/.local/bin. Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place.

Since $HOME might be shared between systems of different achitectures, installing compiled binaries to $HOME/.local/bin could cause problems when used on systems of differing architectures. This is often not a problem, but the fact that $HOME becomes partially achitecture-specific if compiled binaries are placed in it should be kept in mind."

cloudhead: We don't use XDG standards though, eg. we have everything else already under `~/.radicle`

slack-coder: Are there other applications which do this?  I haven't seen this done or recommended before.  The other standards I'd recommend `/opt/radicle/bin/radicle-node` or `/usr/local/bin` which are common approaches.

cloudhead: There are some yes, for example `cargo` puts binaries in ~/.cargo/bin, and `go` puts them in `$GOPATH`.

cloudhead: image

slack-coder: Is that recommended though?  Radicle will be the only project I know of doing this officially.

cloudhead: Cargo & rustup do it too as demonstrated. There's no better location that doesn't require administrator priviledges.

slack-coder: Looks like we commented at the same time.

You're right, from their [page](https://www.rust-lang.org/tools/install):

```
In the Rust development environment, all tools are installed to the ~/.cargo/bin directory, and this is where you will find the Rust toolchain, including rustc, cargo, and rustup. 
```

PR #493, opened 2023-03-28
cli: add semver option

FintanH:
If a consuming application wishes to obtain the semver using the `--version` option, then it is required to parse the version output for the semver itself.

This can be subject to change, so instead add a `--semver` option to print only the semver version to stdout.

An alternative approach to #492

### DIFF
--- a/radicle-cli/src/main.rs
+++ b/radicle-cli/src/main.rs
@@ -16,6 +16,7 @@ enum Command {
     Other(Vec<OsString>),
     Help,
     Version,
+    Semver,
 }
 
 fn main() {
@@ -43,6 +44,9 @@ fn parse_args() -> anyhow::Result<Command> {
             }
             Long("version") => {
                 command = Some(Command::Version);
+            }
+            Long("semver") => {
+                command = Some(Command::Semver);
             }
             Value(val) if command.is_none() => {
                 if val == *"." {
@@ -82,6 +86,9 @@ fn run(command: Command) -> Result<(), Option<anyhow::Error>> {
     match command {
         Command::Version => {
             print_version();
+        }
+        Command::Semver => {
+            println!("{VERSION}")
         }
         Command::Help => {
             print_help()?;


slack-coder:
As pointed out, the second field in the `--version` output is the semantic version with '+GIT_COMMIT' being appended build information to it.

slack-coder:
Hey @FintanH, sorry, I was misspoke.  I meant in relation to #492, what was wrong with using regexp to get the version?  Because the second field in that case is rad's full semantic version.

FintanH:
> Hey @FintanH, sorry, I was misspoke. I meant in relation to https://github.com/radicle-dev/heartwood/pull/492, what was wrong with using regexp to get the version? Because the second field in that case is rad's full semantic version.

The argument was that the output of `--version` could be subject to change, e.g. it turns to `rad labs 6.0.0`. Whereas this would specifically output the semver of the CLI and nothing else.

It _seems_ useful and low maintenance :shrug:

slack-coder:
I believe a better question is what's the standard expectations for VS code plugin writers for getting command versions?

Has the --semver approach been used elsewhere?  Perhaps we should consider '--version' a part of rad's API.  Personally I've always seen '--version' as output for humans, maybe that was the mistake.

cloudhead:
Yeah it's a little strange to me to add this flag, given the information is available in `--version` and is quite easy to parse.

FintanH:
That's fair. Was adding a QoL improvement for the VScode crew, but I'll close it 👍

maninak:
To close the loop, here's our actual implementation to get the CLI version.

https://github.com/cytechmobile/radicle-vscode-extension/blob/e717e507050266b1a814749c63351f9931eead3f/src/utils/radCli.ts#L41

The regex should be robust enough to keep working even as the output of the `rad --version` command changes somewhat, but if you know you're changing it and remember it, please give us a heads-up just in case. :beers:

slack-coder:
> That's fair. Was adding a QoL improvement for the VScode crew, but I'll close it +1

It was a good move either way :+1:

slack-coder:
How about reminding ourselves via function comment? https://github.com/radicle-dev/heartwood/pull/518

PR #495, opened 2023-03-28
node: fetch trusted peers on clone

FintanH:
When a repository does not yet exist during a fetch, i.e. a clone, then only delegates are being fetched.

Augment Namespaces::Many to hold the trusted and optional delegate peers separately. Doing so allows the construction of the variant without the repository existing, but tracking relationships existing.

This variant can then be used to build refspecs for both the trusted peers and delegates when doing a cloning fetch.

### DIFF
--- a/radicle-node/src/worker/fetch.rs
+++ b/radicle-node/src/worker/fetch.rs
@@ -132,11 +132,7 @@ impl<'a> StagingPhaseInitial<'a> {
                 let doc = doc.verified()?;
                 let mut trusted = match self.namespaces.clone() {
                     Namespaces::All => HashSet::new(),
-                    // TODO(finto): this is one of those cases where
-                    // having the `One` variant doesn't make any
-                    // sense.
-                    Namespaces::One(pk) => [pk].into_iter().collect(),
-                    Namespaces::Many(trusted) => trusted,
+                    Namespaces::Trusted(trusted) => trusted,
                 };
                 let delegates = doc.delegates.map(PublicKey::from);
                 trusted.extend(delegates);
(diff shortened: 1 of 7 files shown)


FintanH:
I'm leaving this as a draft for now, but the tests passed locally. I need to go over it tomorrow to double-check things :)

I get an uneasy feeling with `Namespaces` in tandem with the tracking. It tends to lead to little workarounds which feel prone to error, but I can't quite figure out a better way at doing (yet) :thinking:

radicle-node/src/worker/fetch.rs
```diff
@@ -124,14 +127,24 @@ impl<'a> StagingPhaseInitial<'a> {
...
                 let oid = ReadRepository::identity_head(repo)?;
                 log::trace!(target: "worker", "Loading 'rad/id' @ {oid}");
                 let (doc, _) = Doc::<Unverified>::load_at(oid, repo)?;
                 let doc = doc.verified()?;
-                doc.delegates.map(PublicKey::from)
+                let mut trusted = match self.namespaces.clone() {
+                    Namespaces::All => HashSet::new(),
+                    // TODO(finto): this is one of those cases where
+                    // having the `One` variant doesn't make any
+                    // sense.
```
cloudhead, line 141: Should we remove it and just have `Namespaces::All` and `Namespaces::Trusted(HashSet)`?

radicle-node/src/worker/fetch.rs
```diff
@@ -69,9 +67,14 @@ pub struct StagingPhaseFinal<'a> {
...
     /// The original [`Storage`] we are finalising changes into.
     production: &'a Storage,
+    /// The tracked remotes that the fetch is being performed
+    /// for. These are passed through from the
+    /// [`StagingRad::namespaces`], if the variant is `Many`.
+    // trusted: Vec<RemoteId>,
     /// The remotes that the fetch is being performed for. These are
     /// discovered after performing the fetch for [`StagingPhaseInitial`].
-    remotes: NonEmpty<RemoteId>,
+    // delegates: NonEmpty<RemoteId>,
```
cloudhead, line 76:
```suggestion
```

radicle-node/src/#worker.rs#
```diff
@@ -0,0 +1,680 @@
+mod fetch;
```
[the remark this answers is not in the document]
FintanH: Accidentally sneaking in emacs files :facepalm:

PR #497, opened 2023-03-29
cob: Use `LWWReg::new` instead of `LWWReg::from` in `Review::new`

sebastinez:
Using `LWWReg::from` was giving subsequent actions the same `clock` value as their predecessors.
This resulted in a `Reject` not being able to turn into an `Accept` on a second review of a patch.

The solution is to use the `clock` value that was produced for the `Op` and pass that into new, avoiding the call to `from`.


cloudhead (approved):
Nice catch :+1:

PR #498, opened 2023-03-29
radicle: force new or initial construction of LWWReg

### LINKED ISSUE #496: Threading clocks for updating actions
It came to our attention in the Zulip topic, ["patch review modification"](https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/patch.20review.20modification), that using `LWWReg::from` was giving subsequent actions the same clock value as their predecessors. This resulted in a `Reject` not being able to turn into an `Accept` on a second review of a patch.

The solution is to use the `clock` value that was produced for the `Op` and pass that into `new`, avoiding the call to `from`. There could be other potential areas where this is an issue in our cobs.

I would recommend that we remove the `From` impls in favour of always using `new`. This saves us from potentially using these things wrong in the future and `Clock::default()` is not hard to type :)

FintanH:
Using LWWReg::from will always set the clock value to the default, which can end up being used wrong.

Instead, introduce a constructor method `initial` to indicate that it should be used on initial construction, while `new` should be used for newer values of the register.

Fixes #496

### DIFF
--- a/radicle-crdt/src/lwwreg.rs
+++ b/radicle-crdt/src/lwwreg.rs
@@ -14,6 +14,16 @@ pub struct LWWReg<T, C = clock::Lamport> {
 }
 
 impl<T: Semilattice, C: PartialOrd> LWWReg<T, C> {
+    pub fn initial(value: T) -> Self
+    where
+        C: Default,
+    {
+        Self {
+            clock: Max::from(C::default()),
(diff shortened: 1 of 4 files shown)


FintanH:
Leaving this as a draft because I'm not sure how to solve the `Option` issue mentioned in the issue

PR #500, opened 2023-03-29
radicle: simplify patch.latest()

slack-coder:
Make patch.latest() simpler unwrapping the Option out of the return type.  A patch is expected to have one or more revisions.  This is reflected in other methods already such as `head()`.

### DIFF
--- a/radicle-cli/src/commands/patch/checkout.rs
+++ b/radicle-cli/src/commands/patch/checkout.rs
@@ -53,9 +53,7 @@ fn find_patch_commit<'a>(
         Ok(commit) => Ok(commit),
         Err(e) if git::ext::is_not_found_err(&e) => {
             // TODO: Handle case of concurrent revisions.
-            let (_, rev) = patch
-                .latest()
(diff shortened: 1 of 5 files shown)


FintanH:
> A patch is expected to have one or more revisions. This is reflected in other methods already such as head().

_Technically_, that's not true. If you look at the [`revisions`](https://github.com/radicle-dev/heartwood/blob/master/radicle/src/cob/patch.rs#L205-L212) function, that is used by `latest`, it will filter out redacted revisions. So a patch could be created with a single revision which then gets redacted.

I'm unsure how this should affect this PR though :thinking:

cloudhead:
`latest()` should actually return a `Vec`, since there may be concurrent revisions..

slack-coder:
It looks like the codebase is confused in parts and needs some work then.  As mentioned, patch.head() and others already assume one or more revisions.


I'll iterate on it here

slack-coder:
Closed in favor of creating a follow up issue

PR #502, opened 2023-03-30
Use cached canonical `rad/id` refs

cloudhead:
We were recomputing the `rad/id` ref every time it was accessed.

@FintanH any idea why the `rad-id.md` tests fail? Seems like there's some discrepancy between the cached and canonical ref, which makes me think we aren't setting the ref in some crucial place during `rad id`?

### DIFF
--- a/radicle/src/cob/identity.rs
+++ b/radicle/src/cob/identity.rs
@@ -230,9 +230,12 @@ impl Proposal {
             self.description().unwrap_or_default()
         );
         let current = doc.update(remote, &msg, &signatures.collect::<Vec<_>>(), repo.raw())?;
+        let head = repo.set_identity_head()?;
+
+        assert_eq!(head, current);
 
         Ok(Identity {
-            head: current,
+            head,
             root: previous.root,
             current,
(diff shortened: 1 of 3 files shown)


FintanH:
Ah, ya I forgot about that case. Maybe it makes sense to `set_identity_head` within `Doc::commit`, or somewhere along those lines -- instead of having it in the `rad::init` function _and_ in the `id` proposal code. Otherwise, we should also be setting it in here https://github.com/radicle-dev/heartwood/blob/master/radicle/src/cob/identity.rs#L190

cloudhead:
Ah nice, ok.

Not sure about `Doc::commit`, seems slightly orthogonal :thinking: -- I also like that set_head and set_identity_head are called together in `rad-init`..

FintanH:
> Ah nice, ok.
> 
> Not sure about `Doc::commit`, seems slightly orthogonal thinking -- I also like that set_head and set_identity_head are called together in `rad-init`..

Fair, I suggested it so one doesn't have to remember to set the identity head when a new commit is created.

PR #504, opened 2023-03-31
node: Make sure we retry on lock contention

cloudhead:
This fixes an issue that came up in tests where we would be reading and writing to the database at the same time, and it would fail with an error due to the lock being contended on.


FintanH (approved):
Nice 👌

PR #505, opened 2023-03-31
radicle: add test for removing review fields

FintanH:
Add a test case showing that when a review is made with a None value to an already set field, it will remove the original value.


radicle/src/cob/patch.rs
```diff
@@ -603,8 +603,14 @@ impl CodeComment {
...
 pub struct Review {
     /// Review verdict.
+    ///
+    /// Nb. if the verdict is set and a subsequent review is made with
+    /// the verdict as `None`, the original verdict will be nullified.
     verdict: LWWReg<Option<Verdict>>,
     /// Review general comment.
+    ///
+    /// Nb. if the comment is set and a subsequent review is made with
+    /// the comment as `None`, the original comment will be nullified.
```
slack-coder, line 613: Wouldn't this be expected behavior of LWWReg<Option<..>>?

slack-coder:
Otherwise :+1:

radicle/src/cob/patch.rs
cloudhead: It is, but maybe it's not so obvious from an API perspective? I guess another approach would be to add this comment to the API function, ie. `patch.update`? :shrug:

slack-coder: Or patch.review/comment?  I'd tend towards the API comment as well.  Otherwise maybe its a comment on LWWReg<Option<..>> just to avoid repetition.

FintanH: I think this is everywhere `LWWReg<Option<_>>` is used 😄 But I get yis, it does make sense to have it on the API, however, which part of the API? The `Action`? In the `Transaction`? Somewhere else?

I'll be able to update this when I'm fully back, otherwise happy for someone to change it to get the PR over the line :)

PR #512, opened 2023-04-03
Set patch status to `merged` automatically

cloudhead:
When a merge is applied to a patch, we check whether the threshold of delegates has been reached, at which point we set the patch status to `merged`.

This requires threading the identity document oid through the "apply" process.


slack-coder:
> This requires threading the identity document oid through the "apply" process.

What's meant by 'threading' here?  I haven't seen the term used in this way until Radicle.

cloudhead:
By "threading" I mean to pass state through various functions. Comes from "threading a needle"

radicle/src/cob/identity.rs
```diff
@@ -16,7 +16,7 @@ use crate::{
         store::{self, FromHistory as _, Transaction},
     },
     identity::{doc::DocError, Did, Identity, IdentityError},
-    prelude::Doc,
+    prelude::{Doc, ReadRepository},
```
slack-coder, line 19: What about `prelude::*`?  Prelude's should be the special import where we import everything no?

cloudhead: We could treat it that way if we wanted :shrug:

radicle/src/cob/patch.rs
```diff
@@ -376,6 +384,16 @@ impl store::FromHistory for Patch {
...
+                        let doc = repo.identity_doc_at(op.identity)?;
+
+                        if revision
+                            .merges()
+                            .filter(|m| doc.is_delegate(&m.node))
+                            .count()
+                            >= doc.threshold
+                        {
+                            self.state.set(State::Merged, op.clock);
+                        }
```
slack-coder, line 396: How will this threshold work if delegates merge into different levels of the target branch?  For example, what happens if there are two patches, and delegates merge them out of order.

radicle/src/cob/patch.rs
```diff
@@ -376,6 +384,16 @@ impl store::FromHistory for Patch {
                             .into(),
                             op.clock,
                         );
+                        let doc = repo.identity_doc_at(op.identity)?;
```
slack-coder, line 387: Should the doc be verified somewhere?

radicle-cli/src/commands/patch.rs
```diff
@@ -83,7 +91,9 @@ pub enum Operation {
     Checkout {
         patch_id: Rev,
     },
-    List,
+    List {
+        filter: Option<patch::State>,
```
slack-coder, line 95: Maybe `filter_state`, `with_state`, or even `state`?  Something which describes the attribute value used for filtering.

radicle-cli/src/commands/patch/list.rs
```diff
@@ -16,17 +17,25 @@ pub fn run(
...
 
-    for (id, patch, _) in proposed {
+    for patch in all {
+        let (id, patch, _) = patch?;
+
+        if let Some(filter) = filter {
+            if patch.state() != filter {
+                continue;
+            }
+        }
```
slack-coder, line 38: Would this be a performant alternative?

```
        if filter == Some(patch.state()) {
                continue;
        }
```

radicle/src/cob/patch.rs
cloudhead: Good question! Generally I would say that only one delegate should be doing the merges for that reason. Note that this isn't an issue related to radicle: any model with multiple forks will have to answer this question; and the answer is usually that one person (or a bot) takes care of merging, and the others pull from there.

We could have smarter rules for setting the state to Merged, but it's outside the scope of this PR.

radicle/src/cob/patch.rs
cloudhead: The doc is verified on fetch, so it shouldn't be necessary to verify again here.

radicle-cli/src/commands/patch.rs
cloudhead: `Option<State>` is the filter. `with_state` and `filter_state` wouldn't make sense when set to `None`. `state` wouldn't communicate the intent, and we already know it's the state by looking at the type.

radicle-cli/src/commands/patch/list.rs
cloudhead: The first option could work, but we're calling `patch.state()` even if filter is `None`, so it seems wasteful.
`is_some_and` is not part of stable rust.

FintanH: I think you can use [`contains`](https://doc.rust-lang.org/std/option/enum.Option.html#method.contains) but unfortunately that doesn't give you the laziness either

radicle-cli/src/commands/patch.rs
slack-coder: > we already know it's the state by looking at the type

Using the type to infer its use is adding a level of indirection and we shouldn't expect user interfaces for the source code to do it.

cloudhead: Not sure what you mean.. my argument boils down to `state` being less informative than `filter`.

PR #514, opened 2023-04-03
Upload output binaries to a HTTP server

adaszko:
The new location for binaries is https://files.radicle.xyz/


.github/workflows/release.yml
```diff
@@ -105,6 +43,18 @@ jobs:
...
+        env:
+          SSH_KEY: ${{ secrets.SSH_PRIVATE_KEY }}
+      - shell: bash
+        run: chmod go-rwx ssh-private-key
+      - shell: bash
+        run: ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ssh-private-key github-actions@files.radicle.xyz mkdir -p "heartwood-artifacts/${{ matrix.target }}/${{ github.sha }}"
+      - shell: bash
+        run: scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ssh-private-key -r "${{ matrix.target }}/*" "github-actions@files.radicle.xyz:heartwood-artifacts/${{ matrix.target }}/${{ github.sha }}/"
+      - shell: bash
+        run: ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ssh-private-key github-actions@files.radicle.xyz ln -sf "heartwood-artifacts/${{ matrix.target }}/${{ github.sha }}" "heartwood-artifacts/${{ matrix.target }}/latest"
```
cloudhead, line 57: Can we `s/heartwood-artifacts/releases`?

adaszko: Yep

adaszko:
> Could we have the files in the `.tar.gz` not be within a sub-folder? It makes scripting a lot easier..

I believe tarballs like that are called a "tarbomb" 😉

Are you aware of the `--strip-components=NUMBER` option to tar?

cloudhead:
Now I am :+1:

.github/workflows/release.yml
```diff
@@ -105,15 +101,12 @@ jobs:
...
-      - id: 'upload-file'
-        uses: 'google-github-actions/upload-cloud-storage@v1'
-        with:
-          path: ${{ matrix.target }}
-          destination: heartwood-artifacts/${{ github.sha }}
+      - shell: bash
+        run: |
+          echo "$SSH_PRIVATE_KEY" >ssh-private-key
+          chmod go-rwx ssh-private-key
+          ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ssh-private-key github-actions@files.radicle.xyz mkdir -p "/mnt/HC_Volume_30083854/releases/${{ matrix.target }}/${{ github.sha }}"
```
cloudhead, line 108: What is this `HC_Volume` thing, it seems like something that would break if we changed the volume?

adaszko: It’s a Hetzner mounted volume.  Those volumes are supposed to outlive VMs so I doubt it’s ever going to change

adaszko:
> If we want to keep the sub-folder, I would name the tar archive and sub-folder `radicle-<target>.tar.gz` I think. It's a bit hard to find in your downloads/ folder otherwise, or when you extract it.

Done: https://files.radicle.xyz/aarch64-apple-darwin/3c72cf0fa7635dcea72b9b178f612de667a824a4/

PR #515, opened 2023-04-03
node: Don't prune owned refs

cloudhead:
Owned refs were being pruned on fetch.


FintanH (approved):
I think this is the right decision, I thought it made sense initially but I think it needs some further thought and perhaps look into what Link was doing for pruning (if at all)

Also, thanks for the docs clean up!

PR #516, opened 2023-04-04
tui: Introduce view pages

erak:
**Rationale**

`tuirealm`'s event and prop system is designed to work with flat component hierarchies. Building deep nested component hierarchies would need a lot more additional effort to properly pass events and props down these hierarchies. This makes it hard to implement full app views (home, patch details etc) as components. 

**Implementation**

This introduces view pages that take into account the flat component hierarchies, and provide switchable sets of components.


cloudhead:
Merged.

PR #517, opened 2023-04-04
Add `install` script

cloudhead:
Downloads binaries and installs them, based on your operating system.

Can be used in a one liner, for example:

    sh <(curl https://radicle.xyz/install)

Depends on #514

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -1,0 +1,120 @@
+#!/bin/sh
+#
+# Radicle isntallation script.
+#
+set -e
+
+url() {
+  echo "https://files.radicle.xyz/$1/latest/radicle-$1.tar.gz"
+}
+
+info() {
+  printf "\033[36m$*\033[0m\n"
+}
+
+success() {
+  echo
+  printf "\033[32m✓\033[0m Radicle has been installed successfully. Run \`rad auth\` to get started.\n"
+}
+
+fatal() {
+  printf "\033[31merror\033[0m: $*\n" >&2
+  exit 1
+}
+
+target() {
+  TARGET=""
+
+  case "$(uname)/$(uname -m)" in
+  Darwin/arm64)
+    TARGET="aarch64-apple-darwin" ;;
+  Darwin/x86_64)
+    TARGET="x86_64-apple-darwin" ;;
+  Linux/arm64|Linux/aarch64)
+    TARGET="aarch64-unknown-linux-musl" ;;
+  Linux/x86_64)
+    TARGET="x86_64-unknown-linux-musl" ;;
+  *)
+    fatal "Your operating system is currently unsupported. Sorry!" ;;
+  esac
+  echo $TARGET
+}
+
+tempdir() {
+  if [ -n "$TMPDIR" ]; then
+    echo "$TMPDIR"
+  elif [ -d "/tmp" ]; then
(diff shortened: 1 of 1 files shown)


adaszko (approved):
Perhaps we should have a CI job that makes sure the installation script keeps on working on various OS’es?  Installation is the first point of contact with the project

scripts/install
```diff
@@ -0,0 +1,120 @@
...
+    TARGET="aarch64-unknown-linux-musl" ;;
+  Linux/x86_64)
+    TARGET="x86_64-unknown-linux-musl" ;;
+  *)
+    fatal "Your operating system is currently unsupported. Sorry!" ;;
+  esac
+  echo $TARGET
+}
+
+tempdir() {
```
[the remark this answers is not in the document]
cloudhead: Good catch, can delete.

cloudhead:
> Perhaps we should have a CI job that makes sure the installation script keeps on working on various OS’es? Installation is the first point of contact with the project

Hmm yeah that might be a good idea if we're changing the script a lot.

scripts/install
```diff
@@ -0,0 +1,120 @@
...
+  info "Configuring PATH in '$PROFILE'..."
+  echo                                    >> "$PROFILE"
+  echo "# Radicle binaries."              >> "$PROFILE"
+  echo "export PATH=\"\$PATH:$RAD_PATH\"" >> "$PROFILE"
+
+  success
+
+  # Ensure that the current shell has the updated PATH.
+  export PATH="$PATH:$RAD_PATH"
+  exec "$SHELL"
```
slack-coder, line 117: Might be better to give the user instructions what to do like what foundry does.

Also, would this create the new shell never printing the success below?

slack-coder: Also, $SHELL may be the users preferred shell, but it could be some other shell.

slack-coder:
Would a better place for the script be in the website's git repository?  It uses path's related to infrastructure, including this stuff here would make it this project's dependency.

cloudhead:
> Would a better place for the script be in the website's git repository? It uses path's related to infrastructure, including this stuff here would make it this project's dependency.

Good point... yes I think that makes sense. I will merge it here but eventually move it to the website.

scripts/install
slack-coder: Maybe foundry changed this? It doesn't appear in the [install script](https://raw.githubusercontent.com/foundry-rs/foundry/master/foundryup/install).

Also would this suspend running other commands after the script?  It would mean it could not be run from a script.  For instance "hello" would not be printed here.

```
curl ... | sh && echo "hello"
```

cloudhead: It is in the install script you linked:
```

# Only add foundryup if it isn't already in PATH.
if [[ ":$PATH:" != *":${FOUNDRY_BIN_DIR}:"* ]]; then
    # Add the foundryup directory to the path and ensure the old PATH variables remain.
    echo >> $PROFILE && echo "export PATH=\"\$PATH:$FOUNDRY_BIN_DIR\"" >> $PROFILE
fi
```

cloudhead: > Also would this suspend running other commands after the script? It would mean it could not be run from a script. For instance "hello" would not be printed here.

It works fine actually :)

PR #518, opened 2023-04-04
cli: Note '--version' output as API

slack-coder:
Consider our command line version output part of our API via a function comment.  The output from version is tested too to ensure no breaking changes occur accidentally.

The 'regex' crate is imported to support the test.  Though its heavy to bring it in for a test, 'regex' is a commonly used tool, and the crate itself is somewhat official, coming from rust's github organization repository.

Note: updated for new test

### DIFF
--- a/radicle-cli/src/main.rs
+++ b/radicle-cli/src/main.rs
@@ -1,4 +1,5 @@
 use std::ffi::OsString;
+use std::io;
 use std::{io::ErrorKind, iter, process};
 
 use anyhow::anyhow;
@@ -62,16 +63,20 @@ fn parse_args() -> anyhow::Result<Command> {
     Ok(command.unwrap_or_else(|| Command::Other(vec![])))
 }
 
-fn print_version() {
+/// Print the Radicle CLI's version.
+///
+/// Third party applications use it to parse Radicle Cli's version.
+fn print_version(mut w: impl std::io::Write) -> anyhow::Result<()> {
     if VERSION.contains("-dev") {
-        println!("{NAME} {VERSION}+{GIT_HEAD}")
+        write!(w, "{NAME} {VERSION}+{GIT_HEAD}")?;
     } else {
-        println!("{NAME} {VERSION} ({GIT_HEAD})")
+        write!(w, "{NAME} {VERSION} ({GIT_HEAD})")?;
     }
+    Ok(())
 }
 
 fn print_help() -> anyhow::Result<()> {
-    print_version();
+    print_version(&mut io::stdout())?;
     println!("{DESCRIPTION}");
     println!();
 
@@ -81,7 +86,7 @@ fn print_help() -> anyhow::Result<()> {
 fn run(command: Command) -> Result<(), Option<anyhow::Error>> {
     match command {
         Command::Version => {
-            print_version();
+            print_version(&mut io::stdout())?;
         }
         Command::Help => {
             print_help()?;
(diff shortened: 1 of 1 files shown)


radicle-cli/src/main.rs
```diff
@@ -62,6 +62,9 @@ fn parse_args() -> anyhow::Result<Command> {
     Ok(command.unwrap_or_else(|| Command::Other(vec![])))
 }
 
+/// Print radicle client's version.
```
cloudhead, line 65:
```suggestion
/// Print the Radicle CLI's version.
```

radicle-cli/src/main.rs
```diff
@@ -62,6 +62,9 @@ fn parse_args() -> anyhow::Result<Command> {
...
+///
+/// Third party applications use it to parse Radicle Cli's version.
```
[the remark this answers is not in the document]
slack-coder: Anything in particular to test?  The direct method of testing expected output wont give us much, just a rewrite of `print_version`.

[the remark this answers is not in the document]
cloudhead: I would test that it matches the pattern `radicle vX.X.X (<hex>)` maybe?

I don't think it's important, but neither do I think the comment will help.. It's on a private function.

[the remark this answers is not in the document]
slack-coder: > I would test that it matches the pattern radicle vX.X.X (<hex>) maybe?

Agreed, I'll do this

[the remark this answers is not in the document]
gsaslis: how about moving closer to semver ?

sounds like the `hex` could be added as either a pre-release version (https://semver.org/#spec-item-9), or build metadata (https://semver.org/#spec-item-10), no ?

slack-coder:
`print_version()` was updated to write to a stream to support testing.  Should we follow this method with other commands?

It will allow testing output via unit test's etc.

radicle-cli/src/main.rs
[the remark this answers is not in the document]
slack-coder: @gsaslis we're doing this for '-dev' builds i.e. deployments build from the repository directly, and not via releases.  I'm unsure there's a point for regular releases, because version at that point should always point to the given commit.

radicle-cli/Cargo.toml
```diff
@@ -18,6 +18,7 @@ json-color = { version = "0.7" }
 lexopt = { version = "0.2" }
 log = { version = "0.4", features = ["std"] }
 nonempty = { version = "0.8" }
+regex = { version = "1.7" }
```
cloudhead, line 21: :-1:

cloudhead: Regex is a big crate -- not worth depending on just for a test.

radicle-cli/src/main.rs
```diff
@@ -62,16 +63,20 @@ fn parse_args() -> anyhow::Result<Command> {
     Ok(command.unwrap_or_else(|| Command::Other(vec![])))
 }
 
-fn print_version() {
+/// Print the Radicle CLI's version.
+///
+/// Third party applications use it to parse Radicle Cli's version.
+fn print_version(w: &mut dyn std::io::Write) -> anyhow::Result<()> {
```
cloudhead, line 69:
```suggestion
fn print_version(w: impl io::Write) -> anyhow::Result<()> {
```

radicle-cli/Cargo.toml
slack-coder: The pattern matching has been re-implemented in Rust with good testing https://github.com/radicle-dev/heartwood/pull/518/commits/12bbca9890fb3a855e7b837b43f7bc381018c1e8

radicle-cli/src/main.rs
```diff
@@ -99,7 +104,6 @@ fn run(command: Command) -> Result<(), Option<anyhow::Error>> {
 
     Ok(())
 }
-
```
cloudhead, line 102: Undo?

radicle-cli/src/main.rs
```diff
@@ -346,3 +350,84 @@ fn run_other(exe: &str, args: &[OsString]) -> Result<(), Option<anyhow::Error>>
...
+        }
+        for v in vs {
+            if v.is_empty() || v.contains(|c: char| !(c.is_ascii_alphanumeric() || c == '-')) {
+                return false;
+            }
+        }
+        true
+    }
+
+    // https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions
```
cloudhead, line 372: Can you make this a doc comment?

radicle-cli/src/main.rs
```diff
@@ -346,3 +351,84 @@ fn run_other(exe: &str, args: &[OsString]) -> Result<(), Option<anyhow::Error>>
...
+            !is_semantic_version("1.0.0-beta+exp.sha.5114f85."),
+            "dot separated value must be non-empty"
+        );
+        assert!(
+            !is_semantic_version("1.0.0-alpha+001+002"),
+            "only one '+' allowed"
+        );
+    }
+
+    /// Ensure out version output is consistent for consumption by third parties.
```
cloudhead, line 420:
```suggestion
    /// Ensure version output is consistent for consumption by third parties.
```

cloudhead:
Merged.

PR #519, opened 2023-04-05
Do not swallow errors in rad ls

adaszko:


radicle-cli/src/commands/ls.rs
```diff
@@ -47,18 +47,37 @@ pub fn run(_options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+                continue;
+            }
+        };
+        let proj = match repo.project_of(profile.id()) {
+            Ok(proj) => proj,
+            Err(e) => {
+                eprintln!("{}", e);
+                continue;
+            }
+        };
```
cloudhead, line 71: Nah this needs more thought, it will be very annoying as-is if users don't know how to get rid of these errors.

FintanH: Could we have a  function `projects` that lists projects with `impl Iterator<Item = Result<ProjectInfo, Error>>`. In the `for_each` we could then `term::warn` that a project is broken if it's an error and `--verbose` is used.

adaszko: > Could we have a function `projects` that lists projects with `impl Iterator<Item = Result<ProjectInfo, Error>>`. In the `for_each` we could then `term::warn` that a project is broken if it's an error and `--verbose` is used.

This patch has been merged via Radicle already, so what you’re proposing would require a new one

FintanH: Cool, can you point me to the relevant patch, please?

PR #533, opened 2023-04-18
cli: add verbose flag to rad ls

FintanH:
A repository may fail to load when trying to run `rad ls`, but the failure will be hidden from the output.

Add a `--verbose` flag to add warning information to stdout when a repository fails to load.

### DIFF
--- a/radicle-cli/src/commands/ls.rs
+++ b/radicle-cli/src/commands/ls.rs
@@ -18,40 +18,69 @@ Usage
 
 Options
 
-    --help    Print help
+    --versbose, -v  Verbose output
+    --help          Print help
 "#,
 };
 
-pub struct Options {}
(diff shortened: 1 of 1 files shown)


cloudhead:
I've merged the other ls patch first, could you fix conflicts and maybe propose this as a patch on radicle?

FintanH:
Published here instead https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/1e8226b441e7beda2c817d76b196405fa0ac7046

Discussion here https://radicle.zulipchat.com/#narrow/stream/383670-patches/topic/cli.3A.20add.20verbose.20flag.20to.20rad.20ls

PR #534, opened 2023-04-18
radicle-httpd: Add cob state to listing endpoints

sebastinez:
This commit adds the cob state of either patches or issues to the corresponding endpoint querystring

So we can either query a cob listing with the required state or expect the default state.
And once filtered by cob state we paginate.

### DIFF
--- a/radicle-httpd/src/api.rs
+++ b/radicle-httpd/src/api.rs
@@ -9,7 +9,7 @@ use axum::http::Method;
 use axum::response::{IntoResponse, Json};
 use axum::routing::get;
 use axum::Router;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::json;
 use tokio::sync::RwLock;
 use tower_http::cors::{self, CorsLayer};
@@ -113,6 +113,29 @@ async fn root_handler() -> impl IntoResponse {
 pub struct PaginationQuery {
     pub page: Option<usize>,
     pub per_page: Option<usize>,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct CobsQuery<T> {
+    pub page: Option<usize>,
+    pub per_page: Option<usize>,
+    #[serde(default)]
+    #[serde(deserialize_with = "parse_state")]
+    #[serde(bound(deserialize = "T: std::str::FromStr, T::Err: std::fmt::Display"))]
+    pub state: Option<T>,
+}
+
+fn parse_state<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
+where
+    T: std::str::FromStr,
+    T::Err: std::fmt::Display,
+    D: Deserializer<'de>,
+{
+    let state: String = Deserialize::deserialize(deserializer)?;
+    T::from_str(&state)
+        .map(Some)
+        .map_err(serde::de::Error::custom)
 }
 
 mod project {
--- a/radicle-httpd/src/api/v1/projects.rs
+++ b/radicle-httpd/src/api/v1/projects.rs
@@ -21,7 +21,7 @@ use radicle_surf::{Glob, Oid, Repository};
 
 use crate::api::error::Error;
 use crate::api::project::Info;
-use crate::api::{self, Context, PaginationQuery};
+use crate::api::{self, CobsQuery, Context, PaginationQuery};
 use crate::axum_extra::{Path, Query};
 
 const CACHE_1_HOUR: &str = "public, max-age=3600, must-revalidate";
@@ -416,15 +416,30 @@ async fn readme_handler(
 async fn issues_handler(
     State(ctx): State<Context>,
     Path(project): Path<Id>,
-    Query(qs): Query<PaginationQuery>,
+    Query(qs): Query<CobsQuery<issue::State>>,
 ) -> impl IntoResponse {
-    let PaginationQuery { page, per_page } = qs;
+    let CobsQuery {
+        page,
+        per_page,
+        state,
+    } = qs;
     let page = page.unwrap_or(0);
     let per_page = per_page.unwrap_or(10);
+    let state = state.unwrap_or_default();
     let storage = &ctx.profile.storage;
     let repo = storage.repository(project)?;
     let issues = issue::Issues::open(&repo)?;
-    let mut issues: Vec<_> = issues.all()?.filter_map(|r| r.ok()).collect::<Vec<_>>();
+    let mut issues: Vec<_> = issues
+        .all()?
+        .filter_map(|r| {
+            let (id, issue, clock) = r.ok()?;
+            if issue.state() == &state {
+                Some((id, issue, clock))
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<_>>();
     issues.sort_by(|(_, a, _), (_, b, _)| b.timestamp().cmp(&a.timestamp()));
     let issues = issues
         .into_iter()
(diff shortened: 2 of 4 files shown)


radicle/src/cob/issue.rs
```diff
@@ -59,6 +59,21 @@ pub enum State {
...
+                reason: CloseReason::Other,
+            },
+            "closed_solved" => Self::Closed {
+                reason: CloseReason::Solved,
+            },
+            "open" => Self::Open,
+            _ => Self::default(),
+        }
+    }
+}
```
FintanH, line 75: I would say the same here, we should use `FromStr`.

It irks me that the `Display` and `FromStr` will differ though, so we can't say `state.to_string().parse() == state`. Not sure what a good solution would be though.

radicle-httpd/src/api.rs
```diff
@@ -115,6 +115,14 @@ pub struct PaginationQuery {
     pub per_page: Option<usize>,
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct CobsQuery {
+    pub page: Option<usize>,
+    pub per_page: Option<usize>,
+    pub state: Option<String>,
```
FintanH, line 123: I'm wondering if this should be:
```suggestion
    pub state: Option<State>,
```

You wouldn't need the `From` or `FromStr` and leave it up to `Deserialize` to handle the conversion.

radicle/src/cob/issue.rs
```diff
@@ -59,6 +59,21 @@ pub enum State {
     Open,
 }
 
+impl From<String> for State {
+    fn from(s: String) -> Self {
+        match s.as_str() {
+            "closed_other" => Self::Closed {
+                reason: CloseReason::Other,
+            },
+            "closed_solved" => Self::Closed {
```
cloudhead, line 68: I think we use camelCase in the URL?

radicle-httpd/src/api.rs
sebastinez: Yeah thats what I also wanted to do in the beginning, but since we serialize `State` to `{ "status": "open" }` I would need to pass that through a querystring which is not what I was after.

My goal was to pass e.g. `?state=open` and for the endpoint to parse it into `patch::State::Open`
Also wrote a custom deserializer, but in the end I thought it to be easier to convert the passed `String` into `State`

radicle/src/cob/issue.rs
sebastinez: I think we made the case at some point to use kebap-case for querystring params. Like we do for pagination `?per-page=100`.

radicle-httpd/src/api.rs
FintanH: Wait, how does the deserializer work? :thinking: does it not accept `?state=open`? I'm not sure I'm understanding what the problem is

sebastinez: as far as I understand, the serializer serializes `patch::State::Open` into `{ "status": "open" }` due to the serde container tag attribute on `State` `#[serde(rename_all = "camelCase", tag = "status")]`
for reference: https://serde.rs/container-attrs.html#tag
So the deserializer expects a `{ "status": "open" }` to be able to parse it back to `patch::State::Open`
Or maybe I'm missing something with serde.

FintanH: Ahhh I see... Ya that's annoying. I tried googling for any solutions to this, but I can't seem to see any. It might be worth asking in the `axum` project if there's a way of doing this

radicle/src/cob/patch.rs
```diff
@@ -530,6 +530,18 @@ pub enum State {
...
+    fn from(s: String) -> Self {
+        match s.as_str() {
+            "open" => Self::Open,
+            "draft" => Self::Draft,
+            "merged" => Self::Merged,
+            "archived" => Self::Archived,
+            _ => Self::default(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
sebastinez: Ok was able to add a `Deserializer` to convert the required querystring into the `State` enum
6ec2614

radicle-httpd/src/api.rs
sebastinez: Ok I think this commit does the trick
6ec2614

radicle/src/cob/issue.rs
sebastinez: 6ec2614

radicle/src/cob/patch.rs
```diff
@@ -530,6 +530,22 @@ pub enum State {
...
+        let s = String::deserialize(deserializer)?.to_lowercase();
+        match s.as_str() {
+            "open" => Ok(State::Open),
+            "draft" => Ok(State::Draft),
+            "archived" => Ok(State::Archived),
+            "merged" => Ok(State::Merged),
+            _ => Err(serde::de::Error::custom("Unrecognised patch state")),
+        }
+    }
+}
```
FintanH, line 547: Sorry, I don't think this is any better. Now `Deserialize` isn't the inverse of `Serialize` which seems like a bad idea. The `FromStr` version is a better route or we figure out if `axum` can handle tagged enums.

sebastinez: Okay I understand, will keep looking 😎

radicle-httpd/src/api.rs
FintanH: I've asked them here https://github.com/tokio-rs/axum/discussions/1945

radicle/src/cob/issue.rs
```diff
@@ -49,7 +49,7 @@ pub enum CloseReason {
 }
 
 /// Issue state.
-#[derive(Debug, Default, Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Default, Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Serialize)]
 #[serde(rename_all = "camelCase", tag = "status")]
```
cloudhead, line 53: So which is it? camelCase or kebab-case?

sebastinez: It's camelCase, will take it into account, once I found a solution how to deserialize the tagged enum 🫡

radicle-httpd/src/api.rs
FintanH: Ok, I think we can do the following:
* Implement `FromStr` for the `State` operations
* Change this field to `pub state: Option<State>`
* Add the pragma `#[serde(deserialize_with="from_str")]`

Could you see if that works? I'm unsure if the `Option` will be a problem with the pragma or not

sebastinez: Ok I tried that and it seems that `Option` is a problem with the pragma.
And IMO by doing that we also change the `Deserialize` of `State` which isn't inverse to `Serialize` anymore.

So either we accept strings and parse them, or write our own Deserializer.

sebastinez: Or we could also remove the `tag = "status"` which I don't think makes much sense now..
If at some point we want to show more information under the `state` property we could certainly revisit this, since we use `serde` mostly for serialization for `radicle-httpd` right?

FintanH: >  And IMO by doing that we also change the Deserialize of State which isn't inverse to Serialize anymore.

It wouldn't change the `Deserialize` of `State`, it would be specifying that `QueryCobs` is using a different way of deserializing `State` :)

 > So either we accept strings and parse them, or write our own Deserializer.
 
 Thanks for trying, I'd say we go with accept strings and use `FromStr`. I was hoping this version would work. A very last alternative is to define a standalone function:
 ```rust
 fn parse_state(state: String) -> Result<Option<State>, String> {
     if state.is_empty() { None } else { state.parse().map(Some).map_err(|err| err.to_string())
 }
 ```
and use `#[serde(deserialize_with = "parse_state")]`

> Or we could also remove the tag = "status" which I don't think makes much sense now..

Aye, why did you opt for this @cloudhead? Removing it now will break backwards compatibility though and we're now in production ;)

FintanH: > > So either we accept strings and parse them, or write our own Deserializer.
> 
> Thanks for trying, I'd say we go with accept strings and use `FromStr`. I was hoping this version would work. A very last alternative is to define a standalone function:
> 
> ```rust
> fn parse_state(state: String) -> Result<Option<State>, String> {
>     if state.is_empty() { None } else { state.parse().map(Some).map_err(|err| err.to_string())
> }
> ```
> 
> and use `#[serde(deserialize_with = "parse_state")]`

Actually, this is quite nice. It's basically equivalent to the parsing solution but we localise the parsing at the point of deserialization instead of in the handler :)

radicle-httpd/src/api/v1/projects.rs
```diff
@@ -416,15 +416,30 @@ async fn readme_handler(
...
-    let mut issues: Vec<_> = issues.all()?.filter_map(|r| r.ok()).collect::<Vec<_>>();
+    let mut issues: Vec<_> = issues
+        .all()?
+        .filter_map(|r| {
+            let (id, issue, clock) = r.ok()?;
+            if issue.state() == &state {
+                Some((id, issue, clock))
+            } else {
+                None
+            }
```
FintanH, line 440: This pattern is captured in a method called `then_some`, if you're interested in using it :)
```suggestion
            (issue.state() == &state).then_some((id, issue, clock))
```

radicle-httpd/src/api.rs
```diff
@@ -115,6 +115,29 @@ pub struct PaginationQuery {
...
 
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct CobsQuery<T> {
+    pub page: Option<usize>,
+    pub per_page: Option<usize>,
+    #[serde(default)]
+    #[serde(deserialize_with = "parse_state")]
+    #[serde(bound(deserialize = "T: std::str::FromStr, T::Err: std::fmt::Display"))]
+    pub state: Option<T>,
```
FintanH, line 126: Nice :)

FintanH (commented):
Lurvly

sebastinez:
Has been merged in the radicle repo

PR #535, opened 2023-04-19
tui: module refactoring

erak:


cloudhead:
Could you open this on radicle instead?

cloudhead:
Merged in radicle.

PR #539, opened 2023-04-24
rad-patch: Fix patch open/upate when out-of-sync

slack-coder:
Trying to open or update Patch will fail if the user's working copy if out of sync with the Radicle projects canonical path.

Fix this by using an anonymous fetch to import the required commits while avoiding affecting the user's working copy git references.  The approach is already used for `rad-merge`.

### DIFF
--- a/radicle-cli/src/commands/patch/common.rs
+++ b/radicle-cli/src/commands/patch/common.rs
@@ -37,9 +37,10 @@ fn get_branch(git_ref: git::Qualified) -> git::RefString {
     std::iter::once(head).chain(tail).collect()
 }
 
-/// Determine the merge target for this patch. This can ben any tracked remote's "default"
-/// branch, as well as your own (eg. `rad/master`).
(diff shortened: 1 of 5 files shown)


radicle-cli/src/commands/patch/common.rs
```diff
@@ -51,6 +52,7 @@ pub fn get_merge_target(
         anyhow::bail!("commits are already included in the target branch; nothing to do");
     }
 
+    crate::git::anonymous_fetch(workdir, &git::Url::from(storage.id), &target_oid)?;
```
FintanH, line 55: It's probably not a big deal, but is it necessary to always fetch?

FintanH (approved):
LGTM

radicle-cli/src/commands/patch/common.rs
slack-coder: True

PR #544, opened 2023-04-27
cli: test the deletion of cobs

FintanH:
Ensure that the deletion of cobs is respected by testing a scenario where a peer creates and then deletes an issue, while another peer fetches and also has the issue be present and then removed.

### DIFF
--- a/radicle-cli/tests/commands.rs
+++ b/radicle-cli/tests/commands.rs
@@ -552,6 +552,53 @@ fn test_cob_replication() {
 }
 
 #[test]
+fn test_cob_deletion() {
+    let mut environment = Environment::new();
+    let working = tempfile::tempdir().unwrap();
+    let mut alice = environment.node("alice");
+    let bob = environment.node("bob");
+
+    let rid = alice.project("heartwood", "");
+
+    let mut alice = alice.spawn(Config::default());
+    let mut bob = bob.spawn(Config::default());
+
+    alice.handle.track_repo(rid, Scope::All).unwrap();
+    bob.handle.track_repo(rid, Scope::All).unwrap();
+    alice.connect(&bob);
+    bob.routes_to(&[(rid, alice.id)]);
+
+    let alice_repo = alice.storage.repository(rid).unwrap();
+    let mut alice_issues = radicle::cob::issue::Issues::open(&alice_repo).unwrap();
+    let issue = alice_issues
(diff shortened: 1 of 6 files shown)


cloudhead:
By the way I was thinking about this, and: if a peer deletes a COB that he hasn't interacted with, he shouldn't `sign-refs`, because there will be nothing new to sign, right?

FintanH:
> By the way I was thinking about this, and: if a peer deletes a COB that he hasn't interacted with, he shouldn't `sign-refs`, because there will be nothing new to sign, right?

Ya, she wouldn't have any references to the COB, right?

cloudhead:
Exactly, so somehow in the `delete` code, we should check before calling `sign_refs`, to not create a commit.

FintanH:
> Exactly, so somehow in the `delete` code, we should check before calling `sign_refs`, to not create a commit.

Hmmm, I'm not sure I follow. Is this to avoid needlessly creating a new set of signed refs?

cloudhead:
Yes exactly, it would be identical to the previous one.

FintanH:
> Yes exactly, it would be identical to the previous one.

I guess the only thing is that it would come at the cost of loading the `sigrefs` first. Do you think that's worth it?

cloudhead:
I'm not sure we'd need to load sigrefs. I think we can just check if a ref exists under the remote, for that cob.

FintanH:
> I'm not sure we'd need to load sigrefs. I think we can just check if a ref exists under the remote, for that cob.

Ah ya that's fair. It should be equivalent

cloudhead:
Hmm good question, is `cob::remove` actually a no-op if you don't have your own refs, or does it remove from all remotes? If the former, then we don't need to call it.

FintanH:
> Hmm good question, is `cob::remove` actually a no-op if you don't have your own refs, or does it remove from all remotes? If the former, then we don't need to call it.

So it just deletes the reference, so it would make sense to not remove it if it doesn't exist :) I'll add that to the change as well

FintanH:
Closing in favour of https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/553c795ab9a8b0a7c8e0dfec95568ba11c018ad2?tab=files

PR #546, opened 2023-05-08
Produce debug binaries for testing radicle-interface

adaszko:


.github/workflows/release.yml
```diff
@@ -85,19 +85,19 @@ jobs:
...
+          rustup target add ${{ matrix.target }}
+          cargo build --target ${{ matrix.target }} --package=radicle-httpd --release
           cp target/${{ matrix.target }}/release/radicle-httpd "$staging"/
+          cargo build --target ${{ matrix.target }} --package=radicle-node --release
           cp target/${{ matrix.target }}/release/radicle-node "$staging"/
+          cargo build --target ${{ matrix.target }} --bin rad --release
           cp target/${{ matrix.target }}/release/rad "$staging"/
+          cargo build --target=${{ matrix.target }} --bin rad
+          cp target/${{ matrix.target }}/debug/rad "$staging"/rad-debug
+          cargo build --target ${{ matrix.target }} --bin git-remote-rad --release
```
cloudhead, line 100: Could we perhaps split these into two build steps? Relase and Debug?

cloudhead: Really? Can't we just have it as a variable in the build matrix?

adaszko: We’d need a bit more than a variable: a conditional, and those are messy with the build matrix

sebastinez:
@cloudhead @adaszko Can we get this PR merged? We need the radicle debug binary for every new commit that gets deployed. So we don't get stuck on one specific.

PR #549, opened 2023-05-09
Deploy to geoavailable seed node

adaszko:


cloudhead:
I've moved this to a manual deploy script here: https://github.com/radicle-dev/heartwood/blob/master/scripts/deploy-seed

PR #555, opened 2023-10-14
Use bash from env for better portability

Luz:
Hi
According to [Stackoverflow](https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang#10383546) it looks like the portability of the script could be improved by having this new style for the bash shebang.
(Otherwise you might also consider using `#!/bin/sh` ?)

### DIFF
--- a/scripts/import-issue.sh
+++ b/scripts/import-issue.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Import a GitHub issue into Radicle.
 #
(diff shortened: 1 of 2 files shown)


FintanH:
Hey @Luz, thanks for the contribution!

Just an fyi, we've moved development to using our own forge https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5 :)

This looks good to me, but I'll let @cloudhead do the merging

PR #556, opened 2023-11-12
support multiple fetches in a row

Byron:
> [!NOTE]
> * Related issue in `gitoxide`: https://github.com/Byron/gitoxide/issues/972
> * Related PR with improvement in `gix`: https://github.com/Byron/gitoxide/pull/1107

The server implementation here is the first which actually used the capabilities of the V2
git protocol to perform multiple pack-fetches in a row through the same connection, which
isn't currently tested in `gitoxide`.

Thus it ran into the issue that the pack-writer will read a pack 'perfectly', i.e. without
encountering EOF, which would be encoded by the FLUSH packet (`0000` on the wire).

For that reason the command run after a pack was received would still see the flush packet
and immediately assume it's done.

As the pack-resolver for good reason doesn't know anything about this, it's up to the
fetch implementation to assure the transport has been drained of any remaining bytes.

### Review Notes

* The implementation plays it very safe with 'all-assertions'. I can imagine this to be a bit too much.
* I left a TODO to get rid fo the Delegate-style implementation which I failed to use myself in `gix`. 
  This is an opportunity for improvements, but needs communication to figure out how this could look like.
* I didn't see CI kicking in, which I tried to check while this was a draft PR. However, locally on MacOS I have 61 failed tests, and validated that it didn't get worse. `cargo test -p radicle-node tests::e2e` failed before the fix (and after removing the previous fix), and works now, so I am optimistic nothing is broken by it.

### DIFF
--- a/radicle-fetch/src/transport/fetch.rs
+++ b/radicle-fetch/src/transport/fetch.rs
@@ -12,7 +12,12 @@ use gix_protocol::{
     handshake::{self, Ref},
     ls_refs, FetchConnection,
 };
-use gix_transport::{bstr::BString, client, Protocol};
+use gix_transport::{
+    bstr::BString,
+    client,
+    client::{ExtendedBufRead, MessageKind},
+    Protocol,
+};
 
 use super::{agent_name, indicate_end_of_interaction, Connection, WantsHaves};
 
@@ -252,11 +257,24 @@ where
             //       needed as it needs `'static`. As the top-level seems to pass `Discard`,
             //       there should be no repercussions right now.
             (&mut delegate).receive_pack(
-                reader,
+                &mut reader,
                 progress.add_child("receiving pack"),
                 &[],
                 &response,
             )?;
+            assert_eq!(
+                reader.stopped_at(),
+                None,
+                "packs are read without 'overshooting', hence it never encountered EOF"
+            );
+            // Consume anything that might still be left on the wire - this is 'EOF' most of the time,
+            // but some tests have 'garbage' here as well.
+            std::io::copy(&mut reader, &mut std::io::sink())?;
+            assert_eq!(
+                reader.stopped_at(),
+                Some(MessageKind::Flush),
+                "the flush packet was now consumed"
+            );
             break 'negotiation;
         } else {
             match action {
@@ -269,33 +287,6 @@ where
         && matches!(conn.mode, FetchConnection::TerminateOnSuccessfulCompletion)
     {
         indicate_end_of_interaction(&mut conn)?;
-    }
-
-    // N.b. the flush packet is never read in the packfile parser. To
-    // remedy this, we read from our connection until we see it ensure
-    // that there is no leftover data.
-    let (mut r, _) = conn.inner.into_inner();
-    let mut buf = [0; 4096];
-
-    loop {
-        match r.read(&mut buf)? {
-            0 => break,
-            n => match std::str::from_utf8(&buf[..n]) {
-                Ok(progress) => {
-                    let lines = progress.split('\n');
-                    if lines.into_iter().any(|line| line == "0000") {
-                        break;
-                    }
-                }
-                Err(e) => {
-                    return Err(std::io::Error::new(
-                        std::io::ErrorKind::Other,
-                        format!("found leftover packfile bytes: {e}"),
-                    )
-                    .into());
-                }
-            },
-        }
     }
 
     log::trace!(target: "fetch", "fetched refs: {:?}", delegate.out.refs);
--- a/radicle-fetch/src/transport/fetch.rs
+++ b/radicle-fetch/src/transport/fetch.rs
@@ -5,7 +5,7 @@ use std::{
     sync::{atomic::AtomicBool, Arc},
 };
 
-use gix_features::progress::Progress;
+use gix_features::progress::NestedProgress;
 use gix_pack as pack;
 use gix_protocol::{
     fetch::{self, Delegate, DelegateBlocking},
@@ -45,11 +45,15 @@ pub struct PackWriter {
 impl PackWriter {
     /// Write the packfile read from `pack` to the `objects/pack`
     /// directory.
-    pub fn write_pack(
+    pub fn write_pack<P>(
         &self,
-        pack: impl BufRead,
-        progress: impl Progress,
-    ) -> Result<pack::bundle::write::Outcome, error::PackWriter> {
+        mut pack: impl BufRead,
+        mut progress: P,
+    ) -> Result<pack::bundle::write::Outcome, error::PackWriter>
+    where
+        P: NestedProgress,
+        P::SubProgress: 'static,
+    {
         use gix_odb::FindExt as _;
 
         let options = pack::bundle::write::Options {
@@ -68,16 +72,16 @@ impl PackWriter {
         };
         let thickener = Arc::new(gix_odb::Store::at_opts(
             self.git_dir.join("objects"),
-            [],
+            &mut [].into_iter(),
             odb_opts,
         )?);
         let thickener = thickener.to_handle_arc();
         Ok(pack::Bundle::write_to_directory(
-            pack,
-            Some(self.git_dir.join("objects").join("pack")),
-            progress,
+            &mut pack,
+            Some(&self.git_dir.join("objects").join("pack")),
+            &mut progress,
             &self.interrupt,
-            Some(Box::new(move |oid, buf| thickener.find(oid, buf).ok())),
+            Some(Box::new(move |oid, buf| thickener.find(&oid, buf).ok())),
             options,
         )?)
     }
@@ -104,7 +108,7 @@ impl<'a> Delegate for &'a mut Fetch {
     fn receive_pack(
         &mut self,
         input: impl io::BufRead,
-        progress: impl Progress,
+        progress: impl NestedProgress + 'static,
         _refs: &[handshake::Ref],
         previous_response: &fetch::Response,
     ) -> io::Result<()> {
@@ -184,7 +188,7 @@ pub(crate) fn run<P, R, W>(
     progress: &mut P,
 ) -> Result<FetchOut, Error>
 where
-    P: Progress,
+    P: NestedProgress,
     P::SubProgress: 'static,
     R: io::Read,
     W: io::Write,
@@ -241,11 +245,18 @@ where
         let response = fetch::Response::from_line_reader(*protocol, &mut reader, true)?;
         previous_response = if response.has_pack() {
             progress.step();
-            progress.set_name("receiving pack");
             if !sideband_all {
                 setup_remote_progress(progress, &mut reader);
             }
-            (&mut delegate).receive_pack(reader, progress, &[], &response)?;
+            // TODO: remove delegate in favor of functional style to fix progress-hack,
+            //       needed as it needs `'static`. As the top-level seems to pass `Discard`,
+            //       there should be no repercussions right now.
+            (&mut delegate).receive_pack(
+                reader,
+                progress.add_child("receiving pack"),
+                &[],
+                &response,
+            )?;
             break 'negotiation;
         } else {
             match action {
@@ -295,7 +306,7 @@ fn setup_remote_progress<P>(
     progress: &mut P,
     reader: &mut Box<dyn gix_transport::client::ExtendedBufRead + Unpin + '_>,
 ) where
-    P: Progress,
+    P: NestedProgress,
     P::SubProgress: 'static,
 {
     reader.set_progress_handler(Some(Box::new({
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,7 +45,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.10",
  "once_cell",
  "version_check",
 ]
@@ -1014,16 +1013,16 @@ dependencies = [
 
 [[package]]
 name = "gix-actor"
-version = "0.23.0"
+version = "0.28.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1969b77b9ee4cc1755c841987ec6f7622aaca95e952bcafb76973ae59d1b8716"
+checksum = "948a5f9e43559d16faf583694f1c742eb401ce24ce8e6f2238caedea7486433c"
 dependencies = [
  "bstr",
  "btoi",
  "gix-date",
  "itoa",
- "nom",
  "thiserror",
+ "winnow",
 ]
 
 [[package]]
@@ -1037,22 +1036,22 @@ dependencies = [
 
 [[package]]
 name = "gix-command"
-version = "0.2.9"
+version = "0.2.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0f28f654184b5f725c5737c7e4f466cbd8f0102ac352d5257eeab19647ee4256"
+checksum = "3c576cfbf577f72c097b5f88aedea502cd62952bdc1fb3adcab4531d5525a4c7"
 dependencies = [
  "bstr",
 ]
 
 [[package]]
 name = "gix-commitgraph"
-version = "0.17.1"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ed42baa50075d41c1a0931074ce1a97c5797c7c6fe7591d9f1f2dcd448532c26"
+checksum = "7e8bc78b1a6328fa6d8b3a53b6c73997af37fd6bfc1d6c49f149e63bda5cbb36"
 dependencies = [
  "bstr",
  "gix-chunk",
- "gix-features 0.31.1",
+ "gix-features",
  "gix-hash",
  "memmap2",
  "thiserror",
@@ -1060,9 +1059,9 @@ dependencies = [
 
 [[package]]
 name = "gix-config-value"
-version = "0.12.5"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e874f41437441c02991dcea76990b9058fadfc54b02ab4dd06ab2218af43897"
+checksum = "ea7505b97f4d8e7933e29735a568ba2f86d8de466669d9f0e8321384f9972f47"
 dependencies = [
  "bitflags 2.4.0",
  "bstr",
@@ -1073,9 +1072,9 @@ dependencies = [
 
 [[package]]
 name = "gix-credentials"
-version = "0.16.1"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "75a75565e0e6e7f80cfa4eb1b05cc448c6846ddd48dcf413a28875fbc11ee9af"
+checksum = "1c5c5d74069b842a1861e581027ac6b7ad9ff66f5911c89b9f45484d7ebda6a4"
 dependencies = [
  "bstr",
  "gix-command",
@@ -1089,9 +1088,9 @@ dependencies = [
 
 [[package]]
 name = "gix-date"
-version = "0.7.4"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0a825babda995d788e30d306a49dacd1e93d5f5d33d53c7682d0347cef40333c"
+checksum = "fc7df669639582dc7c02737642f76890b03b5544e141caba68a7d6b4eb551e0d"
 dependencies = [
  "bstr",
  "itoa",
@@ -1101,21 +1100,20 @@ dependencies = [
 
 [[package]]
 name = "gix-diff"
-version = "0.32.0"
+version = "0.37.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aaf5d9b9b521b284ebe53ee69eee33341835ec70edc314f36b2100ea81396121"
+checksum = "931394f69fb8c9ed6afc0aae3487bd869e936339bcc13ed8884472af072e0554"
 dependencies = [
  "gix-hash",
  "gix-object",
- "imara-diff",
  "thiserror",
 ]
 
 [[package]]
 name = "gix-features"
-version = "0.31.1"
+version = "0.36.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "06142d8cff5d17509399b04052b64d2f9b3a311d5cff0b1a32b220f62cd0d595"
+checksum = "51f4365ba17c4f218d7fd9ec102b8d2d3cb0ca200a835e81151ace7778aec827"
 dependencies = [
  "crc32fast",
  "flate2",
@@ -1129,40 +1127,29 @@ dependencies = [
 ]
 
 [[package]]
-name = "gix-features"
-version = "0.32.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "882695cccf38da4c3cc7ee687bdb412cf25e37932d7f8f2c306112ea712449f1"
-dependencies = [
- "gix-hash",
- "gix-trace",
- "libc",
-]
-
-[[package]]
 name = "gix-fs"
-version = "0.4.1"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4d5b6e9d34a2c61ea4a02bbca94c409ab6dbbca1348cbb67298cd7fed8758761"
+checksum = "8cd171c0cae97cd0dc57e7b4601cb1ebf596450e263ef3c02be9107272c877bd"
 dependencies = [
- "gix-features 0.32.1",
+ "gix-features",
 ]
 
 [[package]]
 name = "gix-hash"
-version = "0.11.4"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4b422ff2ad9a0628baaad6da468cf05385bf3f5ab495ad5a33cce99b9f41092f"
+checksum = "1884c7b41ea0875217c1be9ce91322f90bde433e91d374d0e1276073a51ccc60"
 dependencies = [
- "hex",
+ "faster-hex",
  "thiserror",
 ]
 
 [[package]]
 name = "gix-hashtable"
-version = "0.2.4"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "385f4ce6ecf3692d313ca3aa9bd3b3d8490de53368d6d94bedff3af8b6d9c58d"
+checksum = "409268480841ad008e81c17ca5a293393fbf9f2b6c2f85b8ab9de1f0c5176a16"
 dependencies = [
  "gix-hash",
  "hashbrown 0.14.0",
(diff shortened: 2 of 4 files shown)


FintanH:
> I didn't see CI kicking in, which I tried to check while this was a draft PR. However, locally on MacOS I have 61 failed tests, and validated that it didn't get worse. cargo test -p radicle-node tests::e2e failed before the fix (and after removing the previous fix), and works now, so I am optimistic nothing is broken by it.

What kind of failures were you running into?

On this branch if I run `cargo test -p radilce-node -- --nocapture` or `cargo test -p radicle-cli -- --nocapture` will fail by panicking at one of the assertions:
```
thread 'z6Mkveba9Mgj6YrpTJpm68YtrjAaZBaPzcjmGX6h7YhfBzD5 <worker#1>    ' panicked at 'the flush packet indicates there is nothing to peek', /home/haptop/Developer/heartwood/radicle-fetch/src/transport/fetch.rs:270:13
```

Byron:
> What kind of failures were you running into?

I think the source of all my 26 errors is `Io(Error { kind: InvalidInput, message: "path must be shorter than SUN_LEN" })`, and I didn't look into what that could mean yet. That's also the reason I didn't see the legitimate error, from what I can tell the `SUN_LEN` error happens before that when running `cargo test -p radicle-cli -- --nocapture`.

It's like I am missing some initial setup.

> ```
> thread 'z6Mkveba9Mgj6YrpTJpm68YtrjAaZBaPzcjmGX6h7YhfBzD5 <worker#1>    ' panicked at 'the flush packet indicates there is nothing to peek', /home/haptop/Developer/heartwood/radicle-fetch/src/transport/fetch.rs:270:13
> ```


That's interesting, as this might mean something similar to what I encountered in the `gitoxide` test-suite with tests that deal with shallow clones. There it would end up seeing two bytes, `1` and `30` and didn't investigate what these might be.

This failure means there is more than expected left on the wire. In `gitoxide` I just ended up consuming all of it. For now I have adjusted the code blindly (but with working `e2e` tests) to implement that. However I plan to investigate the left-over bytes the one test is encountering once I can run it.

I have force-pushed the adjustment in case you want to give it another try.

radicle-fetch/src/transport/fetch.rs
```diff
@@ -241,11 +250,31 @@ where
...
+                &response,
+            )?;
+            assert_eq!(
+                reader.stopped_at(),
+                None,
+                "packs are read without 'overshooting', hence it never encountered EOF"
+            );
+            // Consume anything that might still be left on the wire - this is 'EOF' most of the time,
+            // but some tests have 'garbage' here as well.
+            std::io::copy(&mut reader, &mut std::io::sink())?;
```
FintanH, line 272: I suppose this is quite similar to how I was doing it before, except we're not bothering to check if it's the flush packet or not.

fwiw, I tried running the test suite with `cargo test --all` and it worked :+1:

FintanH:
> I think the source of all my 26 errors is Io(Error { kind: InvalidInput, message: "path must be shorter than SUN_LEN" }), and I didn't look into what that could mean yet. That's also the reason I didn't see the legitimate error, from what I can tell the SUN_LEN error happens before that when running cargo test -p radicle-cli -- --nocapture.

Hmmmm what OS are you using? From a quick look around it seems to be the Radicle socket paths are too long on your system https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars

radicle-fetch/src/transport/fetch.rs
Byron: Good to hear!

Actually it doesn't seem to always be the flush package, otherwise there wouldn't have been that one test failure. Once my tests are working, I will look at the contents that it's copying, maybe it's the same 'garbage' bytes that I have been seeing.

Byron:
It looks like BSDs, and so MacOS, have a 104 character limit. I am on MacOS 14.1.1, so probably it's that. Is there a chance to make the socket path shorter on MacOS?

I investigated real quick and it does turn out that setting `RAD_SOCKET=/tmp/rad` makes more tests pass. Those who don't are unhappy that the socket is already running. Strangely, even with when making tests single-threaded, it still fails like that as if the socket-device/file isn't removed after the test runs.

This is what I tried:

```
RAD_SOCKET=/tmp/rad_socket cargo test --all -- --test-threads=1
```

Which fails with:

```
'test_replication_via_seed' panicked at 'called `Result::unwrap()` on an `Err` value: AlreadyRunning("/tmp/rad_socket")', /Users/byron/dev/github.com/radicle-dev/heartwood/radicle-node/src/test/environment.rs:379:10
```

Even when running the tests with `RAD_SOCKET=/tmp/rad_socket cargo nextest run -j1 --no-fail-fast`, which typically does a better job at coherently running things, it still stumbles over the socket.

When looking at a detailed error log, at least it becomes clear why the sockets are too long - I presume it would create them in the temporary directory…

```
running 1 test
test: Node::init alice: z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y
test: Node::init bob: z6MkpLDPEXJAdgnZ2xLsJFotNNZCw4CcFEgddpZhRWzT8g6E
1699888427 tests::e2e::test_replication_ref_in_sigrefs radicle_cob::change_graph: loading object 'xyz.radicle.id' '7d16f93be9371c051505352f34c150fe5d12bdc9'
1699888427 tests::e2e::test_replication_ref_in_sigrefs   storage: Setting ref: refs/rad/id -> 7d16f93be9371c051505352f34c150fe5d12bdc9
1699888427 tests::e2e::test_replication_ref_in_sigrefs   storage: Setting ref: refs/heads/master -> f2de534b5e81d7c6e2dcaf58c3dd91573c0a0354
1699888427 tests::e2e::test_replication_ref_in_sigrefs   storage: Setting ref: HEAD -> refs/heads/master
test: Initialized project rad:z2wxtY8oFfoNhJrX84GKCEqmnjo5j for node z6MkpLDPEXJAdgnZ2xLsJFotNNZCw4CcFEgddpZhRWzT8g6E
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Opening address book /private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/hL7Ixmp6/node/addresses.db..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Opening gossip store from /private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/hL7Ixmp6/node/addresses.db..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Opening routing table /private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/hL7Ixmp6/node/routing.db..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Opening tracking policy table /private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/hL7Ixmp6/node/tracking.db..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Default tracking policy set to 'block'
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Initializing service (Test)..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Address book is empty. Adding bootstrap nodes..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: 0 nodes added to address book
1699888427 tests::e2e::test_replication_ref_in_sigrefs   service: Init @1699888427000
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Listening on 0.0.0.0:59216..
1699888427 z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y <runtime>           node: Running node z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y in /private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/hL7Ixmp6..
1699888427 z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y <runtime>           node: Binding control socket /tmp/rad_socket..
1699888427 z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y <control>        control: Control thread listening on socket..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Opening address book /private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/rS4hfeIl/node/addresses.db..
1699888427 z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y <service>        service: Received command QueryState(..)
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Opening gossip store from /private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/rS4hfeIl/node/addresses.db..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Opening routing table /private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/rS4hfeIl/node/routing.db..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Opening tracking policy table /private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/rS4hfeIl/node/tracking.db..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Default tracking policy set to 'block'
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Initializing service (Test)..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Address book is empty. Adding bootstrap nodes..
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: 0 nodes added to address book
1699888427 tests::e2e::test_replication_ref_in_sigrefs   service: Init @1699888427000
1699888427 tests::e2e::test_replication_ref_in_sigrefs      node: Listening on 0.0.0.0:59217..
test: Node z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y shutting down..
1699888427 z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y <control>        control: Received `{"type":"shutdown"}` on control socket
1699888427 z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y <control>        control: Shutdown requested..
1699888427 z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y <runtime>           pool: Worker pool shutting down..
1699888427 z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y <runtime>           node: Node shutdown completed for z6MksUg8HWbQavnDZv6QDmtTN7UFqdysZ8t3Bg3qPRPoU68y
test tests::e2e::test_replication_ref_in_sigrefs ... FAILED

failures:

failures:
    tests::e2e::test_replication_ref_in_sigrefs

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 93 filtered out; finished in 0.13s


--- STDERR:              radicle-node tests::e2e::test_replication_ref_in_sigrefs ---
thread 'tests::e2e::test_replication_ref_in_sigrefs' panicked at 'called `Result::unwrap()` on an `Err` value: AlreadyRunning("/tmp/rad_socket")', radicle-node/src/test/environment.rs:379:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

… `/private/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/.tmpt4bJiZ/rS4hfeIl/node/a` which seems pretty long indeed :).

To make this work, I think on MacOS there would have to be a special case at least for the tests - would you like me to implement that as well? Happy to contribute this as I really do want to run the tests successfully :).

sebastinez:
Hi @Byron, so I was also running into this issue on MacOS and I solved it by defining the `TMPDIR` variable to make the path shorter.

Byron:
Thanks for the hint, @sebastinez, thanks to it I am now down to 3 failures which seem (more) legit. But since you are running MacOS as well I wonder why you (probably) don't see them.

```
--- STDERR:              radicle-cli::commands rad_diff ---
thread 'rad_diff' panicked at 'Expected success, was 1
stderr:
sed: 1: "main.c": invalid command code m

', radicle-cli-test/src/lib.rs:472:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```
--- STDERR:              radicle-cli::commands rad_patch_pull_update ---
thread 'rad_patch_pull_update' panicked at '
--- Expected
++++ actual:   stdout
   1      - {"type":"refsSynced","remote":"z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi","rid":"rad:zhbMU4DUXrzB8xT6qAJh6yZ7bFMK"}
        1 + ✗ Error: timed out reading from control socket

Exit status: 1
', radicle-cli-test/src/lib.rs:468:36
```

```
--- STDERR:              radicle-httpd api::v1::stats::routes::test_stats ---
thread 'api::v1::stats::routes::test_stats' panicked at 'called `Result::unwrap()` on an `Err` value: Git(Error { code: -1, klass: 4, message: "src refspec 'refs/heads/master' does not match any existing object" })', radicle-httpd/src/test.rs:194:6
```

The last one occours for 35 tests in `radicle-httpd`, and I only post this here for posterity as I am offering to look into making this work on MacOS if there is interest.

----

@FintanH, regarding the two failures of the previous version of the line that would assert with `peek_data_line()`…

```patch
diff --git a/radicle-fetch/src/transport/fetch.rs b/radicle-fetch/src/transport/fetch.rs
index 5dfc1535..4558df49 100644
--- a/radicle-fetch/src/transport/fetch.rs
+++ b/radicle-fetch/src/transport/fetch.rs
@@ -269,7 +269,7 @@ where
             );
             // Consume anything that might still be left on the wire - this is 'EOF' most of the time,
             // but some tests have 'garbage' here as well.
-            std::io::copy(&mut reader, &mut std::io::sink())?;
+            assert!(reader.peek_data_line().is_none());
             assert_eq!(
                 reader.stopped_at(),
                 Some(MessageKind::Flush),
```

…are not reproducing for me. I only get my expected 37 errors, but none in `radicle-node` for example, where [you definitely saw that assertion fail](https://github.com/radicle-dev/heartwood/pull/556#issuecomment-1807744629).

That's strange, except that maybe this is related to the `git` installed on the system, which probably still is the one building packs. I have this one, maybe yours is newer?

```
git version 2.39.3 (Apple Git-145)
```

When I learn about your `git` version, I can retry with that and see if it reproduces then.

FintanH:
> The last one occours for 35 tests in radicle-httpd, and I only post this here for posterity as I am offering to look into making this work on MacOS if there is interest.

For this case, it's an oversight in the test suite that I have a fix for [here](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/06e482a7a1373b08e97f7852c7a1e5af5df63b48?tab=changes). Essentially, you have a different `init.defaultBranch` value than `master` and it was assumed in the test suite that it would be `master`.

Regarding this failure:
```
--- STDERR:              radicle-cli::commands rad_patch_pull_update ---
thread 'rad_patch_pull_update' panicked at '
--- Expected
++++ actual:   stdout
   1      - {"type":"refsSynced","remote":"z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi","rid":"rad:zhbMU4DUXrzB8xT6qAJh6yZ7bFMK"}
        1 + ✗ Error: timed out reading from control socket

Exit status: 1
', radicle-cli-test/src/lib.rs:468:36
```
I actually witnessed this running on my Mac recently. We have no fix and it's an annoying race condition, so in this case it can be ignored.

However, this one is new and I'd need to dig into what's going on:
```
--- STDERR:              radicle-cli::commands rad_diff ---
thread 'rad_diff' panicked at 'Expected success, was 1
stderr:
sed: 1: "main.c": invalid command code m

', radicle-cli-test/src/lib.rs:472:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


> When I learn about your git version, I can retry with that and see if it reproduces then.

Indeed, I am on a newer version of git:
```
git version 2.40.1
```

Byron:
> For this case, it's an oversight in the test suite that I have a fix for [here](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/06e482a7a1373b08e97f7852c7a1e5af5df63b48?tab=changes). Essentially, you have a different `init.defaultBranch` value than `master` and it was assumed in the test suite that it would be `master`.

Ah, good to hear it's something trivial. In the `gitoxide` test-suite, I do my best to prevent `git` from picking up anything more than what's local to the repository at hand, and to control a few key variables - these have accumulated over time and might be useful here as well. Here is how this looks like right now: https://github.com/Byron/gitoxide/blob/072ee32f693a31161cd6a843da6582d13efbb20b/tests/tools/src/lib.rs#L500-L524

> I actually witnessed this running on my Mac recently. We have no fix and it's an annoying race condition, so in this case it can be ignored.

Ok, I will keep that in mind. Thus far, it was so consistent though that I wasn't able to identify it as such though. There are at least two other tests, one with `concurrency` in it's name if I recall correctly, that can deadlock. Those I ignored, thinking that maybe one day there is a solution for that as well. It took a while to get `gitoxide`'s test-suite stable and it's so nice now (*to not have flaky failures*), I would wish you the same :).

> However, this one is new and I'd need to dig into what's going on:

I dug in yesterday but forgot to share my findings.  I could reproduce it with:

```sh
❯ sed -i 's/Hello World/Hello Radicle/' main.c
sed: 1: "main.c": invalid command code m
```

…and I think the MacOS fix for 'edit in place - save no backup' is:

```
❯ sed -I '' 's/Hello World/Hello Radicle/' main.c
sed: main.c: No such file or directory
```

Of course I wouldn't know yet how to make such a change in the sandbox test description except for copying it entirely, and will leave the fix to you.

Regarding you seeing two test failures with `peek_data_line()` where I see none, I did retry with the latest git compiled from source, v2.43.0rc2, but still could not reproduce the issue. To be sure I replaced that executable with a script that always fails, and indeed, now 24 out of 75 tests in `radicle-node` would fail as they definitely picked up the git binary I put in the `PATH`.

This leaves me thinking that *maybe*, this failure you observed is related to some other raciness that somehow leaves (or puts) data after the pack itself. Maybe it's related to the platform as well, and it would be interesting to know if this reproduces reliably for you, maybe even on different platforms.

FintanH:
I dug in yesterday but forgot to share my findings. I could reproduce it with:

> ```
> ❯ sed -i 's/Hello World/Hello Radicle/' main.c
> sed: 1: "main.c": invalid command code m
> ```
> …and I think the MacOS fix for 'edit in place - save no backup' is:
> ```
> ❯ sed -I '' 's/Hello World/Hello Radicle/' main.c
> sed: main.c: No such file or directory
> ```

Ah, something similar happened to me where `ls` would show hidden files on MacOS if you were the superuser :facepalm: Seems like we need to think through cross-platform testing for the CLI code.

> This leaves me thinking that maybe, this failure you observed is related to some other raciness that somehow leaves (or puts) data after the pack itself. Maybe it's related to the platform as well, and it would be interesting to know if this reproduces reliably for you, maybe even on different platforms.

Interesting, I can rerun again a few times. I think I'll ask some folks to try it out too since we're all running a host of different Linux OSes!

FintanH:
@Byron can you push a version that includes the assertions? I just realised that the current version is still the one with the sink.

Byron:
I could have, but didn't want to keep spamming force-pushes which also trigger notifications, and instead opted for providing a patch which is reproduced here:

```patch
diff --git a/radicle-fetch/src/transport/fetch.rs b/radicle-fetch/src/transport/fetch.rs
index 5dfc1535..4558df49 100644
--- a/radicle-fetch/src/transport/fetch.rs
+++ b/radicle-fetch/src/transport/fetch.rs
@@ -269,7 +269,7 @@ where
             );
             // Consume anything that might still be left on the wire - this is 'EOF' most of the time,
             // but some tests have 'garbage' here as well.
-            std::io::copy(&mut reader, &mut std::io::sink())?;
+            assert!(reader.peek_data_line().is_none());
             assert_eq!(
                 reader.stopped_at(),
                 Some(MessageKind::Flush),
```

If you want me to force-push nonetheless or push a separate commit for this change, I will of course (it's just that the current version is the known good one that I didn't want to meddle with for trying things).

FintanH:
The patch works :) I meant just push a separate branch though :grin:

Byron:
Alright, I just pushed a new commit then that is for later removal, knowing that I now caused the opposite of what I originally intended: to reduce notifications 😅.

By the way, please do feel free to push or force-push directly into this branch if it helps with testing in any way, no need to go through me. Reviews are much faster and easier that way I find. When checked out with `gh pr checkout 556` changes can be pushed back without doing anything special.

FintanH:
Thanks for that! So I got some help from colleagues, and you can see the thread here https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/debugging.20gitoxide.20packets

I realised that you might not have seen errors because if `radicle-cli` fails, for some reason, `radicle-node` tests will not run. Perhaps you could try run `cargo test -p radicle-node` with the assertion.

You can see, in the thread, that rudolfs is running MacOS and recent git version runs into the panic by running `cargo test -p radicle-node`.

Lars also ends up getting the panic on Debian.

Byron:
> Thanks for that! So I got some help from colleagues, and you can see the thread here https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/debugging.20gitoxide.20packets
> 
> I realised that you might not have seen errors because if `radicle-cli` fails, for some reason, `radicle-node` tests will not run. Perhaps you could try run `cargo test -p radicle-node` with the assertion.
> 
> You can see, in the thread, that rudolfs is running MacOS and recent git version runs into the panic by running `cargo test -p radicle-node`.

I was also running the tests like this to bypass `radicle-cli` failures initially, but could run everything together using `TMPDIR=/tmp cargo nextest run --no-fail-fast --workspace`, or individually. Unfortunately, with the patch applied, I never managed to reproduce the issue.

At this point I think one could do all or any of the following:

* ignore the spurious failures and make what's here mergeable, we know the 'good' version
* provide a patch to rudolfs so they can share which bytes are still in the channel (which is interesting to me as I'd hope it's 1 and 30 in decimal)

Maybe something else should be done as well - just let me know where we are heading :).

FintanH:
> At this point I think one could do all or any of the following:
> 
>     * ignore the spurious failures and make what's here mergeable, we know the 'good' version
> 
>     * provide a patch to rudolfs so they can share which bytes are still in the channel (which is interesting to me as I'd hope it's 1 and 30 in decimal)
> 
> 
> Maybe something else should be done as well - just let me know where we are heading :).

Why not both :stuck_out_tongue: 

I'm happy to take the 'good' version and merge it in :) I'll push the patch via Radicle and can add the link back to here.

Tagging @rudolfs to see if he has time for checking out the leftover bytes. I'm also happy to do it on my system too :+1:

Byron:
'All of the following' it is :).

I just have reverted the commit with test-code and have a patch here which does show the remaining bytes left in the channel (*I only see an empty string there though*).

```patch
diff --git a/radicle-fetch/src/transport/fetch.rs b/radicle-fetch/src/transport/fetch.rs
index 5dfc1535..5b45e927 100644
--- a/radicle-fetch/src/transport/fetch.rs
+++ b/radicle-fetch/src/transport/fetch.rs
@@ -1,3 +1,4 @@
+use bstr::ByteSlice;
 use std::{
     borrow::Cow,
     io::{self, BufRead},
@@ -269,7 +270,9 @@ where
             );
             // Consume anything that might still be left on the wire - this is 'EOF' most of the time,
             // but some tests have 'garbage' here as well.
-            std::io::copy(&mut reader, &mut std::io::sink())?;
+            let mut buf = Vec::new();
+            std::io::copy(&mut reader, &mut buf)?;
+            dbg!(buf.as_bstr());
             assert_eq!(
                 reader.stopped_at(),
                 Some(MessageKind::Flush),
```

I hope that helps to get closer to solving that riddle 😅.

Maybe contributing here isn't the preferred way of doing so, is there something I missed on how to do the same on Radicle? What I am missing in particular is CI as I can't really gauge the quality of my commits while there are local test-failures.

FintanH:
So I was trying to use the `dbg` patch and it was constantly showing up as empty. So I tried to dig in a bit deeper and used the following:
```rust
            let mut buf = Vec::new();
            assert!(reader.peek_data_line().is_none(), "buf={:?}", {
                std::io::copy(&mut reader, &mut buf)?;
                buf
            });
```
Here's the interesting part... 
```
thread 'z6Mkt67GdsW7715MEfRuP4pSZxJRJh6kj6Y48WRqVv4N1tRk <worker#1>    ' panicked at 'buf=[]', /home/haptop/Developer/heartwood/radicle-fetch/src/transport/fetch.rs:274:13
```

So it seems like `peek_data_line` can return `Some` with no data!

FintanH:
> Maybe contributing here isn't the preferred way of doing so, is there something I missed on how to do the same on Radicle? What I am missing in particular is CI as I can't really gauge the quality of my commits while there are local test-failures.

Aye, we create patches using the `rad` command-line tool which you can pick up here: https://radicle.xyz/#try

I've created a [patch](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/2ed2ba0689ac3847a743e8ea79fa983898242447). Unfortunately, we don't have review and conversational elements yet -- that happens on Zulip curently.

Byron:
Thanks for trying the patch, and for the pointer for trying the `rad` tool! I'd definitely love to try it particularly when conversation is available with it, and will be trying the `rad` + `zulip` workflow for the next contribution as well.

> So it seems like `peek_data_line` can return `Some` with no data!

If the code presented [in the comment](https://github.com/radicle-dev/heartwood/pull/556#issuecomment-1814205379) is the one to draw that conclusion from, then I have a different view on it. `peek_data_line()` returned `None` which means it didn't see any data (due to the flush packet). When using `std::io::copy()` from a source that is EOF, which is what the flush packet does, it comes out empty as well. So all there is to see is that there is nothing, or EOF by flush packet.

This seems to indicate though that having bytes left in there is racy? Maybe @rudolfs has more luck with it, seeing the bytes might help to learn where they are from.

And since there now is [the Radicle version of this PR](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/2ed2ba0689ac3847a743e8ea79fa983898242447) I think I will close this :).

FintanH:
> peek_data_line() returned None which means it didn't see any data (due to the flush packet). 

Perhaps I'm wrong, but the `assert!` will panic when `peek_data_line` is `Some`, since we're saying that we expect it to be `None`, i.e. `is_none`. So, `peek_data_line` is `Some` and the next argument to `assert!` will execute, copying whatever bytes are in the `reader` to the `buf`. Those copied bytes are empty, judging by the `buf=[]`.

> When using std::io::copy() from a source that is EOF, which is what the flush packet does, it comes out empty as well. So all there is to see is that there is nothing, or EOF by flush packet.

Are you saying that the `buf=[]` indicates that the flush packet is leftover?

Byron:
I enthusiastically exhale that I got this wrong. What I described is what I thought it would be, and what you described is what it actually is.
Yes, it's indeed funny and I think it's a bug which is 'implemented' here: https://github.com/Byron/gitoxide/blob/c3edef1c0c49accbb037bdf086dade3ed0e5e507/gix-packetline/src/read/blocking_io.rs#L152-L154 .

When it encounters EOF during peek, it doesn't return `None` but the then empty buffer, which explains why it's some. I should be able to add a test for 'peek EOF' to fix it.

Maybe the 'extra bytes' were that bug all along?

FintanH:
> I enthusiastically exhale that I got this wrong. What I described is what I thought it would be, and what you described is what it actually is.

Hehe, this happens to me all the time too.

I'm thinking it might be that we were unnecessarily panicking due to the `is_none` check when indeed we were finished reading :)

Byron:
And digging into this, there are tests which want validate that behaviour, too. But running into this here clearly shows it should be improved, if only the documentation should make clear when to expect `None`.

In the end, if there are no bytes available, there is we learn nothing new :/.

Byron:
Actually I have no explanation for this - it should return `None` on EOF and it does return it, too, there are tests for that and I even added a specific one to be very sure.

Maybe you could give it one last-ditch attempt and do `dbg!(peek_data_line())`, as the `Some()` might contain an error. If not, I could imagine it's actually empty, which can happen if the server sends `0004` I presume, even though [this is actively discouraged](https://github.com/git/git/blob/cfb8a6e9a93adbe81efca66e6110c9b4d2e57169/Documentation/gitprotocol-common.txt#L89). 

Thanks again!

Byron:
Can you imagine an archeologist who keeps digging, thinking it's going to be something big, and then all that's showing is a shard of a coffee mug, or something equally benign 😁?

[The playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=998ff76f15f6db85efa6b1580fc126d0) revealed the decoded message is:

```
"\u{2}Total 5 (delta 1), reused 0 (delta 0), pack-reused 0"
```

The `2` means it's a sideband message for stderr, which is typically used for progress messages while receiving/waiting for the pack. These definitely happen out of order and apparently they can also arrive a little late, probably particularly for small packs that are basically just a single packetline. The progress-handler is set for the reader, but it's not triggering while peeking as that would be a consuming action, which is why it's visible in that way. This also explains why it's empty when actually reading from it - it's progress after all, not data.

At least that riddle is solved now, case closed :D.

PR #557, opened 2024-03-06
Fix linker error on macos 14.3 with nix build

telekom-tbehner:
Using nix build resulted in a linker error

ld: framework not found Security

Following the proposed solution in
https://discourse.nixos.org/t/compile-a-rust-binary-on-macos-dbcrossbar/8612 resolves the issue.


telekom-tbehner:
I created a corresponding patch with `rad`: a6734b6
Awesome project!

PR #559, opened 2026-04-30
Spelling

jsoref:


.radicle/ambient.yaml
```diff
@@ -40,7 +40,7 @@ plan:
       # `debian` directory, compared to what is committed to Git, from
       # which the "upstream tarball" is created.
       git reset --hard
-      git clean -fdx
+      git clean -xdf
```
lorenzleutgeb: What's the reason for this change? It does not seem spelling related to me. You say to "harmonize" and "match jj", but I do not understand what you mean by that, sorry.

crates/radicle-protocol/src/service.rs
```diff
@@ -100,7 +100,7 @@ pub const MAX_CONNECTION_ATTEMPTS: usize = 3;
 /// How far back from the present time should we request gossip messages when connecting to a peer,
 /// when we come online for the first time.
 pub const INITIAL_SUBSCRIBE_BACKLOG_DELTA: LocalDuration = LocalDuration::from_mins(60 * 24);
-/// When subscribing, what margin of error do we give ourselves. A igher delta means we ask for
+/// When subscribing, what margin of error do we give ourselves. A higher delta means we ask for
```
lorenzleutgeb:
```suggestion
/// When subscribing, what margin of error do we give ourselves. A greater delta means we ask for
```

lorenzleutgeb (changes requested):
Thanks! Apart from the comment and the suggestion, I am wondering about the commit messages. They don't really match the format we commonly use. I am also leaning towards squashing most of these changes, so that this PR would carry less commits. For example:

```
treewide: Spelling
(this commit would contain all corrections for comments and documentation, but not Rust code)
term: Rename `Row::Divid{i,}er`
(this commit would contain the rename which actually touches Rust code)
node/test: Fix Spelling
(this commit would contain the renaming of the test functions)
```

The intention here is...
 1. ...to have less commits for comments and documentation.
 2. ...to have separate commits, per crate, for changing Rust code.

.radicle/ambient.yaml
jsoref: This is output by jj for the same purpose:
https://github.com/search?q=repo%3Aradicle-dev%2Fheartwood+%22git+clean+-xdf%22&type=code

crates/radicle-protocol/src/service.rs
jsoref: Sure

jsoref:
I'm not attached to the commit messages. I default to `spelling: ...` per change because it makes it easy for me to fix / drop things as requested. I tried to conform to my interpretation of https://github.com/radicle-dev/heartwood/blob/master/CONTRIBUTING.md#writing-commit-messages

https://github.com/radicle-dev/heartwood/blob/b482845e712d54fdee00b0548c828e6694a3f359/CONTRIBUTING.md?plain=1#L256-L261

upon rereading, it seems I misunderstood the prompt since this clearly violates the above:
https://github.com/radicle-dev/heartwood/blob/b482845e712d54fdee00b0548c828e6694a3f359/CONTRIBUTING.md?plain=1#L263-L265

I can try to apply the squashes you're suggesting... But, I'll wait for your reply to the `jj` item before i do anything.

.radicle/ambient.yaml
lorenzleutgeb: Okay. I don't see why we would have to order our arguments the same way here though. What would be the reason for doing so?

jsoref: Order doesn't matter per se. Being consistent in order makes it easy to see "where else is this used".

jsoref: I've dropped this.

crates/radicle-fetch/src/stage.rs
```diff
@@ -270,7 +270,7 @@ impl ProtocolStage for CanonicalId {
         let verified = repo
             .identity_doc_at(
                 *s.canonical_rad_id()
-                    .expect("ensure we got canonicdal 'rad/id' ref"),
+                    .expect("ensure we got canonical 'rad/id' ref"),
```
lorenzleutgeb, line 273: This should move to the commit titled "treewide: ...".

crates/radicle-node/src/wire.rs
```diff
@@ -722,7 +722,7 @@ where
                                     ChannelsConfig::new(FETCH_TIMEOUT)
                                         .with_reader_limit(reader_limit),
                                 ) else {
-                                    log::debug!(target: "wire", "Peer attempted to open already-open stream stream {stream}");
```
lorenzleutgeb, line 725: This should move to the commit titled "treewide: ...".

crates/radicle-oid/src/lib.rs
```diff
@@ -247,7 +247,7 @@ pub mod str {
         }
 
         #[quickcheck]
-        fn gix_roundrip(oid: Oid) {
```
lorenzleutgeb, line 250: I missed that earlier. It should go into another commit "oid: ...".

crates/radicle-fetch/src/stage.rs
FintanH: @lorenzleutgeb I’m wondering why these things aren’t getting picked up by our hooks 🤔

jsoref: What hooks do you use? This is what I'm using: https://github.com/jsoref/heartwood/actions/runs/25186147402/attempts/1#summary-73843841790

FintanH: We use justfile for running all our checks

https://github.com/radicle-dev/heartwood/blob/master/justfile

specifically we’re using typos and codespell

jsoref: Speaking as the author of check-spelling, those are basically hard-coded lists of known mistakes / corrections.

By contrast, check-spelling compares your code against a full dictionary and identifies anything it can't find in the dictionary.

If you make a creative misspelling, typos/codespell won't help. As long as your creative misspelling isn't a word or is in a list of patterns for known errors, check-spelling can help.

The misspelling here is a good example of this. It isn't a _common_ mistake, so those other tools won't help. But it isn't in the dictionary, so check-spelling will.

<details><summary>p.s. If you're interested in a model that's compatible w/ justfile</summary>

You could look at how Rancher Desktop deploys check-spelling:
https://github.com/rancher-sandbox/rancher-desktop/blob/main/.github/workflows/spelling.yml

https://github.com/rancher-sandbox/rancher-desktop/blob/main/scripts/spelling.sh

(You don't need to do it quite that way, check-spelling can install its dependencies on its own, but the Rancher Desktop folks wanted to control things...)

</details>

FintanH: Oh amazing! Thanks for the pointers, we’ll have to check it out

lorenzleutgeb (approved):

